Issue #1, opened 2017-03-30
Remove bullets before results on grant lookup page


TomSteinberg:
Just a test!

PR #3, opened 2017-03-31
Remove bullet points from grants

mattandrews:
Fixes https://github.com/biglotteryfund/blf-alpha/issues/1

Now looks like this:
image

Is this okay to merge @TomSteinberg?


TomSteinberg:
Great - and you don't really need to ask me before merging things like this

PR #172, opened 2017-09-27
Track contrast

mattandrews:
Quick fix as suggested by @davidrapson – add a bit of Google Analytics tracking to the contrast switcher so we can see who's using it.

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -54,13 +54,15 @@ $('#js-close-overlay').on('click', () => {
     $('#js-overlay').hide();
 });
 
-// if the env is production
-// then only load analytics
-// if the domain name is live domain
+// if the env is PRODUCTION then we only load analytics
+// if the page's domain name matches our live domain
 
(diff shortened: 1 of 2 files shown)


assets/js/main.js
```diff
@@ -83,21 +86,41 @@ if (!_BLF.blockAnalytics) { // set in main.njk
...
-}
 
-$('.js-track-clicks').on('click', function () {
-    let category = $(this).data('category');
-    let action = $(this).data('action');
-    let label = $(this).data('label');
-    analytics.track(category, action, label);
-});
+    // track interactions with in-page elements
+    $('.js-track-clicks').on('click', function (e) {
```
davidrapson: FWIW the official auto-track plugin for Analytics implements this style of declarative event tracking and outbound link tracking so might be worth looking into that to see that does the job and we can avoid writing any custom code: https://ga-dev-tools.appspot.com/autotrack/

PR #176, opened 2017-09-28
Order form validation fix

mattandrews:
This is an edge case, but has bothered me for a while, so here comes the fix.

Basically: when you order something from the [free materials form](http://www.biglotteryfund.org.uk/funding/funding-guidance/managing-your-funding/ordering-free-materials?lang=bilingual#your-details), it's possible to fail server-side validation and be returned to the form. 

Sadly, when this happens, the page's content is then hidden via JavaScript (eg. to make you choose monolingual/bilingual), so the form you're being returned to is hidden from you.

In practice this is unlikely to happen often, as the form has client-side validation (for required fields, and email fields), so the only users who will experience this bug are people in older browsers that don't support HTML5 form validation, or people who edit the form HTML to remove the validation(!).

This change adds two things: it updates the URL when a language preference is made for products (so sharing/reloading that URL will now show the correct products), and also passes this preference through the form, so when a validation error happens, the user is returned to the correct product page *and* shown their invalid form to correct.

### DIFF
--- a/controllers/funding/index.js
+++ b/controllers/funding/index.js
@@ -146,7 +146,7 @@ module.exports = (pages, sectionPath, sectionId) => {
                         let redirectUrl = returnUrl;
 
                         // add their langage choice (if valid)
-                        let langChoice = req.body['languageChoice'];
+                        let langChoice = req.body.languageChoice;
                         if (langChoice && redirectUrl.indexOf(langParam) === -1 &&
                             ['monolingual', 'bilingual'].indexOf(langChoice) !== -1) {
                             redirectUrl += langParam + langChoice;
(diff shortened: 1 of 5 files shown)


controllers/funding/index.js
```diff
@@ -138,13 +138,31 @@ module.exports = (pages, sectionPath, sectionId) => {
...
-                        res.redirect(req.baseUrl + freeMaterials.path + '#your-details');
+                        // build a redirect URL based on the route, the language of items,
+                        // and the form anchor, so the user sees the form again via JS
+                        let returnUrl = req.baseUrl + freeMaterials.path;
+                        let formAnchor = '#your-details';
+                        let langParam = '?lang=';
+                        let redirectUrl = returnUrl;
+
+                        // add their langage choice (if valid)
+                        let langChoice = req.body['languageChoice'];
```
davidrapson: Any reason for `req.body['languageChoice']` over `req.body.languageChoice`?

mattandrews: No – good call, have changed to match other uses!

PR #180, opened 2017-09-29
Make contrast buttons high-contrast

mattandrews:
cc @ChloeAlper 

The buttons we use for switching to high contrast accessibility mode...

image

... fail accessibility tests for contrast!

image

So this change makes them dark grey:

image

... which passes:

image

Let me know if this is okay @ChloeAlper.


ChloeAlper:
Looks good to me too

PR #189, opened 2017-09-29
Prettier Setup

davidrapson:
Put together a quick example of using [prettier](https://github.com/prettier/prettier) alongside husky and lint-staged. This does two main things:

- Auto formats javascript files to improve consistency
- Automates management of git hooks

I've set this to auto format an controller files that are staged for now. Formatted the tools controller for an example. Not expecting to necessarily take this any further at this point but though it was work raising a straw-man PR to give an example of the output.


davidrapson:
Tidied this up so it's not sitting around and can get merged:

- Expands scope of prettier to cover server source (as discussed things like this might get easier if we move server code into a `server/` sub-directory, but this is more than fine for now)
- Removes manual pre-commit script and moves tests to a `prepush` hook

So in summary:

1. On commit any files in scope that are staged will get auto formated
2. On push the tests will be run so you can't push if tests fail (you can pass a `--no-verify` flag to skip this process)

PR #225, opened 2017-10-04
Fix Welsh legacy site

mattandrews:
This change allows the proxied homepage to detect a request for the welsh homepage and request it from the legacy site (rather than always serving English).

It also removes an unused `http-proxy` server which was an earlier attempt at fetching the legacy page 😊 

Finally it updates the Cloudfront script to remove trailing slashes from behaviour paths, which was breaking the Welsh homepage path (turning it from `/welsh` to `/welsh/`).


davidrapson:
(also nice cleanup)

PR #227, opened 2017-10-04
Update legacy homepage to proxy on post requests

davidrapson:
Ensures that the funding finder and global search still works on the legacy homepage when proxied through cloudfront and our express app.


controllers/toplevel/index.js
```diff
@@ -71,7 +71,7 @@ const oldHomepage = (req, res) => {
         resolveWithFullResponse: true
     }).then((response) => {
         let body = response.body;
-        // convert all links in the document to be root-relative
+        // convert all links in the document to be absolute
```
mattandrews: 😢

PR #229, opened 2017-10-04
Configure eslint to run on all files and not just assets

davidrapson:
Currently only running on client-side code. This PR expands eslint to run on all application code. Could've helped flagged unused code the code removed like https://github.com/biglotteryfund/blf-alpha/pull/225 automatically.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,3 @@
-node_modules
+assets/js/libs/
+node_modules/
+public/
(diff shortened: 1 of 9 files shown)


.eslintignore
```diff
@@ -1 +1,4 @@
-node_modules
+assets/js/libs/
+bin/data/data.js
```
mattandrews: This got removed in my cleanup so should go from here too

gulpfile.js
```diff
@@ -1,7 +1,6 @@
-/* jshint node: true */
```
davidrapson: `gulpfile.js` has also been formatted with `prettier`

PR #234, opened 2017-10-05
Add hero rotation functionality

davidrapson:
hero-rotation

This PR adds support for randomising images on reload.

- Migrate JS / IE classes to head to reduce flash of styles
- Convert super-hero to standalone component w/picture, no longer coupled to previous hero styles
- Add support for randomising images on reload and fallback to hero image for no-JS.


assets/js/modules/heroImages.js
```diff
@@ -0,0 +1,50 @@
...
+    image: '.js-random-hero__image',
+    caption: '.js-random-hero__caption'
+};
+
+const rand = xs => xs[Math.floor(Math.random() * xs.length)];
+
+function replaceImage($imageEl, randomImage) {
+    $imageEl.html(`
+        <picture>
+            <source srcset="${randomImage.large}" media="(min-width: 760px)">
```
mattandrews: Is this `760` tied to the tablet breakpoint? It's available in JS – have a look at line 6 of `carousel.js`. Just in case we want to make this dynamic.

controllers/toplevel/index.js
```diff
@@ -20,20 +20,47 @@ const robots = require('../../config/app/robots.json');
...
-        let lang = req.i18n.__('toplevel.home');
+    const serveHomepage = news => {
+        const lang = req.i18n.__('toplevel.home');
+
+        const heroImageDefault = {
+            small: assets.getCachebustedPath('images/hero/home-hero-1--small.jpg'),
+            large: assets.getCachebustedPath('images/hero/home-hero-1--large.jpg'),
+            default: assets.getCachebustedPath('images/hero/home-hero-1--small.jpg'),
+            caption: 'Cloughmills Community Action, Grant £4,975*'
+        };
```
mattandrews: Half-wondering if we should make this a function that will complain if we miss out all the parameters etc? It could also do the image cachebusting in one go, too, so you'd just need to pass the paths.

mattandrews: ... would also be easier to later extend it to accept a translated caption for the inevitable time we need do this.

views/includes/metaHeadJS.njk
```diff
@@ -0,0 +1,28 @@
+<script id="script-head">
+(function() {
+    var docEl = document.documentElement;
+    var docClass = docEl.className;
+    var classes = ['js-on'];
+
+    var AppConfig = {
+        isModernBrowser:
```
mattandrews: 💯

views/components/hero.njk
```diff
@@ -1,18 +1,28 @@
...
-            <div class="hero__content inner">
-                <div class="hero__header" id="content">
-                    <h1 class="hero__title">{{ title | safe }}</h1>
-                    {{ content | safe }}
-                </div>
-                {% if caption %}
-                    <span class="hero__caption u-caption u-desktop-only">{{ caption }}</span>
-                {% endif %}
+{% macro superHero(title, content, imageDefault, imageCandidates) %}
+    <div class="super-hero js-random-hero" data-image-candidates='{{ imageCandidates | jsonStringify | safe }}'>
```
mattandrews: Do you think it'd be neater to output a `<script>` block here (with a unique ID somehow) and refer to it in the `data` attribute, and then just grab it directly in the JS, rather than parsing JSON? Or would that just increase the chance of errors with global vars etc?

assets/js/modules/heroImages.js
davidrapson: I've just changed this so it's now far less tied to the general breakpoints…otherwise yes!

views/components/hero.njk
davidrapson: I personally like the self-container nature of this, even with the extra step. Prefer avoid global variables if possible.

controllers/toplevel/index.js
davidrapson: Great shout, I've done that now 💯

views/components/hero.njk
mattandrews: 🇹🇭

mattandrews: 👍

PR #239, opened 2017-10-05
Update 404 test to validate we have a real page

davidrapson:
Update 404 test to validate we have a real page to avoid regressions identified in https://github.com/biglotteryfund/blf-alpha/pull/236


davidrapson:
This should fail until I put the error handler code back in. Replicating the original issue.

PR #240, opened 2017-10-05
Responsive tweaks

davidrapson:
Implements a few responsive tweaks mainly focussed on the global header and heroes but also adjusting global gutters.

- Normalises gutters to pixels: 10px -> 20px -> 40px for small -> medium -> large breakpoints
- Decreases gap between logo and navigation (point 1. in annotations)
- Increases spacing between navigation items at "iPad portrait" breakpoints (point 2. in annotations)
- Decreases spacing around search icon (point 3. in annotations)
- Increases padding between buttons at "iPad portrait" breakpoints
- Prevents buttons from wrapping at "iPad  landscape" breakpoints

**~Portrait (Before)**

ipad-portrait-before

**~Portrait (After)**

ipad-portrait-after

**~Landscape (Before)**

ipad-landscape-before

**~Landscape (After)**

ipad-landscape-after

@ChloeAlper @mattandrews

### DIFF
--- a/views/pages/toplevel/home.njk
+++ b/views/pages/toplevel/home.njk
@@ -155,16 +155,19 @@
                     </div>
 
                     {# social buttons #}
-                    <div class="gutter-mob-only">
-                        <a href="{{ appData.config.get('social.instagram') }}" class="btn btn--small btn--outline accent--blue btn--block spaced">
+                    <div class="o-button-group-block u-gutter-half">
+                        <a class="btn btn--small btn--outline accent--blue"
+                            href="{{ appData.config.get('social.instagram') }}">
                             <span class="btn__icon">{% include "../../includes/svg/instagram.svg" %}</span>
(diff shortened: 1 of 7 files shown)


views/pages/toplevel/home.njk
```diff
@@ -48,9 +48,8 @@
                     <span class="u-wide-only">{{ copy.callsToAction.under10k.long }}</span>
                     <span class="u-not-wide">{{ copy.callsToAction.under10k.short }}</span>
                 </span>
-            </a>
-
-            <a class="btn btn--medium
+            </a><!-- Fix inline-block spacing
+            --><a class="btn btn--medium
```
mattandrews: I'm not a fan in general of this technique but understand why we're using it – if we need to do it anywhere else though maybe let's use flexbox or similar instead?

ChloeAlper:
These are looking great David. Would it be possible to shoehorn a bit more space between the 10k buttons? (see amend 5)

PR #241, opened 2017-10-05
Remove <p> from base font-stacks

davidrapson:
This was bugging me: `<p>`s are currently included in the base font-stacks, this makes it harder to cascade typographic styles for blocks of text. Using `body` is enough. This is a one-line change but scoping it to a PR as it's in core styles.


davidrapson:
Ran through all the active pages to double check and I'm happy this doesn't cause any issues to remove.

Issue #201, opened 2017-10-02
Ensure that all users clicking on country links on the above 10k page are served pages that meet their expectations


TomSteinberg:

mattandrews:
This is now live – visitors to the [Over £10k](https://www.biglotteryfund.org.uk/over10k) page will now see a filtered (proxied) Funding Finder, eg: https://www.biglotteryfund.org.uk/funding/funding-finder?area=Wales&over=10k

PR #244, opened 2017-10-06
Proxy the funding finder

mattandrews:
This fixes the opaquely-named #201.

It moves the legacy homepage proxy code out into a generic middleware that can be applied to any route. This supports both `POST` and `GET` HTTP methods (in the case of the former, it follows redirects as this appears to be all that's required with the ASP.NET forms the legacy site uses).

There's also some specific code here for modifying the DOM returned by the (legacy) Funding Finder, specifically to filter out funding programs under £10,000 (as the URL parameters don't allow this directly). This is brittle (if the text content changes on the proxied pages then this will break) but likely the best method available (for now).

Finally, in order to actually launch these proxied pages we need to add the Funding Finder path to the Cloudfront script, but I plan to do that in a separate PR as it'll effectively make this change live once merged.


modules/proxy.js
```diff
@@ -0,0 +1,131 @@
...
+
+            // fix meta tags in HTML which use the wrong CNAME
+            body = body.replace(/wwwlegacy/g, 'www');
+
+            // parse the DOM
+            let dom = new JSDOM(body);
+
+            if (req.app.get('env') === 'development') {
+                let titleText = dom.window.document.title;
+                dom.window.document.title = '[PROXIED] ' + titleText;
```
mattandrews: Not sure what you think about leaving this in @davidrapson, it was solely to make it easier to eyeball a page that's been modified/proxied, as opposed to redirected/external.

controllers/toplevel/index.js
```diff
@@ -99,104 +95,60 @@ const newHomepage = (req, res) => {
...
-
-            const form = dom.window.document.getElementById('form1');
-            if (form) {
-                const newAction = form.getAttribute('action').replace('https://www.biglotteryfund.org.uk/', '/');
-                form.setAttribute('action', newAction);
-            }
+// serve the legacy site funding finder (via proxy)
+router.get('/funding/funding-finder', (req, res) => {
+    // parse the string (eg. "£10,000" => 10000, "£1 million" => 1000000 etc)
+    let parseValueFromString = str => {
```
davidrapson: @mattandrews If we're going to add unit tests for any of this, this `parseValueFromString` function is probably a good candidate.

modules/proxy.js
davidrapson: Yep, I like this (it's scoped to development envs right?)

controllers/toplevel/index.js
mattandrews: Have a look at the newest commits!

davidrapson:
👍  tests look good

PR #249, opened 2017-10-09
New case study

davidrapson:
- Adds new case study (will contact Dave for translation once this is up on the test site and has been reviewed)
- Extracts case study template into standalone macro
- Allows support for HTML for citation

screen shot 2017-10-09 at 15 18 35


davidrapson:
@mattandrews Added tracking for carousel as part of this too:

PR #250, opened 2017-10-09
Add nodemon and gulp-cli to devDependencies

davidrapson:
Small change to add `nodemon` and `gulp-cli` to `devDependencies` so they don't need to be installed globally.


mattandrews:
Good call – not sure why it was like this to start with, perhaps only because I've only recently came down on the `nodemon` standard side... 👍

PR #252, opened 2017-10-10
Styleguide

davidrapson:
A rough pass at bringing the style-guide online. Simplified underline and tracking modifiers whilst I was here.

localhost-3000-styleguide 1

### DIFF
--- a/assets/sass/globals/accents.scss
+++ b/assets/sass/globals/accents.scss
@@ -16,7 +16,6 @@
     $colour: palette($name);
 
     .accent--#{$name} {
-
         // text color
         &.a--text {
             color: $colour;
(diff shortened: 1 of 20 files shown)


assets/sass/globals/accents.scss
```diff
@@ -66,11 +65,12 @@
...
-        // border-bottom below some type styles
-        &.t3--a {
-            &:after {
-                border-bottom-color: $colour;
-            }
+        /**
+         * Modifier class which adds a border below the text
+         * (see typography.scss)
+         */
+        &.t--underline:after {
```
mattandrews: 👏

views/layouts/styleguide.njk
```diff
@@ -0,0 +1,47 @@
+<!doctype html>
+<html lang="{{ locale }}" class="no-js locale--{{ locale }}{% if highContrast %} contrast--high{% endif %}">
```
mattandrews: Would it be overkill to make this a shared template too? Feels like these are quite "core" styles so might be a bit confusing if they're in two places (even for an exceptional page like this).

mattandrews: (to be specific, I mean the `<html>` element)

PR #265, opened 2017-10-12
Module code refactor

davidrapson:
~To avoid needing to preprocess lots of code in order run unit tests this PR reconfigures our tests to all run directly through mocha, replacing `phantomjs` for `jsdom` for the one test that requires a DOM (tabs). This PR also loads `jquery` as an npm dependency.~

The scope of this has expanded so reframing this:

- Removes all committed client-side vendor code (jQuery, Swiper, Grapnel) and loads them as npm dependencies instead
- Removes need for Grapnel and moves associated code into modules
- Decouples unit test code to not need to be precompiled, uses jsdom instead of phantomjs
- Removes need for global gulp in travis.yml and runs build step separately to tests


.travis.yml
```diff
@@ -9,7 +9,7 @@ before_script:
 - npm install
 script:
 - npm run lint
-- gulp test --production
```
mattandrews: This command does two things really, runs the tests but also compiles everything with the `--production` flag meaning that CSS/JS is minified etc. We'd need either add another step after this to compile assets, or add that flag to the `npm test` script.

assets/js/main.js
```diff
@@ -1,12 +1,14 @@
...
 
+const $ = require('jquery');
+
 // initialise router (eg. run conditional code for certain URLs)
 const Grapnel = require('./libs/grapnel');
 const router = new Grapnel({ pushState: true });
 
 // initialise Vue
-const Vue = require('../../node_modules/vue/dist/vue.common');
+const Vue = require('vue');
```
mattandrews: This is wrong – look in devtools and you'll see the warnings!

.travis.yml
davidrapson: Yep, as discussed, added that in as a separate step.

assets/js/main.js
davidrapson: Accidental hold-over from some webpack experimentation. Reinstated now.

PR #266, opened 2017-10-13
Switch to Webpack for builds

davidrapson:
This PR switches our JavaScript build step to Webpack and streamlines the existing gulp tasks as well as opting for a simpler method for hashing assets.

Plenty here so probably worth a chat through all of this.

## Webpack

Configures Webpack with the responsibility of bundling JavaScripts only, doesn't attempt to bundle images, stylesheets, or do any asset hashing. These processes are still left to gulp.

## NPM script aliases

Rather than needing to have `gulp-cli` installed globally and because this build approach uses a hybrid of gulp and non-gulp tooling I've aliased all commands to npm scripts.

- `npm run build` to run a one off development build
- `npm run build-production` to run a production build
- `npm run watch` to run a local watcher for webpack and gulp changes

## Commit based asset hashing

In order to avoid the accidental complexity involved in hashing individual files I've opted to switch to a less complex / albeit more naive asset hashing method. Instead of hashing individual files all assets are bundled into a build directory based on the current commit hash in production (and a 'latest' directory in development).

This comes with the tradeoff that all assets will be cache busted when deploying rather than just the changed files. Given we're on a ~daily rather than continuous cadence for deployments I'm OK with this trade-off but feel free to shoot holes in this argument.

### Pros

- Moves javascript bundling to a more actively supported tool
- Improves module loading opportunities (conditional imports for analytics code / code-splitting) 
- Allows us to use import/export syntax in client code if we want to
- Fewer gulp-* plugin dependencies and a smaller gulpfile
- Simpler asset hashing process

### Cons

- Asset hashing is more naive, all files are cache-busted at once
- Mix of tooling? Gulp and npm at the same time


webpack.config.js
```diff
@@ -0,0 +1,67 @@
...
+        var plugins = [
+            new webpack.DefinePlugin({
+                __DEV__: JSON.stringify(JSON.parse(!isProduction || 'false')),
+                'process.env': { NODE_ENV: JSON.stringify(process.env.NODE_ENV) }
+            })
+        ];
+
+        if (isProduction) {
+            plugins = plugins.concat([
+                new webpack.optimize.UglifyJsPlugin({
```
mattandrews: Can we silence this a bit? [Logs](https://travis-ci.org/biglotteryfund/blf-alpha#L1211) are quite noisy from this.

gulpfile.js
```diff
@@ -140,39 +74,37 @@ gulp.task('styles', sassTaskDeps, function() {
...
 
-// dev task: start the server and watch for changes
-gulp.task('default', ['clean:assets', 'dev']);
+    fs.writeFile(MANIFEST_PATH, JSON.stringify(manifestData, null, 4), done);
+});
 
-// used for local dev testing (doesn't version files)
-gulp.task('dev', ['styles', 'scripts', 'watch']);
+/**
+ * Watchers
```
mattandrews: 👀

davidrapson: But who watches them…

webpack.config.js
davidrapson: Done

PR #267, opened 2017-10-13
Funding Finder CMS proof-of-concept

mattandrews:
Mildly uncertain about merging this, although no (live) pages will make a call to the CMS endpoint.

To see the new pages, `funding/programmes?location=northernIreland&min=10000` or similar URLs. The Over £10k page linked to these URLs at one stage in this branch but that would mean merging this would cause those links to break.

This has had almost no design/UX effort but the point really is to explore driving a page from dynamic data via a CMS rather than language files etc.

The `content.js` module right now is really basic but perhaps the wider discussion here should be about how we plan to integrate deployment of the CMS and testing changes to its API endpoint alongside this app... thoughts very welcome.


mattandrews:
(remember to update your secrets before running this locally, so you have the CMS endpoint URL)

PR #268, opened 2017-10-16
Improve analytics bootstrapping

davidrapson:
- Configure support for code-splitting
- Extracts analytics code out into a separate code-split module
- Configures analytics to use beacon transport where supported
- Configures maxScrollDepth tracker plugin

The latter point is the main change of note:

Configures metrics based on https://github.com/googleanalytics/autotrack/blob/master/docs/plugins/max-scroll-tracker.md#overview to track max scroll depth through analytics reports.

screen shot 2017-10-16 at 10 23 19
screen shot 2017-10-16 at 10 07 47
screen shot 2017-10-16 at 10 07 28


webpack.config.js
```diff
@@ -14,7 +14,9 @@ module.exports = {
     },
     output: {
         filename: '[name].js',
-        path: path.resolve(__dirname, buildSummary.buildDir, 'javascripts')
+        chunkFilename: '[name].bundle.js',
+        path: path.resolve(__dirname, buildSummary.buildDir, 'javascripts'),
+        publicPath: `${buildSummary.buildDir.replace('./public', '/assets')}/javascripts/`
```
davidrapson: Need to find a better way to define this 🙈

davidrapson:
Blergh…https://github.com/googleanalytics/autotrack/issues/137#issue-208245680

Seems like this import-individual-modules method is a quagmire as autotrack plugins are published as ES6 only so require compiling in our project (which is a weird choice). Seems like it's more viable to create a custom build…

davidrapson:
Custom build + async loading method ending up working well. Build should pass now 🤞

mattandrews:
Changes look good 👍

davidrapson:
(going to merge this in but wanted to run some tests and maybe wait until post-homepage meeting to avoid interfering to much with analytics setups mid-data-gathering)

davidrapson:
Ran through a few more tests on this and I'm happy this doesn't change reporting, going to deploy to test and run one more check with the a/b test in place.

PR #272, opened 2017-10-17
Cloudfront update improvements

mattandrews:
This took a few different forms along the way, but I think this is probably the most sensible one.

With this change, every time a modification is made to `controllers/routes.js`, a `prepush` script will run `./modules/cloudfront.js`. 

This script is a modified version of our original Cloudfront updater: it'll generate the `CacheBehaviors` array which is used by Cloudfront to map URL routes to different origins. This data is stored in the app (and committed here).

Separately, the Cloudfront update script has been modified to fetch this config file from Github, rather than operating on whatever settings exist locally (which may be on a non-`master` branch, may contain errors/test settings, etc). Run without arguments, the script will fetch the build currently on TEST/PROD (depending on whether the `-l`/`--live` flag was supplied) and use the associated config file. Otherwise, a commit ID can be passed for manual deployment (via the `-c`/`--commit`) flag, and finally a local configuration file can be supplied by passing a file via the (pre-existing) `-f`/`--file` option).

This means we can track changes in Cloudfront configuration via source control, and have more confidence when running Cloudfront update scripts that a reviewed, valid config will be used.


mattandrews:
Also this is basically Schrödinger's pull request, because the Cloudfront script will fail (Github's API won't find the JSON files) until this PR is merged. To test locally you can generate the config files and then point the script at them:

    ./modules/cloudfront.js
    ./bin/scripts/update-cloudfront.js -f ./bin/cloudfront/test.json

(you should see a warning about adding the styleguide and funding finder to Cloudfront)

package.json
```diff
@@ -34,6 +34,9 @@
     "assets/**/*.scss": [
       "prettier --write",
       "git add"
+    ],
+    "controllers/routes.js": [
```
davidrapson: Few thoughts on this.

1. This is in the lint-staged block, so will run on pre-commit, not pre-push. Is that intentional?
2. Should `./modules/cloudfront.js` not live in `bin/scripts` as it's executable?

modules/cloudfront.js
```diff
@@ -0,0 +1,221 @@
...
+        MinTTL: BehaviourConfig.TTLs.min,
+        Compress: false
+    };
+};
+
+const generateUrlList = () => {
+    let urlList = _.cloneDeep(URLs);
+
+    // add auto URLs from route config
+    for (let s in routes.sections) {
```
davidrapson: Wonder if it makes sense to pull this out into a form we could write a unit test for, given it's complexity 🤔

mattandrews:
@davidrapson have a look at the new commits when you get a sec!

bin/scripts/generate-cloudfront-config.js
```diff
@@ -127,6 +127,9 @@ const generateBehaviours = (distribution, environment) => {
             });
         }
     }
+
+    // sort by path (to make diffing easier)
+    behaviours = _.sortBy(behaviours, 'PathPattern');
```
davidrapson: Could straight `return sortBy(behaviours, 'PathPattern')` rather than temporarily reassigning?

test/specs/utilities.test.js
```diff
@@ -5,6 +5,42 @@ const expect = chai.expect;
...
+    sections: {
+        purple: {
+            path: '/purple',
+            pages: {
+                monkey: {
+                    path: '/monkey/dishwasher',
+                    live: true,
+                    isWildcard: false,
+                    isPostable: false,
+                    aliases: ['/green/orangutan/fridge']
```
davidrapson: 💚  🐒 ❄️ 

(this is the best I could do with emoji)

modules/utilities.js
```diff
@@ -46,8 +46,83 @@ const createHeroImage = opts => {
     };
 };
 
+// take a route config and format it for cloudfront
```
davidrapson: Mildly wary of this becoming a dumping ground file. Maybe worth splitting this file up at some point.

test/specs/utilities.test.js
davidrapson: Also important question, is this a green fridge for orangutan or a fridge for _green orangutan_?

PR #285, opened 2017-10-18
Initial pass of improving print css

davidrapson:
Pass of a print stylesheet, using the WIP about page as a test:

**Before**
screen shot 2017-10-18 at 10 44 13

**After**
screen shot 2017-10-18 at 10 43 47

Continues with the rough approach that had already been started

- Adds a common print stylesheet inline with some global resets
- All other print styles are inlined in components
- Adds a `u-dont-print` utility for contextually opting parts out of printing

(Highlighted the fact that the miniature hero styles could definitely stand to be refactored ahead of any about page work so I might try and do that next)

### DIFF
--- a/assets/sass/_print.common.scss
+++ b/assets/sass/_print.common.scss
@@ -1,0 +1,56 @@
+/* =========================================================================
+   Common Print Styles
+   - Component specific print styles inlined in component files
+   ========================================================================= */
+
+@page {
+    margin: 0.5in;
+}
+
+@media print {
+    *,
+    *:before,
+    *:after {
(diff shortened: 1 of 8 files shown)


assets/sass/_print.common.scss
```diff
@@ -0,0 +1,56 @@
...
+        text-shadow: none !important;
+    }
+
+    a,
+    a:visited {
+        text-decoration: underline;
+    }
+
+    a[href]:after {
+        content: ' (' attr(href) ')';
```
mattandrews: Really nice 👍

PR #288, opened 2017-10-19
Add e-bulletin signups

mattandrews:
Using the newly-discovered Dotmailer API, this lets us automate signups and reinstate the original form. Hooray!

This tracks successful signups via (anonymous) Google Analytics events, and logs failed ones to Sentry. Thanks @davidrapson for both those suggestions.

I've verified this working on Wired's marketing tool but worth someone else having a go just to double check.


controllers/toplevel/index.js
```diff
@@ -6,13 +6,17 @@ const rp = require('request-promise');
 const ab = require('express-ab');
 const { sortBy } = require('lodash');
 const xss = require('xss');
+const Raven = require('raven');
+const _ = require('lodash');
```
davidrapson: I don't see lodash being used anywhere in this file?

controllers/toplevel/index.js
```diff
@@ -180,16 +184,14 @@ module.exports = (pages, sectionPath, sectionId) => {
                     res.redirect('/#' + config.get('anchors.ebulletin'));
                 });
             } else {
-                // convert location into proper field value
-                let location = xss(req.body['location']);
-                req.body[location] = 'yes';
-                delete req.body['location'];
-
+                let newsletterLocation = req.body['location'];
```
davidrapson: `req.body.location` instead of `req.body['location']`?

controllers/toplevel/index.js
```diff
@@ -180,16 +184,14 @@ module.exports = (pages, sectionPath, sectionId) => {
...
                 let locale = req.body.locale;
                 let localePrefix = locale === 'cy' ? config.get('i18n.urlPrefix.cy') : '';
 
                 // redirect errors back to the homepage
-                let handleSignupError = () => {
+                let handleSignupError = errMsg => {
+                    // pass error to Sentry
```
davidrapson: This comment feels like it could go

controllers/toplevel/index.js
mattandrews: Ah yeah, was using it and forgot to remove it. Normally my IDE alerts me to unused variables, but in this case, the variable name `_` made the error underline quite difficult to spot...
image

davidrapson:
🙏  Working for me, LGTM

PR #292, opened 2017-10-24
Remove Cumbria funding manager job

mattandrews:


mattandrews:
@davidrapson can you give this another quick sanity check for me?

PR #300, opened 2017-10-25
[Experimental] Form modelling

davidrapson:
This is an exploration of modelling forms to answer some of the questions posed in https://github.com/biglotteryfund/blf-alpha/issues/295

forms

**Structure**

The forms are modelled in the main parts

- The router, which handles boilerplate code involved in rendering each step, displaying errors, and persisting progress
- Some plumbing code in `form-model` which helps us model individual forms 
- `example-form.js` which is the main business logic, defines a form as a series of steps with a schema for individual fields 

**Goals**

- Aims to validate if how easy it is to model multi-page forms with conditional fields without needing to hand-craft individual fields (i.e. can we model a form schema and add fields as requirements change over time)
- Should achieve comprehensive validation for each field
- Validates if the above can be done using server side code, with client-side code only use for progressive enhancement

**Caveats**

This change involves updating to the latest version of `express-validator` as it has a much more flexible API which enables this approach. This means the **tests will fail** as the materials form still depends on the old API.

PR #301, opened 2017-10-26
Upgrade express validator

davidrapson:
This is a bit of yak shaving :why_is_there_not_a_yak_emoji: 🐮  from https://github.com/biglotteryfund/blf-alpha/pull/300 but worth doing. Upgrades `express-validator` to v4.

Forms affected are:

- Newsletter signup
- Materials form
- News Editor

There should be no changes in behaviour due to this upgrade, but as the API is quite different the are a fair few code changes so I'd appreciate a second pair of 👀  on this.

(Best viewed with https://github.com/biglotteryfund/blf-alpha/pull/301/files?w=1)

### DIFF
--- a/controllers/toplevel/index.js
+++ b/controllers/toplevel/index.js
@@ -1,12 +1,14 @@
 'use strict';
+const Raven = require('raven');
 const express = require('express');
 const config = require('config');
-const router = express.Router();
+const { sortBy } = require('lodash');
 const rp = require('request-promise');
 const ab = require('express-ab');
-const { sortBy } = require('lodash');
+const { body, validationResult } = require('express-validator/check');
 const xss = require('xss');
-const Raven = require('raven');
+
+const router = express.Router();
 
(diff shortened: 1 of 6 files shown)


controllers/toplevel/index.js
```diff
@@ -164,20 +166,35 @@ module.exports = (pages, sectionPath, sectionId) => {
...
+        [
+            body('firstName', 'Please provide your first name')
+                .exists()
+                .not()
+                .isEmpty(),
+            body('lastName', 'Please provide your last name')
+                .exists()
+                .not()
+                .isEmpty(),
+            body('email', 'Please provide your email address')
```
mattandrews: Could possibly add the valid email check here, too?

davidrapson:
Added email type validation to the ebulletin form and tested the materials form using my email address.

mattandrews:
Perfect. Shave that yak.

PR #303, opened 2017-10-27
Authentication

mattandrews:
There's a lot here, so apologies in advance. Also a lot of commits representing the various states of the feature (eg. using the `activator` module, then the npm dependency hell, then ripping it out and doing it from scratch).

There's a fair amount of boilerplate in here, mostly around error handling and verification (eg. did the user really request that change?), which means callback hell. Open to suggestions on improving it, hopefully breaking each distinct feature set (registration, login, password reset) into separate controllers helps a little.

There aren't any tests yet: I'm writing these now, but wanted @davidrapson's eyes on what I've done so far.

Things to note before testing locally: you'll need to re-sync the `users` table in your database if you're running this with local MySQL (which you're not by default, so if not, it's already done). You'll also need to fetch the latest app secrets to get the JWT signing key. Also note that emails probably won't send to your BLF email account due to internal blacklisting – not crucial for launch, but will need fixing before eventual go-live.

Have at it!


controllers/user/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+const routeStatic = require('../utils/routeStatic');
+const auth = require('../../modules/authed');
+const register = require('./register');
+const login = require('./login');
+const password = require('./password');
+const { userBasePath, userEndpoints, makeUserLink, emailPasswordValidations, formValidations } = require('./utils');
+
+// serve a logged-in user's dashboard
+routeStatic.injectUrlRequest(router, userEndpoints.dashboard);
```
davidrapson: More a question for another day but should `injectUrlRequest` live under a different namespace than `routeStatic`? This confused me a little as this step doesn't seem to have much to do with static pages. Do all routes need to call this `injectUrlRequest` method?

controllers/user/login.js
```diff
@@ -0,0 +1,48 @@
+const passport = require('passport');
+const { makeErrorList, makeUserLink } = require('./utils');
+
+// try to validate a user's login request
+// @TODO consider rate limiting?
```
davidrapson: Suggestion: Depending on the level at which you'd want to do this at [nginx has pretty good in built rate-limiting](http://nginx.org/en/docs/http/ngx_http_limit_req_module.html) and AFAIK WAF can do this kind of thing too I'd probably err towards WAF > NGINX > Application in terms of preference.

controllers/user/login.js
```diff
@@ -0,0 +1,48 @@
...
+                    // user not valid, send them to login again
+                    req.flash('formValues', req.body);
+                    req.flash('formErrors', makeErrorList(info.message));
+                    req.session.save(() => {
+                        return res.redirect(makeUserLink('login'));
+                    });
+                } else {
+                    // user is valid, send them on
+                    // we don't use flash here because it gets unset in the GET route above
+                    // @TODO is this still true?
```
davidrapson: We're post-truth Matt, don't worry about checking.

controllers/user/password.js
```diff
@@ -0,0 +1,240 @@
...
+    res.cacheControl = { maxAge: 0 };
+    let token = req.query.token;
+    if (!token) {
+        return res.redirect(userBasePath + userEndpoints.login);
+    }
+
+    // a user has followed a reset link
+    jwt.verify(token, secrets['user.jwt.secret'], (err, decoded) => {
+        if (err) {
+            console.error('Password reset token expired', err);
```
davidrapson: Do these console logs still need to be in here? If you are interesting in tracking this we could send an info level event to Sentry instead?

controllers/user/utils.js
```diff
@@ -0,0 +1,73 @@
...
+        ];
+    }
+};
+
+// generic function to return the user to the form with an error
+const renderUserError = (msg, req, res, path) => {
+    if (msg) {
+        req.flash('formErrors', makeErrorList(msg));
+    }
+    req.session.save(() => {
```
davidrapson: Why is this saving the session here? Particularly in a utils function I'm wary of something so stateful.

controllers/user/password.js
```diff
@@ -0,0 +1,240 @@
...
+                        return res.render('user/resetpassword', {
+                            mode: 'pickNewPassword',
+                            token: token,
+                            makeUserLink: makeUserLink
+                        });
+                    },
+                    error => {
+                        console.error('User attempted to reset a password for an non-resettable user', error || {});
+                        req.flash('genericError', 'Your password reset link was invalid - please try again');
+                        return renderUserError(null, req, res, makeUserLink('resetpassword'));
```
davidrapson: 👃  A little wary of functions that start with null as an argument…should this take an object instead?

controllers/user/password.js
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+    });
+};
+
+const sendResetEmail = (req, res) => {
+    // the user wants to trigger a reset email
+    const email = xss(req.body.username);
+    if (!email) {
+        req.flash('genericError', 'Please provide a valid email address');
```
davidrapson: This could probably still be modelled using express-validator

controllers/user/password.js
```diff
@@ -0,0 +1,240 @@
...
+            req.flash('formErrors', errors.array());
+            req.session.save(() => {
+                res.redirect(currentUrl);
+            });
+        } else {
+            // check the token again
+            jwt.verify(changePasswordToken, secrets['user.jwt.secret'], (err, decoded) => {
+                if (err) {
+                    console.error('Password reset token expired', err);
+                    req.flash('genericError', 'Your password reset period has expired - please try again');
```
davidrapson: This is a general comment that I'm unsure about the use of flash messages for this (and in general). Is it an option to pass errors from `const errors = validationResult(req)` directly into the template? I feel I might be oversimplifying though.

controllers/user/utils.js
```diff
@@ -0,0 +1,73 @@
...
+        return res.redirect(path);
+    });
+};
+
+// configure form validation
+const PASSWORD_MIN_LENGTH = 8;
+const formValidations = {
```
davidrapson: 👍  Nice

models/index.js
```diff
@@ -7,15 +7,15 @@ let db = {};
 
 // try getting credentials from env vars (eg. for travis)
 let dbCredentials = {
-    host: secrets['mysql.host'] || process.env.mysqlHost,
-    user: secrets['mysql.user'] || process.env.mysqlUser,
-    pass: secrets['mysql.password'] || process.env.mysqlPassword
+    host: process.env.mysqlHost || secrets['mysql.host'],
```
davidrapson: 🙈

modules/boilerplate/auth.js
```diff
@@ -3,29 +3,38 @@ const passport = require('passport');
...
-            return done(err);
-        });
-    }
-));
+passport.use(
+    new LocalStrategy((username, password, done) => {
+        models.Users
+            .findOne({ where: { username: username } })
+            .then(user => {
+                // use generic error messages here to avoid exposing existing accounts
```
davidrapson: 👍

controllers/user/index.js
mattandrews: Yeah, this is a bit ugly isn't it. Definite code smell.

Basically, templates that need access to `request` data (eg. for form input values or flash data) have it injected via `routeStatic` – there was a nasty race condition when I tried just setting an app-wide global called `request` (eg. opening lots of pages at once saw them override one another). Non-static routes (like these ones) don't have that variable automatically, so this method injects it.

But yeah, it feels a bit broken doesn't it. Will have a think...

controllers/user/login.js
mattandrews: Yeah, this is perhaps a post-launch problem.

controllers/user/password.js
mattandrews: Hmm yeah, good point – will swap these out.

PR #305, opened 2017-10-27
Expose new getSecret method, throw error if unknown secret is requested

davidrapson:
Updates secrets fetching to be behind a `getSecrets` method rather than raw property access so we can add some extra warnings in.


mattandrews:
... so all the tests are failing because they depend on modules which in turn depend on secrets (which Travis doesn't have). We got away with this before with the object property technique as it didn't throw an error in the same way.

Hm. Wonder if we should make it a "soft" error/warning? Not super keen on giving Travis access to the app secrets. We could add them as Travis environment variables? Or make them optional somehow... or explore how running the tests can be done without needing access to secrets – once I swap out live MySQL for the local DB file, that'll help a little.

Friday evenings...

PR #308, opened 2017-10-30
Increase homepage to 50%

davidrapson:
Main question: I've opted to change the name of the test cookie, which will essentially re-bucket people, does that sound reasonable?

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -15,7 +15,7 @@
         "contrast": "contrastMode",
         "session": "blf-alpha-session",
         "csrf": "_csrf",
-        "abTest": "blf-ab-1"
+        "abTest": "blf-ab-1-50"
     },
     "anchors": {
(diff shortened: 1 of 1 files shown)


config/default.json
```diff
@@ -31,8 +30,9 @@
         "enabled": true,
         "tests": {
             "homepage": {
+                "cookieName": "blf-ab-1-50",
```
mattandrews: It's confusing but the reason this cooke's name was defined in the `cookies` array is because `generate-cloudfront-config.js` reads this array out and adds the cookies inside it to each Cloudfront route (eg. to allow those cookies through the load balancer). Either we need to modify that function to look in this location too, or leave the name in the `cookies` array.

davidrapson: Ah gotcha 👌

PR #313, opened 2017-10-30
Use local database for tests

mattandrews:
This cherry picks a bunch of stuff from the `authentication` branch (which will likely cause some merge conflicts, but that's Future Matt's problem) – this should allow `master` to build properly and test without dependencies on external databases.


davidrapson:
giphy

PR #320, opened 2017-11-01
Upgrade to Node 8, latest LTS version

davidrapson:
As Node 8 is now [officially in LTS](https://nodejs.org/en/) it's a good a time as any to upgrade. I've updated the travis and provision file, but not sure what the actual process it to get codedeploy to rebuild the instances. Guessing this doesn't happen during an in-place deploy?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '6'
+- '8'
 cache:
   directories:
     - node_modules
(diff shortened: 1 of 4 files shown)


mattandrews:
Hmm... it should, the `provision.sh` script is in the `AfterInstall` block in the CodeDeploy AppSpec file (eg. after it's copied the new code onto the server). We could merge this and find out... I think it'll work!

davidrapson:
I've been running node 8 locally for a couple of weeks and has been fine.

I've also added the node version into the `appData` global so we can output a temporary debug comment.

```
<!-- deploy DEV (build #DEV) (Node version: v8.6.0) -->
```

mattandrews:
Nice.

davidrapson:
🤞

Issue #275, opened 2017-10-17
Replace funding finder results with new page(s) for users on 10k+ journey


TomSteinberg:
Involves:
- Further CMS work
- Design
- Testing
- Building

ChloeAlper:
Please see below @mattandrews @yasheen @davidrapson @stevekeene @TomSteinberg - a suggestion for the Funding Finder Results page following yesterday's meeting 

![funding-finder-results-20171024](https://user-images.githubusercontent.com/27225981/31949172-9807d946-b8d0-11e7-95d6-6cb80e3c1f3c.png)

mattandrews:
Looks good! Would equal height boxes be easier to scan?

davidrapson:
Aesthetically I really like this, there's definitely something to this in terms of how you are handling information density.

Couple of minor observations:

- I think there are less than 20 active funding programmes at the moment, so we can most likely get away without pagination list everything.
- This might just be me but circular photos feel a little over-stylish at the expense of clarity
- My only concern about the staggered layout is the political-factor of what order things are in (but I do really like the look of it)

👍

ChloeAlper:
Thanks for the feedback @davidrapson and @mattandrews . I'll mock up another version with your comments in mind.

ChloeAlper:
@davidrapson @mattandrews please see design below. Changes made:
• Boxes made equal height
• Photos made square 
• Pagination removed 

![funding-finder-results-20171030](https://user-images.githubusercontent.com/27225981/32167415-b25056cc-bd60-11e7-8a0e-7e3d7fe25016.png)

davidrapson:
👍

TomSteinberg:
For a quick upgrade without having deeply reconsidered the content, this is good and almost ready to go. The only thing that I think is essential to change is that you can't have a the middle > wales > as a link because there's no actual page to link to there. The breadcrumb trail can only have to parts, or if you want three, add the homepage.

stevekeene:
Looks good to me!

davidrapson:
> For a quick upgrade without having deeply reconsidered the content, this is good and almost ready to go. The only thing that I think is essential to change is that you can't have a the middle > wales > as a link because there's no actual page to link to there. The breadcrumb trail can only have to parts, or if you want three, add the homepage.

[...]

PR #324, opened 2017-11-01
Programmes list

davidrapson:
For https://github.com/biglotteryfund/blf-alpha/issues/275

localhost-3000-funding-programmes-min 1000 location wales 1

There's a couple of pieces here that could be pulled out into other PRs. Namely I made a first step at improving the granularity of our sass breakpoints and fixed a semantic issue which are largely tangential to the work here, let me know if you'd prefer that be extracted.

TODO:

- [x] Style programmes list
- [x] Create breadcrumb logic to show active filters, and allow people to return to over10k page
- [x] Update over10k links
- [ ] Update and crop images
- [ ] Translate copy for programmes in CMS based on current funding finder
- [ ] Send UI copy for translation


assets/sass/components/programmes.scss
```diff
@@ -0,0 +1,69 @@
...
+        margin-bottom: $spacingUnit / 2;
+
+        .programme-card__media {
+            order: 1;
+            flex: 0 0 100px;
+            margin-right: $spacingUnit;
+        }
+
+        .programme-card__title {
+            order: 2;
```
davidrapson: Why do you not like this GitHub?

config/locales/cy.json
```diff
@@ -941,6 +947,13 @@
 					}
 				}
 			}
+		},
+		"programmes": {
+			"title": "Funding Programmes",
```
davidrapson: 👀 Translate me

views/layouts/main.njk
```diff
@@ -11,12 +11,11 @@
     <body{% if bodyClass %} class="{{ bodyClass }}"{% endif %}>
         {% include "../includes/accessibility.njk" %}
 
-        {# main content area #}
-        <main>
+        <div class="global-container">
```
davidrapson: This is bit of a semantic fix. This arguably shouldn't be a `main` as

1. The _content_ is really the main area, not the whole page
2. We have `role=main` on all container elements inside page templates so semantically we have two mains

davidrapson:
https://github.com/biglotteryfund/blf-alpha/issues/275

controllers/funding/programmes.js
```diff
@@ -0,0 +1,96 @@
...
+    };
+}
+
+module.exports = function(config) {
+    return function(req, res) {
+        const lang = req.i18n.__(config.lang);
+
+        const locationParamToTranslation = key => {
+            const regions = {
+                england: req.i18n,
```
mattandrews: Is this right?

views/layouts/main.njk
mattandrews: Fair enough!

views/pages/toplevel/over10k.njk
```diff
@@ -33,19 +33,19 @@
 
             <ul class="unstyled grid grid--padded grid--equal grid--4-up grid--wide-only">
                <li class="grid__item">
-                   <a href="{{ buildUrl('funding/funding-finder?area=England&over=10k') }}"
+                   <a href="{{ buildUrl('funding/programmes?location=england&min=1000') }}"
```
mattandrews: This was my code originally, but wondering now if we're happy introducing this new URL format here? On the other hand it'd be quite tricky if we kept the same `funding-finder` path as the existing tool, at least until we change it... hmm.

assets/sass/globals/typography.scss
```diff
@@ -10,8 +10,7 @@ h1,
     font-weight: 600;
     @include font-and-leading(18px, 26px);
 
-    // @TODO: Add more granular named breakpoints
-    @media screen and (min-width: 380px) {
+    @media screen and (min-width: xsmall) {
```
mattandrews: Is this some CSS magic or is it an error? eg. does it know what `xsmall` means?!

davidrapson: Urm… it's sadly an error. A magic error? 💥

views/pages/toplevel/over10k.njk
davidrapson: Yeah, maybe one to discuss. I'm going to roll this change back anyway as I'd like to get this PR out as is, and leaving these URLs in would make it public should we need to deploy…

PR #325, opened 2017-11-01
Surveys [WIP]

mattandrews:
This will still change (likely visually) but probably worth getting your eyeballs on it in its current state @davidrapson.

Quite a lot here really and a few `@TODO`s to clean up by me, but initial thoughts welcome.


modules/boilerplate/getSurveys.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+const _ = require('lodash');
+const models = require('../../models/index');
+
+// @TODO fetching new surveys requires an app restart... do we care?
```
mattandrews: Yes we do care – will change this!

mattandrews: ... would it be better to check for new/updated surveys every x minutes, or to simply make the query on every pageload? Or stick to as-is right now (eg. they're retrieved at app start time)?

davidrapson:
Probably worth chatting through this in person, but a thought:

Should we consider this entirely a progressive enhancement? If so, we could fetch the surveys client-side via a cached endpoint rather than needing to couple the rendering of the app to fetching surveys. Would also bypass the problem of restarting the app to get new surveys or getting into the territory of worker queues to fetch them.

assets/js/modules/surveys.js
```diff
@@ -0,0 +1,135 @@
+'use strict';
+
+const $ = require('jquery');
+
+// via https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API
+let storageAvailable = type => {
```
davidrapson: Maybe pop this in a util file?

mattandrews:
Yeah, good point – so just make an AJAX call on pageload to check if the current page has a survey, that kind of thing?

assets/js/modules/surveys.js
```diff
@@ -0,0 +1,135 @@
+/* global Vue */
+'use strict';
+
+const $ = require('jquery');
+
+// via https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API
+let storageAvailable = type => {
```
davidrapson: Should this go in a utils file?

assets/sass/components/survey.scss
```diff
@@ -0,0 +1,83 @@
...
+            margin-bottom: 0;
+        }
+
+        .survey__submit {
+            flex-grow: 0;
+            margin-bottom: 10px; // fixes weird alignment issue
+        }
+    }
+
+    label {
```
davidrapson: Any reason these elements don't have BEM-style class names for consistency?

assets/sass/utilities/utilities.scss
```diff
@@ -31,6 +31,13 @@
     display: none;
 }
 
+// hide elements for users with JS
+.js-on {
+    .u-hidden-for-js {
```
davidrapson: Does this need to be nested?

controllers/toplevel/index.js
```diff
@@ -296,8 +298,9 @@ module.exports = (pages, sectionPath, sectionId) => {
     // handle contrast shifter
     router.get('/contrast/:mode', (req, res) => {
         res.cacheControl = { maxAge: 1 };
-        let duration = 6 * 30 * 24 * 60 * 60; // 6 months
+
         let cookieName = config.get('cookies.contrast');
+        let duration = moment.duration(6, 'months').asMilliseconds();
```
davidrapson: 👍 Nice

controllers/toplevel/index.js
```diff
@@ -310,6 +313,113 @@ module.exports = (pages, sectionPath, sectionId) => {
         res.redirect(redirectUrl);
     });
 
+    // https://stackoverflow.com/questions/20053949/how-to-get-the-google-analytics-client-id
```
davidrapson: Is this comment connected? 🤔

controllers/toplevel/index.js
```diff
@@ -310,6 +313,113 @@ module.exports = (pages, sectionPath, sectionId) => {
...
+        body('choice')
+            .exists()
+            .not()
+            .isEmpty()
+            .isInt()
+            .withMessage('Please supply a valid choice')
+    ];
+
+    // @TODO add this endpoint to cloudfront
+    router.post('/get-survey', (req, res) => {
```
davidrapson: As discussed, lets make this a cached GET endpoint.

modules/boilerplate/middleware.js
```diff
@@ -21,6 +21,8 @@ const routes = require('../../controllers/routes');
 // load auth strategy
 require('../../modules/boilerplate/auth');
 
+let sessionSecret = getSecret('session.secret') || process.env.sessionSecret;
```
davidrapson: Should this be the other way around, with env first?

test/features/survey.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+let testSurveyData = {
+    name: 'Test Survey',
+    question_en: 'What is the meaning of life?',
+    question_cy: 'Beth yw ystyr bywyd?',
+    activePath: '/foo',
+    active: 1,
+    choices: [
+        {
+            title_en: '42',
```
davidrapson: :neckbeard:

bin/cloudfront/test.json
```diff
@@ -2887,6 +2887,129 @@
                 "Items": [],
                 "Quantity": 0
             },
+            "QueryString": false
+        },
+        "MaxTTL": 31536000,
```
davidrapson: This diffs are genuinely nice 😍

davidrapson:
:shipit:

mattandrews:
I thought about filtering on the client but preferred keeping the URL rewriting (eg. treating `/welsh/contact` and `/contact` as the same thing) on the server-side where it happens elsewhere too.

davidrapson:
"…but preferred keeping the URL rewriting (eg. treating /welsh/contact and /contact as the same thing) on the server-side where it happens elsewhere too."

Good point. 👍

davidrapson:
👍  New commits look 🆗

PR #328, opened 2017-11-02
Programmes translations

davidrapson:
The other side of https://github.com/biglotteryfund/craft-dev/pull/5 🌐 

Add translations for details headings.

screen shot 2017-11-02 at 13 10 14

In two minds as to whether this translation should come from the CMS, but this feels more like a display concern so having it in the app makes most sense for me. 💭 

Also makes a tweak to allow overriding environment variables through the `.env` file locally. Helps with switching out to use local CMS without changing any parameter files.


davidrapson:
Closing in favour of https://github.com/biglotteryfund/blf-alpha/pull/327

PR #329, opened 2017-11-02
Programmes translation tweaks

davidrapson:
Add UI translations. Most of this language was already used on the legacy site. Making the assumption that is accurate but will double check.

Also updates the programme card template to use the `programmeTitle` field rather than the top-level title.

screen shot 2017-11-02 at 15 13 36

### DIFF
--- a/config/locales/cy.json
+++ b/config/locales/cy.json
@@ -951,8 +951,8 @@
 		},
 		"programmes": {
 			"title": "Rhaglenni Ariannu",
-			"under10k": "Rhaglenni ariannu o dan 10k",
-			"over10k": "Rhaglenni ariannu dros 10k",
+			"under10k": "Rhaglenni ariannu o dan £10,000",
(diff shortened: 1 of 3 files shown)


config/locales/cy.json
```diff
@@ -950,11 +950,11 @@
 			}
 		},
 		"programmes": {
-			"title": "Funding Programmes",
-			"under10k": "Funding programmes under 10k",
-			"over10k": "Funding programmes over 10k",
-			"breadcrumbAll": "All Programmes",
-			"breadcrumbLocation": "Funding programmes in %s",
+			"title": "Rhaglenni Ariannu",
+			"under10k": "Rhaglenni ariannu o dan 10k",
```
mattandrews: Do we need a £ sign in here?

davidrapson: I've switched this to use real money (£10,000) rather than the cryptic 10k. Matches format used on https://www.biglotteryfund.org.uk/over10k too.

PR #371, opened 2017-11-09
Funding programme tests and co-locate tests

davidrapson:
Adds some missing tests for the programme page filters.

This PR also restructures tests to allow unit tests to be co-located next to the  module they test


modules/utilities.test.js
```diff
@@ -3,16 +3,16 @@
 const chai = require('chai');
 const expect = chai.expect;
 
-const utilities = require('../../modules/utilities');
+const m = require('./utilities');
```
mattandrews: Why `m`?

davidrapson: Yeah, maybe it doesn't make too much sense here. I don't mind it as a convention for "the thing being tested" but it's a probably a bit silly as it's so terse, will change this back.

PR #376, opened 2017-11-10
Workaround iOS bug with surveys and normalise styling 🐛 🐛🐛

davidrapson:
Fixes iOS font-size scaling (inputs need to be a minimum of 16px otherwise iOS will scale up the viewport to help with entering text, this is an accessibility thing)

Attempts to workaround another exciting bug in iOS 11 where if an text field is in a position fixed container the input cursor will appear offscreen. The fix is…messy…so thoughts welcome.

Normalises markup and styles to be more self contained, removes fragments of `.grid` classes being dotted around.

### DIFF
--- a/assets/js/modules/surveys.js
+++ b/assets/js/modules/surveys.js
@@ -71,12 +71,11 @@ const showSurvey = survey => {
             }
         },
         methods: {
+            localeify: function(obj, field, locale) {
+                return obj[field + '_' + locale];
+            },
             toggleSurvey: function() {
                 this.isActivated = !this.isActivated;
-
-                if (hasiOSBug()) {
-                    $('body').toggleClass('is-ios-editing-survey');
-                }
             },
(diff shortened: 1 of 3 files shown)


assets/js/modules/surveys.js
```diff
@@ -60,6 +73,10 @@ const showSurvey = survey => {
         methods: {
             toggleSurvey: function() {
                 this.isActivated = !this.isActivated;
+
+                if (hasiOSBug()) {
+                    $('body').toggleClass('is-ios-editing-survey');
```
mattandrews: Does this mean that on iOS, anyone who activates the survey can no longer scroll? Might be nicer if this class only gets toggled when they focus on the textbox?

davidrapson:
Good shout, I've scoped the body class to focus/blur events on the message field to avoid accidentally preventing scroll at any point.

PR #399, opened 2017-11-15
Normalise content box styles

davidrapson:
Normalise content-box colours, removes previous tinted version on over/under10k pages and standardises on regular content-box and the tinted one used on EYP. Adds tinted version to styleguide.

### DIFF
--- a/views/pages/toplevel/eyp.njk
+++ b/views/pages/toplevel/eyp.njk
@@ -208,7 +208,7 @@
     {{ hero.header('eyp', heroContent, 'pink', "L'Arche Belfast, Grant: £573,164") }}
 
     <article role="main" class="nudge-up">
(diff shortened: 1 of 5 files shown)


views/pages/toplevel/eyp.njk
```diff
@@ -208,7 +208,7 @@
     {{ hero.header('eyp', heroContent, 'pink', "L'Arche Belfast, Grant: £573,164") }}
 
     <article role="main" class="nudge-up">
-        <div class="inner--wide-only accent--pink a--border-top content-box content-box--pink mob--no-space">
+        <div class="inner--wide-only accent--pink a--border-top content-box content-box--tinted">
```
mattandrews: Good catch on `mob--no-space` too, whatever that was...

PR #400, opened 2017-11-17
RC Prototype

davidrapson:
Initial prototype work. Getting this up for visibility.

### DIFF
--- a/assets/sass/scaffolding/content.scss
+++ b/assets/sass/scaffolding/content.scss
@@ -52,7 +52,7 @@
     ul li,
     ol li {
         margin-left: $spacingUnit;
-        margin-bottom: 0.5em;
+        margin-bottom: 0.3em;
         margin-top: 0;
     }
 
--- a/assets/sass/scaffolding/content.scss
+++ b/assets/sass/scaffolding/content.scss
@@ -34,3 +34,22 @@
         border-left: none;
     }
 }
+
+/* =========================================================================
+   Content: Prose Scope
+   ========================================================================= */
+/*
+    Opt-in typography:
+    http://anthonyshort.me/global-typographic-styles-suck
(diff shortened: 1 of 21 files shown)


assets/sass/scaffolding/content.scss
```diff
@@ -34,3 +34,22 @@
...
+/* =========================================================================
+   Content: Prose Scope
+   ========================================================================= */
+/*
+    Opt-in typography:
+    http://anthonyshort.me/global-typographic-styles-suck
+    http://dbushell.com/2012/04/18/scoping-typography-css/
+    http://css-tricks.com/opt-in-typography/
+*/
+.s-prose {
```
mattandrews: I had a go at something like this with `.has-body-copy`, maybe rename the latter to this if applicable?

views/pages/experimental/apply/form.njk
```diff
@@ -95,21 +95,29 @@
...
-                                    {{ inputText(field) }}
-                                {% endif %}
-                                {% if field.type === 'textarea' %}
-                                    {{ inputTextarea(field) }}
-                                {% endif %}
-                                {% if field.type === 'radio' %}
-                                    {{ inputRadio(field) }}
+                            <div class="form-field__main">
+                                <div class="form-field__body">
+                                    {% if field.type === 'text' %}
```
mattandrews: If we find this pattern coming up a lot maybe we need a macro to do this switching for us? But fine for now.

views/pages/experimental/apply/form.njk
```diff
@@ -143,6 +143,15 @@
                                     </div>
                                 </details>
                             {% endif %}
+
+                             {% if field.canBeDuplicated === true %}
```
mattandrews: Do you need the `=== true` here?

PR #411, opened 2017-11-23
Move nodemon config into package.json, reload on njk files

davidrapson:
🤖

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,12 @@
       ]
     ]
   },
+  "nodemonConfig": {
+    "verbose": true,
+    "ignore": ["assets", "public"],
+    "exec": "bin/www",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -68,6 +68,12 @@
       ]
     ]
   },
+  "nodemonConfig": {
+    "verbose": true,
+    "ignore": ["assets", "public"],
+    "exec": "bin/www",
+    "ext": "js,json,njk"
```
mattandrews: I know this is merged now, but `.njk` files should have already been refreshing on change (eg. no need for the app to reload)?

davidrapson: Hmm, interesting. Will double check.

PR #419, opened 2017-11-24
Improve no-cache and csrf middleware

davidrapson:
Builds on https://github.com/biglotteryfund/blf-alpha/pull/418 to normalise how we handle csrf and no-cache headers.

- Adds a shared middleware to normalise adding no-cache headers in the same way
- Adds a function to the test helpers to get a CSRF token (copied from the materials tests, thanks @mattandrews )
- Updates all routes that have inline `maxAge` to either use `noCache` or `csrfProtection` middleware


integration-tests/materials.test.js
```diff
@@ -28,17 +28,10 @@ describe('Material order form', () => {
...
+    beforeEach(async () => {
         const funding = routes.sections.funding;
         const path = funding.path + funding.pages.freeMaterials.path;
-        agent.get(path).end((err, res) => {
-            // res.should.have.cookie('_csrf');
-            const dom = new JSDOM(res.text);
-            csrfToken = dom.window.document.querySelector('input[name=_csrf]').value;
-            done();
-        });
+        csrfToken = await helper.getCsrfToken(agent, path);
```
mattandrews: 💯

PR #421, opened 2017-11-27
SES update

mattandrews:
This is to aid with my other branch ([`form-success-callback`](https://github.com/biglotteryfund/blf-alpha/tree/form-success-callback)) – updates Nodemailer to use an IAM-powered SES email transport.

This means we can send email from any SES-approved domain we own, and removes the need to pass SES secrets around via config (it just uses the IAM roles for the EC2 instances).

This also adds error tracking via Sentry and returns a Promise for email sending, allowing us to add some error handling for failed sends (if any).

### DIFF
--- a/modules/mail.js
+++ b/modules/mail.js
@@ -4,13 +4,6 @@ const config = require('config');
 const AWS = require('aws-sdk');
 const Raven = require('raven');
 
-// Use local environment AWS credentials in dev mode.
-// Otherwise, EC2 instances already have IAM instance roles
-// with valid ses:SendRawEmail permissions
-if (process.env.NODE_ENV === 'development') {
-    AWS.config.credentials = new AWS.SharedIniFileCredentials({ profile: 'default' });
-}
-
 // create Nodemailer SES transporter
 const transport = nodemailer.createTransport({
     SES: new AWS.SES({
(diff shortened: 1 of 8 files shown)


modules/mail.js
```diff
@@ -1,19 +1,22 @@
...
 
-let mailConfig = {
-    user: getSecret('ses.auth.user'),
-    password: getSecret('ses.auth.password')
-};
-// create reusable transporter object using the default SMTP transport
+// Use local environment AWS credentials in dev mode.
+// Otherwise, EC2 instances already have IAM instance roles
+// with valid ses:SendRawEmail permissions
+if (process.env.NODE_ENV === 'development') {
```
davidrapson: The docs make it sound like this isn't needed:

http://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-node-credentials-shared.html

"The SDK for JavaScript automatically searches the shared credentials file for credentials when loading"

"By default, the SDK checks the AWS_PROFILE environment variable to determine which profile to use. If the AWS_PROFILE variable is not set in your environment, the SDK uses the credentials for the [default] profile."

Does this work without these lines?

mattandrews: It does!

PR #437, opened 2017-11-28
Mirror Awards For All Pages

davidrapson:
Initial work to mirror Awards For All pages.

Tasks to complete:

- [x] Confirm we can reliably determine which content to replace
- [x] Set up A/B test buckets to be used for gradual rollout
- [ ] Handle all versions of URLs for awards for all pages
- [ ] Confirm and add real copy for each region (England, Scotland, Wales + Welsh language)
- [ ] Confirm initial percentage to start rollout at

### DIFF
--- a/controllers/toplevel/legacyPages.js
+++ b/controllers/toplevel/legacyPages.js
@@ -50,16 +50,22 @@ function initAwardsForAll(router) {
         return proxyLegacyPage(req, res, dom => dom);
     }
 
-    function proxyAwardsForAllTemporaryChangeMe(req, res) {
+    function proxyAwardsForAllTemporaryChangeMe(applyUrl, req, res) {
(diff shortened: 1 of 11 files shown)


controllers/toplevel/legacyPages.js
```diff
@@ -0,0 +1,67 @@
...
+            const applyTab = dom.window.document.querySelector('#mainContentContainer .panel:last-of-type');
+            applyTab.innerHTML = `
+            <p>Replacement tab content</p>
+        `;
+            return dom;
+        });
+    }
+
+    router
+        .route('/prog_a4a_eng')
```
mattandrews: This might be a nice opportunity to kill that URL (in Google, anyway) and have it redirect to the other one (like our vanity URLs)?

PR #438, opened 2017-11-29
Material order tracking

mattandrews:
Not yet ready for merge yet (needs confirmation from legal) but should be good to review.

There's a fairly messy script in here (which needs credentials to run) – it'll fetch all the orders from a Gmail account and parse them into a JSON file, then reads back that JSON and inserts it into the new `Order` tables (after parsing them a bit more, ugh). This just means we don't lose order data from the last six months.

This also adds order tracking to newly-created orders.

When we merge/deploy this, we'll need to run the Gmail script against the `production` database, to backfill it (or run it on `dev` and import its contents to `production`).


bin/scripts/gmail.js
```diff
@@ -0,0 +1,282 @@
...
+function camelize(str) {
+    return str
+        .replace(/(?:^\w|[A-Z]|\b\w)/g, function(letter, index) {
+            return index === 0 ? letter.toLowerCase() : letter.toUpperCase();
+        })
+        .replace(/\s+/g, '');
+}
+
+let allOrders = [];
+let numMessagesToFetch = 1000;
```
mattandrews: There are currently 800 or so orders so we should be safe with this for now.

controllers/funding/free-materials/formFields.js
```diff
@@ -5,98 +5,98 @@ module.exports = [
...
-            'en': 'Project ID number',
-            'cy': 'Rhif adnabod eich prosiect'
+            en: 'Project ID number',
+            cy: 'Rhif adnabod eich prosiect'
         }
     },
     {
         name: 'yourGrantAmount',
-        type: 'text',
+        type: 'number',
```
mattandrews: I changed this so that new orders being stored won't need all the number-parsing madness in the Gmail script. In the not-too-distant future we'll change this field to be a dropdown/range selector anyway.

davidrapson:
👍 Changes look ok, you've got a merge conflict though.

davidrapson:
🚢 it

Issue #427, opened 2017-11-27
Change logos on pages that offer Wales-specific products


mattandrews:

mattandrews:
@TomSteinberg do we have a list of the pages for this, or do we need to speak to Dave? Presumably the logos and free materials pages are on the list as they're both offering bilingual "products"?

davidrapson:
Contacted Dave in the first instance.

mattandrews:
This is now done (any programme located in Wales automatically gets the bilingual logo)

PR #439, opened 2017-11-29
Refactor global header

davidrapson:
This is some work tangential to https://github.com/biglotteryfund/blf-alpha/issues/427

Currently the header styles are…complex. So in order to avoid things getting too messy if we need to support more variants of the header for compliance in wales then it's worth simplifying things.

The main focus of this PR is to standardise on _one_ header structure, and remove the forking paths in the current code. This means no changes to new-style headers (e.g. Homepage) but does mean updating pages with the previous header style.

- Normalise header styles so there's only one style of header
- Decouple nav links macros so they can be included individually
- Decouple classes shared between the global navigation and the offscreen navigation
- Simplify how welsh language modifications are applied to the header (now only in one place)

**Before**:

screen shot 2017-11-29 at 14 53 03

**After**:

screen shot 2017-11-29 at 14 53 10

**Overlay header unchanged:**

screen shot 2017-11-29 at 14 53 20


assets/sass/scaffolding/header.scss
```diff
@@ -2,215 +2,223 @@
...
+            margin-top: ($spacingUnit / 2);
         }
-    }
 
-    svg {
-        position: absolute;
-    }
-}
+        @media only screen and (min-width: $widestScreen) {
+            margin-left: -20px;
```
mattandrews: `$spacingUnit`?

assets/sass/scaffolding/header.scss
```diff
@@ -235,16 +269,20 @@ $offscreenNavWidth: 65%;
         fill: palette('soft-grey');
     }
 
-    @include mq('tablet') {
+    @include mq(tablet) {
```
mattandrews: The convention for this seems to be the other way around (69 matches for `mq('tablet')` versus 4 for `mq(tablet)` in my repo) – happy to use it unquoted if that's neater but can we make it one or the other?

assets/sass/scaffolding/header.scss
```diff
@@ -235,16 +269,20 @@ $offscreenNavWidth: 65%;
...
         top: 5px; // to line up with the search input
     }
 }
 
-/**
- * Mobile off-screen navigation
- */
-.nav--offscreen {
+/* =========================================================================
+   Offsreen Navigation
```
mattandrews: *Offscreen!

PR #442, opened 2017-11-30
Updates to how legacy proxied pages are handled

davidrapson:
Extract the parts of https://github.com/biglotteryfund/blf-alpha/pull/437 that are suitable to go live now.

- Adds `live` flag to vanity URLs to allow dark-launching of vanity redirects in the same way we do other URLs
- Extract legacy proxied URLs into their own config, allowing us to share paths around the app in an isolated way rather than duplicating URLs
- Update security headers middleware to share legacy routes config

### DIFF
--- a/middleware/securityHeaders.js
+++ b/middleware/securityHeaders.js
@@ -1,56 +1,67 @@
 'use strict';
 const helmet = require('helmet');
+const { map } = require('lodash');
+const { legacyProxiedRoutes } = require('../controllers/routes');
 
-// these URLs won't get the helmet header protection
-// @TODO this should only affect the legacy homepage
-const pathsExemptFromHelmet = ['/', '/welsh', '/legacy', '/funding/funding-finder', '/welsh/funding/funding-finder'];
+module.exports = function(app) {
+    /**
+     * URLs which should be exempt from security headers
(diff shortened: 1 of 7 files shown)


middleware/securityHeaders.js
```diff
@@ -1,56 +1,67 @@
...
 
-// these URLs won't get the helmet header protection
-// @TODO this should only affect the legacy homepage
-const pathsExemptFromHelmet = ['/', '/welsh', '/legacy', '/funding/funding-finder', '/welsh/funding/funding-finder'];
+module.exports = function(app) {
+    /**
+     * URLs which should be exempt from security headers
+     * Only proxied legacy URLs should be exempt.
+     */
+    const exemptLegacyUrls = map(legacyProxiedRoutes, _ => _.path);
```
mattandrews: Should this "welshify" these URLs too?

davidrapson: Noting down what we discussed that, no it doesn't need to welshify as we're using legacy urls as a per-URL opt-in so preferable to not double up urls automatically. Also, worked out that the `/funding/funding-finder` proxy code can go as it's been superseded by `/funding/programmes`

PR #446, opened 2017-11-30
[WIP] Mirror Awards For All Pages

davidrapson:
Initial work to mirror Awards For All pages. Re-do of #437 

Tasks to complete:

- [x] Confirm we can reliably determine which content to replace
- [x] Set up A/B test buckets to be used for gradual rollout
- [x] Handle all versions of URLs for awards for all pages
- [ ] Confirm and add real copy for each region (England, Scotland, Wales + Welsh language)
- [ ] Confirm initial percentage to start rollout at


config/default.json
```diff
@@ -16,7 +16,8 @@
         "contrast": "contrastMode",
         "session": "blf-alpha-session",
         "csrf": "_csrf",
-        "abTest": "blf-ab-1-50"
+        "abTestHomepage": "blf-ab-1-50",
+        "abTestAwardsForAll": "blf-ab-afa"
```
mattandrews: Worth putting an index on this too to reflect the chances of it being renamed/increased at some point?

controllers/toplevel/legacyPages.js
```diff
@@ -0,0 +1,97 @@
...
+
+            const additionalText = `
+                <h4>Apply online</h4>
+                <p><br/><a class="roundedButton blueButton" style="display: inline-block;" href="${
+                    applyUrl
+                }">Apply Online</a></p>
+                <h4>Apply by post</h4>
+            `;
+
+            applyTab.innerHTML = additionalText + applyTab.innerHTML;
```
mattandrews: Is it worth doing a sanity check here to verify we made a change or found the right DOM elements and reporting to Sentry if not? Just to catch the edge case where these pages get modified...

config/default.json
davidrapson: Not sure I follow?

controllers/toplevel/legacyPages.js
davidrapson: Good shout

PR #445, opened 2017-11-30
Remove funding finder proxy

davidrapson:
💡 Whilst working on https://github.com/biglotteryfund/blf-alpha/pull/437 I realised that we no longer need the funding-finder proxy code as that was for links from `/over10k` which have been superseded by `/funding/programmes`.

😢 Sorry to see @mattandrews's code go…but 🗑 away it goes.

- Remove funding finder proxy code (and delete some tests 🙈)
- Remove (now) unused `parseValueFromString`
- Move leftover `stripTrailingSlashes` function into `module/urls` alongside other related code

Issue #318, opened 2017-10-31
Redirect every funding finder results page to a page hosted on the new CMS


TomSteinberg:
Redirect every funding finder results page to a page hosted on the new CMS

- https://www.biglotteryfund.org.uk/funding/funding-finder
- https://www.biglotteryfund.org.uk/welsh/funding/funding-finder
- https://www.biglotteryfund.org.uk/Home/Funding/Funding%20Finder
- https://www.biglotteryfund.org.uk/home/funding/funding-finder

Redirect to: https://www.biglotteryfund.org.uk/funding/programmes

## Discussions to have

- [x] Talk to Nic
- [x] Talk to Engagement

## Development Tasks

- [x] Review analytics data to establish primary URLs funding finder is accessed from
- [x] Add redirects for common URL params to equivalents on new page.
- [x] Redirect everything else to full programmes list
- [x] Handle normalising legacy query strings, notably `&amp;` in URLs.
- [x]  Handle closed programmes
- [x] Test in live-equivalent environment to confirm redirects work as expected
- [x] Deploy redirects to test environment

## Launch tasks  

- [x] Set routes for `/Home/Funding/Funding%20Finder` redirects to `live`. Deploy and test redirects.
- [x] Set routes for `/funding/funding-finder` redirects to `live`. Deploy and test redirects.
- [x] Monitor redirects for any 404s

davidrapson:
Initial experiments around doing this here: https://github.com/biglotteryfund/blf-alpha/pull/448

PR #448, opened 2017-11-30
Funding finder redirects

davidrapson:
Now that we've removed the funding finder proxy code in https://github.com/biglotteryfund/blf-alpha/pull/445 it allows us to…re add that URL again. This time to experiment with https://github.com/biglotteryfund/blf-alpha/issues/318

Sets up some redirects for:

- Location. England, scotland, northern-ireland, wales all get redirected to the equivalent query string 
- If the old `amount` paramater is for anything over 10k we add `min=10000` otherwise we do nothing

That's it. Keeping it simple for now.

Based on a discussion with @TomSteinberg , the small number of programmes we have, and that we now show them all on one page there's not much value in handling more than the main sets of filters people use.

Slightly unsure what happens when someone posts to the funding finder from another page so that'd be the main thing to rule out before doing anything real with this. That and talking to Engagement.

### DIFF
--- a/controllers/funding/index.js
+++ b/controllers/funding/index.js
@@ -1,10 +1,11 @@
 'use strict';
+const config = require('config');
 const express = require('express');
 const router = express.Router();
 const moment = require('moment');
 const _ = require('lodash');
 const xss = require('xss');
-const config = require('config');
+const queryString = require('query-string');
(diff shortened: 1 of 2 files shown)


davidrapson:
Closing this but keeping the branch for now, until we're ready to do something with this.

PR #460, opened 2017-12-06
Don't auto-stage files after running prettier

davidrapson:
The fact that `lint-staged` can't handle partially staged changes has been nagging at me. There's an ongoing issue where where it looks like they are working towards an approach https://github.com/okonet/lint-staged/issues/62 but in the meantime it feels like the least-magic thing to do is to not re-add files and only format.

Not quite as invisible, but I'm erring towards patching in partial changes being more important than some auto-formatting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,19 +24,16 @@
     "prepush": "npm test"
   },
   "lint-staged": {
+    "assets/**/*.scss": [
+      "prettier --write"
+    ],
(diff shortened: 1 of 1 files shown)


mattandrews:
Think this makes sense – on complex commits recently I've taken to disabling commit hooks to get around this, which isn't ideal. 👍

Issue #467, opened 2017-12-08
Deprecation warning from Sequelize


davidrapson:
sequelize deprecated String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators node_modules/sequelize/lib/sequelize.js:236:13

davidrapson:
Closing as resolved in https://github.com/biglotteryfund/blf-alpha/pull/469

PR #469, opened 2017-12-11
Model database queries in services

davidrapson:
This helps with https://github.com/biglotteryfund/blf-alpha/issues/467

- Centralise related database calls into dedicated files (e.g, all user queries) to avoid spreading sequelize syntax around the app
- Updates `where` clauses to use `Op` calls
- Disables `operatorsAliases`

Deprecation warning is gone and tests still pass. Note: The one part I've not updated to this style is the material orders.

The Sequelize docs aren't particularly well written here. So I'm not 100% sure I follow 
exactly to what extent the `Op` calls are needed, but I think it makes sense. http://docs.sequelizejs.com/manual/tutorial/querying.html#operators-security

### DIFF
--- a/services/surveys.js
+++ b/services/surveys.js
@@ -1,0 +1,64 @@
+const xss = require('xss');
+const { Survey, SurveyChoice, SurveyResponse } = require('../models');
+
+function findAll() {
+    return Survey.findAll({
+        include: [
+            {
+                model: SurveyChoice,
+                as: 'choices',
+                required: true,
+                include: [
+                    {
+                        model: SurveyResponse,
+                        as: 'responses',
+                        required: true
+                    }
+                ]
+            }
+        ]
+    });
+}
+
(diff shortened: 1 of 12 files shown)


services/surveys.js
```diff
@@ -0,0 +1,82 @@
...
+        include: [
+            {
+                model: SurveyChoice,
+                as: 'choices'
+            }
+        ]
+    });
+}
+
+function createResponse(submissionBody) {
```
mattandrews: Not 100% sure but I think I'd prefer most of this function to stay in the controller and just have the`SurveyResponse.create` functionality live here – think it's a bit confusing having the stuff dealing with form data / validation mixed in with the model stuff (kind of the inverse of the point of this change, right?).

davidrapson: Yeah, I think you're right.

mattandrews:
👍

PR #473, opened 2017-12-11
Disable livechat, resolve proxy rendering issue

davidrapson:
The live chat widget contains a document.write call which causes rendering issues when we proxy the page. Removing the element containing the widget is the most direct short-term way of resolving the issue. Waiting on approval before merging.


config/app/server-dev.conf
```diff
@@ -0,0 +1,42 @@
+# server {
```
mattandrews: Do we need this file? Might be quite neat to have it around but needs the keys too I guess?

davidrapson: It's any old self-signed key, but I can remove this and just keep a note for myself. There's not much to this config anyway.

PR #478, opened 2017-12-12
Launch homepage

mattandrews:
Removes the A/B test functionality for the homepage routes, removes the `/legacy` testing path, and redirects `/home` to `/`.

### DIFF
--- a/config/cloudfront/live.json
+++ b/config/cloudfront/live.json
@@ -26,10 +26,9 @@
                         "contrastMode",
                         "blf-alpha-session",
                         "_csrf",
-                        "blf-ab-1-50",
                         "blf-ab-afa"
                     ],
-                    "Quantity": 5
+                    "Quantity": 4
                 }
             },
             "QueryStringCacheKeys": {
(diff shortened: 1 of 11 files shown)


config/cloudfront/live.json
```diff
@@ -5651,10 +5558,68 @@
...
+                }
+            },
+            "QueryStringCacheKeys": {
+                "Items": [],
+                "Quantity": 0
+            },
+            "QueryString": false
+        },
+        "MaxTTL": 31536000,
+        "PathPattern": "/welsh/home",
```
davidrapson: How come this is getting added?

controllers/routes.js
```diff
@@ -42,7 +42,8 @@ const routes = {
                     code: 0,
                     static: false,
                     live: true,
-                    isPostable: true
+                    isPostable: true,
+                    aliases: ['/home']
```
davidrapson: Is this to mop up any traffic that might have linked to this?

config/cloudfront/live.json
mattandrews: Any alias listed for a route automatically gets a Welsh version added too. Kind of redundant in this case but causes no harm really.

davidrapson: 👍

controllers/routes.js
mattandrews: Yeah, you see this path in analytics (so presumably in google etc) so making it properly redirect would be nice.

davidrapson:
🚀

PR #479, opened 2017-12-12
Programme tweaks

davidrapson:
Adds placeholder images for AFA and Reaching Communities pages. For expediency I've mapped the images based on the path, rather than modelling in the CMS (which needs some consideration). I don't _hate_ this as an approach.

Also adds a couple of (test-only) vanity redirects.


views/pages/funding/programme-detail.njk
```diff
@@ -6,15 +6,9 @@
...
-{% set heroImage = createHeroImage({
-    small: 'hero/working-families-small.jpg',
-    medium: 'hero/working-families-medium.jpg',
-    large: 'hero/working-families-large.jpg',
-    default: 'hero/working-families-medium.jpg'
-}) %}
-{% set socialImage = heroImage %}
 
 {% block content %}
+    {% set socialImage = heroImage %}
```
mattandrews: Does this work here? I _think_ this needs to be defined outside/before the `content` block.

davidrapson: Good catch

PR #468, opened 2017-12-08
Add HTML parser to scraper

mattandrews:
Still more to do here but this adds basic HTML cleaning/storage features to the scraper.

PR #480, opened 2017-12-12
Add legacy page support

mattandrews:
Open to discussion about the paradigm here.

Essentially, the workflow for importing legacy pages will be:

- Run the scraper (see https://github.com/biglotteryfund/blf-alpha/pull/468)
- Add its cleaned HTML output to the CMS*, specifying a `legacyPath` for the content
- Add the legacy path to the routes file

(this is more of a manual process than the funding programme tool as the URLs we're exploring importing for this aren't in any specific section).

* this could be via automated means, or done manually as I've been doing for now.

### DIFF
--- a/controllers/routes.js
+++ b/controllers/routes.js
@@ -280,6 +280,12 @@ const routes = {
                         // sectionPaths.aboutLegacy + "/customer-service/data-protection",
                         '/data-protection'
                     ]
+                },
+                peopleInTheLead: {
+                    name: 'People in the Lead',
+                    path: '/helping-millions-change-their-lives',
+                    cms: true,
+                    live: false
                 }
             }
         }
--- a/controllers/routes.js
+++ b/controllers/routes.js
@@ -284,7 +284,7 @@ const routes = {
                 peopleInTheLead: {
                     name: 'People in the Lead',
                     path: '/helping-millions-change-their-lives',
-                    cms: true,
+                    isLegacyPage: true,
                     live: false
                 }
             }
--- a/controllers/utils/routeStatic.js
+++ b/controllers/utils/routeStatic.js
@@ -1,5 +1,10 @@
 'use strict';
 const app = require('../../server');
+const contentApi = require('../../services/content-api');
+const { renderNotFound } = require('../http-errors');
+
+const { get } = require('lodash');
+const Raven = require('raven');
 
 // redirect any aliases to the canonical path
 let setupRedirects = (sectionPath, page) => {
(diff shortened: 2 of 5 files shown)


controllers/routes.js
```diff
@@ -280,6 +280,12 @@ const routes = {
                         // sectionPaths.aboutLegacy + "/customer-service/data-protection",
                         '/data-protection'
                     ]
+                },
+                peopleInTheLead: {
+                    name: 'People in the Lead',
+                    path: '/helping-millions-change-their-lives',
+                    cms: true,
```
davidrapson: Could we call this `isLegacyPage`, just `cms` seems ambiguous

controllers/utils/routeStatic.js
```diff
@@ -28,6 +33,28 @@ let servePage = (page, router) => {
...
+        const renderPage = content => {
+            res.render('pages/legacy', {
+                title: content.title,
+                content: content
+            });
+        };
+
+        contentApi
+            .getLegacyPage(req.i18n.getLocale(), sectionId + page.path)
+            .then(response => get(response, 'data.attributes', []))
```
davidrapson: I've been moving this kind of stuff into the content API method, and being explicit about params, so you'd end up with something like:

```
contentApi
  .getLegacyPage({
    locale: req.i18n.getLocale(),
    path: sectionId + page.path
  })
  .then(content => {
    renderPage(content);
  });
```

views/pages/legacy.njk
```diff
@@ -0,0 +1,23 @@
...
+
+{% block content %}
+
+    {% set heroContent %}
+        {{ headers.overlayText('h2', title, 't2') }}
+    {% endset %}
+
+    {{ hero.header('foi', heroContent, 'turquoise') }}
+
+    <article role="main" class="nudge-up s-prose">
```
davidrapson: Try adding `s-prose--full` too as that adds some more details typographic tweaks (in retrospect perhaps these classes should be merged)

middleware/securityHeaders.js
```diff
@@ -25,7 +25,8 @@ module.exports = function(environment) {
         '*.youtube.com',
         'cdn.jsdelivr.net',
         'sentry.io',
-        'dvmwjbtfsnnp0.cloudfront.net'
+        'dvmwjbtfsnnp0.cloudfront.net',
+        '*.biglotteryfund.org.uk'
```
davidrapson: Meta

controllers/routes.js
mattandrews: I guess I was trying to think of a future environment where most/all pages are CMS driven and this flag can tell the router to just pass the request/path onto the Content API. I guess for now though this is specifically for legacy pages so that works.

views/pages/legacy.njk
mattandrews: Good call, the template here is pretty much a proof-of-concept.

PR #483, opened 2017-12-13
Add programme page redirect handlers

davidrapson:
Tried a few different approaches for handling redirects from old => new programme page urls and have landed on something that feels reasonably concise.

Rather than adding loads of vanity redirects it feels like programme pages are discreet enough to warrant their own config. Focuses on handling the 80% case of redirecting `/global-content/programmes` to `/funding/programmes` equivalent + welsh versions. Short urls like `/prog_afa_england` etc. would still go in vanity urls.

I've excluded Awards For All for now to avoid conflicting with testing copy for the active A/B setup, but they should go in here too eventually.

(Also enabled `no-shadow` eslint rules as I noticed a lot of variable names were getting reused in different scopes in the routes code, e.g. `path`)

Best viewed with https://github.com/biglotteryfund/blf-alpha/pull/483/files?w=1

### DIFF
--- a/controllers/routes.js
+++ b/controllers/routes.js
@@ -4,7 +4,7 @@ const anchors = config.get('anchors');
 
 // pass some parameters onto each controller
 // so we can take route config and init all at once
-let loadController = path => {
+const loadController = path => {
     return (pages, sectionPath, sectionId) => require(path)(pages, sectionPath, sectionId);
 };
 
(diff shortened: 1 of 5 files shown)


controllers/routes.js
```diff
@@ -286,13 +286,48 @@ const routes = {
...
+        isPostable: false,
+        allowQueryStrings: false,
+        live: !isLive ? false : true
+    };
+}
+const programmeRedirects = [
+    programmeMigration('england/reaching-communities-england', 'reaching-communities-england', false),
+    programmeMigration('england/parks-for-people', 'parks-for-people', false),
+    programmeMigration('northern-ireland/people-and-communities', 'people-and-communities', false),
+    programmeMigration('wales/people-and-places-medium-grants', 'people-and-places-medium', false),
```
mattandrews: Maybe change `people-and-places-medium` to `people-and-places-medium-grants` to match the format of the `large` one?

PR #485, opened 2017-12-13
Display an admin panel with edit link on test site

davidrapson:
This is somewhat speculative change. This PR adds a small edit link to funding programmes that come from the CMS.

It raises a few questions around permissions. For now I've added a check to make sure we're in a test environment (there's a corresponding change on the API end with the same check). Not sure if outright displaying this on the test site is OK and relying on Craft's authentication is enough for now. We'll likely need something more comprehensive at some point though.

screen shot 2017-12-13 at 16 02 37


mattandrews:
> For now I've added a check to make sure we're in a test environment

Not seeing this...

mattandrews:
I like where this is going – what I'm struggling with is the implicit linking between environments (eg. users of the `TEST` site will be directed to edit content on the `LIVE` CMS, right?).

I have a working branch of our Craft app which supports retrieving draft content. The issue with integrating that is authentication: how can we ensure the user has permission to see draft stuff? Maybe we need to look more into sharing cookies/auth between Craft/this frontend app, or perhaps detecting user IP ranges and anyone on a BLF network gets to see draft content and edit links?

davidrapson:
> Not seeing this...

Urm, I imagined it apparently! But try again.

PR #486, opened 2017-12-14
Display an admin panel with edit link if authed

davidrapson:
Re-do of #485 this time by requesting an authenticated endpoint from the CMS to get details about an entry. Requires configuring cookie sharing. Although the real need for this is on a non-test domain the `test.blf.digital` domain is in a position where we could trial this.

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -24,6 +24,7 @@ require('./modules/heroImages').init();
 require('./modules/logos').init();
 require('./modules/materials').init();
 require('./modules/forms').init();
+require('./modules/adminPanel').init();
 
 /**
  * If we are in the live environment then load analytics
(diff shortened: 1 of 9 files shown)


assets/js/main.js
```diff
@@ -25,6 +25,25 @@ require('./modules/logos').init();
 require('./modules/materials').init();
 require('./modules/forms').init();
 
+console.log(window.AppConfig);
```
mattandrews: Is this meant to be here?

assets/js/main.js
```diff
@@ -25,6 +25,25 @@ require('./modules/logos').init();
...
+if (window.AppConfig.adminEndpoint) {
+    fetch(window.AppConfig.adminEndpoint, { credentials: 'include' })
```
mattandrews: Worth checking for `fetch` support here (eg. not in IE11) before making this call?

PR #490, opened 2017-12-14
Support randomising normal heroes, for legacy pages

davidrapson:
Adds a set of images to rotate through on legacy pages.

### DIFF
--- a/controllers/utils/routeStatic.js
+++ b/controllers/utils/routeStatic.js
@@ -1,6 +1,7 @@
 'use strict';
 const app = require('../../server');
 const contentApi = require('../../services/content-api');
+const { createHeroImage } = require('../../modules/images');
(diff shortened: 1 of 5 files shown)


controllers/utils/routeStatic.js
```diff
@@ -34,20 +35,54 @@ let servePage = (page, router) => {
...
     router.get(page.path, (req, res) => {
-        const renderPage = content => {
-            res.render('pages/legacy', {
-                title: content.title,
-                content: content
-            });
-        };
+        /**
+         * Placeholder images
+         * @TODO: Replace with different ones, or ideally fetch from CMS.
```
mattandrews: Not super keen on this living here – can we put this in its own config file as part of this PR?

PR #494, opened 2017-12-22
[Pending] Add listing pages

mattandrews:
Work in progress, but illustrative of what we can do here.

DO NOT MERGE until the `PRODUCTION` CMS has the content otherwise this page will break.


services/content-api.js
```diff
@@ -52,9 +52,22 @@ function getLegacyPage({ locale, path }) {
...
 
+function getListingPage({ locale, path }) {
+    return request({
+        url: `${API_URL}/v1/${locale}/listing`,
+        qs: {
+            path: path
+        },
+        json: true
+    }).then(response => {
+        return response.data.map(item => item.attributes).find(_ => _.path === path);
```
mattandrews: This might need tweaking if/when we end up refactoring the API endpoint to work for multiple sections, but works okay for now.

mattandrews:
@davidrapson could you give this a review when you get a sec? Think we can merge this but would be good to explore further.

mattandrews:
... looks like a change to Prettier config upstream has added a bunch of changes here?

davidrapson:
Closing in favour of https://github.com/biglotteryfund/blf-alpha/pull/555

PR #504, opened 2018-01-05
Normalise how environment is determined, use appData

davidrapson:
Noticed we had a few different ways of determining current the node environment. This PR tidies things up so that we always get it from `appData` (which in turn uses whatever `node-config` thinks, defaulting to `development` if nothing is set).

### DIFF
--- a/controllers/toplevel/index.js
+++ b/controllers/toplevel/index.js
@@ -11,20 +11,20 @@ const moment = require('moment');
 
 const router = express.Router();
 
-const app = require('../../server');
-const contentApi = require('../../services/content-api');
-const surveyService = require('../../services/surveys');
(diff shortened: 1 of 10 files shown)


controllers/toplevel/index.js
```diff
@@ -11,20 +11,20 @@ const moment = require('moment');
...
+const surveyService = require('../../services/surveys');
+const routeStatic = require('../utils/routeStatic');
 const regions = require('../../config/content/regions.json');
 
 const legacyPages = require('./legacyPages');
 
 const robots = require('../../config/app/robots.json');
 // block everything on non-prod envs
-if (app.get('env') !== 'production') {
+if (appData.isNotProduction) {
```
mattandrews: 💯

PR #507, opened 2018-01-08
Refine funding finder redirects

davidrapson:
Ran a report of most frequently used funding finder query strings and tweaked the rules based on that.

- Correct handling of Northern Ireland query string
- Add support for UK-wide as it's frequently accessed
- Update tests


mattandrews:
👍

PR #493, opened 2017-12-20
Refactor locales

davidrapson:
- Convert locales to YAML
- Centralises locale logic, groups helper functions with more tests
- Cleans up middleware to properly separate i18n logic

🤔 💭

PR #528, opened 2018-01-09
Remove locales editor

davidrapson:
Removes the locales editor, for a couple of reasons:

1. More and more translatable text is been moved into the CMS
2. I'd still like to ship the change to move locales to YAML which likely breaks this editor anyway https://github.com/biglotteryfund/blf-alpha/pull/493
3. Deleting code is satisfying…

Caveat emptor that I've not had a need to use the tool so shout if you are using it all the time @mattandrews


mattandrews:
image

This is basically your ideal change ratio, right?

PR #540, opened 2018-01-11
Migrate awards for all a/b test

davidrapson:
This PR trials an approach for migrating the Awards For All A/B test to the new pages. ~The general plumbing should be uncontroversial, but the main considerations is **where should the text come from** I'm currently assuming a similar approach to the legacy site where the A variant comes from the CMS and then the B variant being hardcoded in the app.~

Edit: Updated this so that the A variant (where applications are offline primarily) is the text that comes from the CMS and that the B variant text is hard-coded. This is the same model as the previous A/B test.

### Tasks 

- [x] Reimplement test
- [ ] Add Welsh copy
- [ ] Create a new Google Experiment, update ID
- [ ] Launch changes, restart test

### DIFF
--- a/controllers/funding/programmes.js
+++ b/controllers/funding/programmes.js
@@ -185,7 +185,7 @@ function initProgrammeDetailAfaEngland(router, options) {
         }
     });
 
-    // const percentageForTest = config.get('abTests.tests.awardsForAll.percentage');
+    // @TODO use config.get('abTests.tests.awardsForAll.percentage') when launching;
     const percentages = splitPercentages(50);
 
     const getSlug = urlPath => last(urlPath.split('/'));
@@ -194,7 +194,7 @@ function initProgrammeDetailAfaEngland(router, options) {
         handleProgrammeDetail(getSlug(req.path))(req, res);
     });
 
-    router.get(options.path, testFn(null, percentages.A), (req, res) => {
+    router.get(options.path, testFn(null, percentages.B), (req, res) => {
         const locale = req.i18n.getLocale();
         const slug = getSlug(req.path);
         contentApi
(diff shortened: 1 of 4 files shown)


controllers/funding/programmes.js
```diff
@@ -128,27 +136,114 @@ function initProgrammesList(router, config) {
...
+    });
+}
+
+function initProgrammeDetailAfaEngland(router, options) {
+    const testFn = ab.test('blf-afa-rollout-england', {
+        cookie: {
+            name: config.get('cookies.abTestAwardsForAll'),
+            maxAge: moment.duration(4, 'weeks').asMilliseconds()
+        },
+        id: 'O9FsbkKfSeOammAP0JrEyA'
```
davidrapson: Note to self: It might make more sense to start a new experiment rather than reusing the previous experiment ID. Even though the test is the same in principle the URLs are different so might upset Google Experiments.

controllers/funding/programmes.js
```diff
@@ -128,27 +136,114 @@ function initProgrammesList(router, config) {
...
+            <p>If you need more information, please contact us at <a href="mailto:general.enquiries@awardsforall.org.uk">general.enquiries@awardsforall.org.uk</a> or call our advice team on <strong>0345 4 10 20 30</strong>. Text Relay users, please use <strong>18001</strong> plus <strong>0345 4 10 20 30</strong>.</p>
+        `,
+        // @TODO: Add translation
```
davidrapson: Update this once the new page is approved and we have revised Welsh language copy.

controllers/funding/programmes.js
```diff
@@ -128,27 +135,89 @@ function initProgrammesList(router, config) {
...
+
+function initProgrammeDetailAfaEngland(router, options) {
+    const testFn = ab.test('blf-afa-rollout-england', {
+        cookie: {
+            name: config.get('cookies.abTestAwardsForAll'),
+            maxAge: moment.duration(4, 'weeks').asMilliseconds()
+        }
+    });
+
+    // const percentageForTest = config.get('abTests.tests.awardsForAll.percentage');
```
mattandrews: Can probably delete this?

controllers/funding/programmes.js
```diff
@@ -128,27 +135,89 @@ function initProgrammesList(router, config) {
...
+    const percentages = splitPercentages(50);
+
+    const getSlug = urlPath => last(urlPath.split('/'));
+
+    router.get(options.path, testFn(null, percentages.A), (req, res) => {
+        handleProgrammeDetail(getSlug(req.path))(req, res);
+    });
+
+    router.get(options.path, testFn(null, percentages.A), (req, res) => {
```
mattandrews: Should this be `percentages.B`, even if they're technically the same?

PR #542, opened 2018-01-12
Add tests for viewGlobals methods

davidrapson:
Adds tests for view globals methods. Particularly focussed on the complex `buildUrl` and `getCurrentUrl` methods. (Branched off `title-fix` branch).

### DIFF
--- a/config/content/sass.json
+++ b/config/content/sass.json
@@ -9,6 +9,5 @@
         "xlarge": "1200px",
         "tablet": "760px",
         "desktop": "980px"
-    },
-    "themeColour": "#ec008c"
+    }
 }
--- a/modules/viewGlobals.js
+++ b/modules/viewGlobals.js
@@ -1,10 +1,11 @@
 'use strict';
 
 const config = require('config');
-const routes = require('../controllers/routes');
+const { get } = require('lodash');
 const { isWelsh, removeWelsh, stripTrailingSlashes } = require('./urls');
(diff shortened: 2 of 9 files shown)


config/content/sass.json
```diff
@@ -9,6 +9,5 @@
         "xlarge": "1200px",
         "tablet": "760px",
         "desktop": "980px"
-    },
-    "themeColour": "#ec008c"
```
davidrapson: This is being moved into the general config as we don't use it anywhere in our sass files so doesn't really need to be shared.

modules/viewGlobals.js
```diff
@@ -15,6 +25,83 @@ function getMetaTitle(base, pageTitle) {
...
+
+    return parts.join(' ');
+}
+
+/**
+ * buildUrl
+ * URL helper, return canonical URL based on sectionName or pageName
+ * Handle fallbacks for toplevel pages or linking to direct paths.
+ */
+function buildUrl(localePrefix) {
```
mattandrews: I'm not a fan of this function, feels hacky and brittle, but I think the more we move to the CMS, the less we'll need it.

davidrapson: Agreed, it took me a while to even test this function as I'm not 100% sure what the second part of the is _really_ intended for. There's room for the first part, where we link to a named route. It is better to test the brittle thing lest it hang around — Aristotle or something.

PR #547, opened 2018-01-15
Fix getCurrentUrl used for locale switcher

davidrapson:
So, it turns out I'd written tests to catch this case but made a syntax error stopping that test from running!

The main issue here was a mismatch between all our URL helpers operating on a URL path but `getCurrentUrl` working with a full URL with protocol and host.

I've reinstated the test and updated the function to work primarily on the URL path and the construct the full URL only at the end. Allows us to safely reuse shared URL helper functions.

### DIFF
--- a/modules/viewGlobals.js
+++ b/modules/viewGlobals.js
@@ -2,7 +2,7 @@
 
 const config = require('config');
 const { get } = require('lodash');
-const { isWelsh, removeWelsh, stripTrailingSlashes } = require('./urls');
+const { getBaseUrl, isWelsh, makeWelsh, removeWelsh, stripTrailingSlashes } = require('./urls');
 const { createHeroImage } = require('./images');
 const appData = require('./appData');
 const routes = require('../controllers/routes');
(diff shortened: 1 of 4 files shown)


modules/viewGlobals.js
```diff
@@ -77,29 +77,26 @@ function buildUrl(localePrefix) {
...
-    } else if (locale === 'en' && isCurrentUrlWelsh) {
-        // un-welshify this URL
-        currentUrl = removeWelsh(currentUrl);
+function getCurrentUrl(req, requestedLocale) {
+    const urlPath = req.originalUrl;
+    const baseUrl = getBaseUrl(req);
+
+    const isCurrentUrlWelsh = isWelsh(urlPath);
+    const isCyWithEnRequested = isCurrentUrlWelsh && requestedLocale === 'en';
+    const isEnWithCyRequsted = !isCurrentUrlWelsh && requestedLocale === 'cy';
```
mattandrews: Typo `Requsted`!

PR #562, opened 2018-01-18
Remove sass.json config sharing

davidrapson:
The only place we were sharing breakpoints between sass and javascript was in the carousel to determine how many slides to show at once. However, the only carousel we have on the site always shows a single item so we don't need the sharing.

For this reason, and to avoid some indirection and prune unneeded dependencies I've inlined all the config into `_vars.scss` and removed the config sharing.


mattandrews:
Yeah, this was probably a good example of me over-engineering something that we didn't really need. Nice work!

PR #588, opened 2018-01-23
Redirect ugly links (PENDING)

mattandrews:
Turns this:
https://www.biglotteryfund.org.uk/~/link.aspx?_id=50FAB7D4B5A248F8A8C8F5D4D33F9E0F&_z=z

Into this:

https://www.biglotteryfund.org.uk/global-content/programmes/england/building-better-opportunities/guide-to-delivering-european-funding

This is fairly straightforward but could do with another pair of eyes on it.

Note that merging this will make it live.

### DIFF
--- a/modules/proxy.js
+++ b/modules/proxy.js
@@ -136,7 +136,7 @@ const proxyLegacyPage = (req, res, domModifications, pathOverride) => {
 };
 
 const redirectUglyLink = (req, res) => {
-    let handleError = () =>  res.redirect('/');
+    let handleError = () => res.redirect('/');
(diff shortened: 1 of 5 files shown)


modules/proxy.js
```diff
@@ -7,6 +7,7 @@ const absolution = require('absolution');
 const jsdom = require('jsdom');
 const { JSDOM } = jsdom;
 const appData = require('../modules/appData');
+const cheerio = require('cheerio');
```
davidrapson: Little thing, but it feels odd to use `cheerio` here when `jsdom` can do the same thing?

mattandrews: Done!

PR #598, opened 2018-01-30
Convert remaining icons into macros

davidrapson:
Continues on from https://github.com/biglotteryfund/blf-alpha/pull/556

Converts remaining icons into macros to help with accessibility: allows us to assign ids to aria- properties to avoid duplicate ids. As part of this PR I moved the description for left and right arrows to the carousel buttons themselves (as these arrows are used elsewhere so the description could be misleading).


mattandrews:
Got a merge conflict, FYI

PR #611, opened 2018-02-02
Research landing page

davidrapson:
localhost_3000_research


config/locales/cy.json
```diff
@@ -190,6 +190,76 @@
                 "label": "Cymorth rheoli’ch grant",
                 "href": "/welsh/funding/funding-guidance/managing-your-funding"
             }]
+        },
+        "research": {
+            "title": "Research",
```
mattandrews: Should this be in Welsh?

davidrapson: Yep, the rest of the page still needs to be sent for translation, but I'll update the title now.

PR #625, opened 2018-02-06
I expect nothing

davidrapson:
Fix the section navigation test so we, y'know, test something.


integration-tests/test-suites/core.test.js
```diff
@@ -48,7 +48,7 @@ describe('Core sections and features', () => {
                     const navItem = document.querySelector(`.qa-nav-link--${activeSection}`);
 
                     // Test the correct navigation item is selected
-                    expect(navItem.classList.contains('is-selected'));
```
davidrapson: If you have no expectations then you can never fail, right?

PR #629, opened 2018-02-07
About page amends

davidrapson:
Updates copy with latest and amends and adds links in.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -191,26 +191,29 @@ toplevel:
     asideAnnualReport:
         quoteText: In 2016/17 we awarded over £700 million, including 10,000 grants under £10,000
         linkText: Read our latest annual report
-        linkHref: #
+        linkHref: "/about-big/publications/corporate-documents"
         imageCaption: Pennine Lancashire Community Farm
-    goodCauses: The National Lottery raises money for good causes. It makes a difference to people by
-      transforming communities, protecting our heritage, and enriching lives through arts, sports and culture.
-      Communities come in all shapes and sizes, and National Lottery funding is there for everyone.
+    howOurFundingIsUsed:
+        part1: People use our funding to do extraordinary things,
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -191,26 +191,29 @@ toplevel:
...
         linkText: "Discover our strategic framework: People in the Lead"
-        linkHref: #
+        linkHref: "/about-big/strategic-framework"
         imageCaption: Dawn Austwick, CEO
-    outro: We are the largest distributor of National Lottery funding,
-      established as a non-departmental public body by an Act of Parliament.
-      Our work is divided into five portfolios, which cover funding across England,
-      Northern Ireland, Wales and Scotland, and the UK as a whole.
+    outro: We’re the largest distributor of National Lottery funding,
+      established as a non- departmental public body by an Act of Parliament.
```
mattandrews: Do we need a space here after `non-`?

config/locales/en.yml
```diff
@@ -191,26 +191,29 @@ toplevel:
...
+      Our work is divided into five major portfolios, which covering funding across
```
mattandrews: Should this be "cover"?

config/locales/en.yml
```diff
@@ -191,26 +191,29 @@ toplevel:
...
+      England, Northern Ireland, Wales and, Scotland, and the UK as a whole.
```
mattandrews: No need for comma after "and"

config/locales/en.yml
davidrapson: Just checked the doc, and I think so. I must've picked up some track-changes leftovers.

config/locales/en.yml
davidrapson: Indeed

davidrapson:
@mattandrews Mind giving this another look, I've added a few more bits.

PR #633, opened 2018-02-07
Add more funding redirects

mattandrews:


controllers/routes.js
```diff
@@ -44,7 +44,7 @@ const routes = {
                     static: false,
                     live: true,
                     isPostable: true,
-                    aliases: ['/home', '/index.html', '/en-gb', '/england', '/uk-wide']
+                    aliases: ['/home', '/index.html', '/en-gb', '/england', '/uk-wide', '/funding/funding-guidance/applying-for-funding']
```
davidrapson: Having this as an alias feels…odd even though it does what we want. IDK. Works well enough.

PR #638, opened 2018-02-08
Streamline routes

davidrapson:
This PR does some more work to streamline routes code. I'm approaching this with a few main goals:

- Routes file should be shorter (Met with 544 lines to 389)
- Reduce nesting in routes file to improve scanability
- Reduce duplication by defining common _shapes_ of routes
- Make route configs more testable

Does this make sense?

The specific changes are:

- Inline loadController code
- Extract route type helpers
- Reduce nesting of primary routes
- Cleanup main routes using route types
- Trim names, use object keys
- Add unit tests for route-types

### DIFF
--- a/controllers/route-types.js
+++ b/controllers/route-types.js
@@ -1,0 +1,100 @@
+'use strict';
+
+function createSection({ path, controllerPath, langTitlePath }) {
+    return {
+        path,
+        controller: function(pages, sectionPath, sectionId) {
+            return require(controllerPath)(pages, sectionPath, sectionId);
+        },
+        langTitlePath,
+        pages: null
+    };
+}
+
+const defaults = {
+    isPostable: false,
+    allowQueryStrings: false,
+    live: false
+};
+
+function withDefaults(props) {
+    return Object.assign({}, defaults, props);
+}
+
+function basicRoute(props) {
+    return Object.assign(
+        {},
+        defaults,
+        {
+            live: true
+        },
+        props
+    );
+}
+
+function staticRoute(props) {
+    const staticDefaults = {
+        static: true,
+        live: true
+    };
+    return Object.assign({}, defaults, staticDefaults, props);
+}
+
+function dynamicRoute(props) {
+    const dynamicDefaults = {
+        static: false,
+        live: true
+    };
+    return Object.assign({}, defaults, dynamicDefaults, props);
+}
+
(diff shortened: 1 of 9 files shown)


controllers/route-types.js
```diff
@@ -0,0 +1,100 @@
...
+
+function cmsRoute(props) {
+    const cmsDefaults = {
+        useCmsContent: true,
+        live: true
+    };
+    return Object.assign({}, defaults, cmsDefaults, props);
+}
+
+function legacyRoute(props) {
```
mattandrews: Might be nice (for our own sanity) to comment each type briefly to define the distinctions?

controllers/routes.js
```diff
@@ -1,358 +1,244 @@
...
+        path: '/funding-guidance/applying-for-funding/*'
+    }),
+    manageFunding: staticRoute({
+        path: '/funding-guidance/managing-your-funding',
+        template: 'pages/funding/guidance/managing-your-funding',
+        lang: 'funding.guidance.managing-your-funding',
+        aliases: ['/funding/funding-guidance/managing-your-funding/help-with-publicity', '/welcome', '/publicity']
+    }),
+
+    // @TODO: This is the only route that uses isWildcard??
```
mattandrews: This is because it `POST`s to `/funding/funding-guidance/managing-your-funding/ordering-free-materials/$id` where `$id` is a product ID. It could quite easily `POST` to a generic endpoint and pass `$id` as data instead... future refactoring job maybe.

controllers/routes.js
```diff
@@ -1,358 +1,244 @@
...
+    }),
+    pressCoverage: staticRoute({
+        path: '/funding-guidance/managing-your-funding/press',
+        template: 'pages/funding/guidance/getting-press-coverage',
+        lang: 'funding.guidance.getting-press-coverage',
+        static: true,
+        live: true
+    }),
+    logos: staticRoute({
+        path: '/funding-guidance/managing-your-funding/grant-acknowledgement-and-logos/logodownloads',
```
mattandrews: Since we're tweaking stuff – this URL is actually wrong, path-wise. Could you remove the `/logodownloads` from the end (make it a redirect here instead) and have the canonical path be `/funding-guidance/managing-your-funding/grant-acknowledgement-and-logos`?

PR #658, opened 2018-02-09
New search handler [Draft]

davidrapson:
Adds a new (draft) route handler which redirects search queries out to a google site search. Opted to add a new route handler rather than just updating the action of the search form so that we can also pick up search queries from the legacy site.

This PR ended up a little broader and also includes the following:

- Extract homepage handler into a separate file
- Fix another little bug with selected pages in the nav, pushed all related code into a new `setup` function.
- Added an integration test for search


controllers/toplevel/search.js
```diff
@@ -0,0 +1,19 @@
+const querystring = require('querystring');
+const { noCache } = require('../../middleware/cached');
+const { trackPageview } = require('../../modules/analytics');
+
+function init({ router, routeConfig }) {
+    const queryBase = 'https://www.google.co.uk/search?q=site%3Abiglotteryfund.org.uk';
```
mattandrews: Do you think we should use Welsh-language Google for the CY locale?

davidrapson: Good shout

PR #659, opened 2018-02-09
Test improvements

davidrapson:
Never sure how useful code-coverage reports are, but this is easy to integrate with mocha (a one-liner) so thought it worth a punt. Also updates secrets module to be testable.

screen shot 2018-02-09 at 12 59 59


modules/secrets.js
```diff
@@ -0,0 +1,43 @@
...
+    return rawParameters;
+}
+
+function parseSecrets(rawParameters) {
+    const mapKeyedValues = flow(keyBy('Name'), mapValues('Value'));
+    const secrets = mapKeyedValues(rawParameters);
+    return secrets;
+}
+
+function getSecretFromRawParamters(rawParameters, name) {
```
mattandrews: Typo here (`Paramters`)

PR #667, opened 2018-02-13
Search tracking update

davidrapson:
- Switch tracking to event tracking rather than pageviews
- Normalise query string to account for legacy site url encoding
- Add integration test for legacy search queries

### DIFF
--- a/controllers/toplevel/search.js
+++ b/controllers/toplevel/search.js
@@ -1,12 +1,15 @@
 const querystring = require('querystring');
 const { noCache } = require('../../middleware/cached');
-const { trackPageview } = require('../../modules/analytics');
+const { customEvent } = require('../../modules/analytics');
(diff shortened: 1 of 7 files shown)


controllers/toplevel/search.js
```diff
@@ -1,12 +1,15 @@
...
+const { normaliseQuery } = require('../../modules/urls');
 
 function init({ router, routeConfig }) {
     const queryBase = 'https://www.google.co.uk/search?q=site%3Abiglotteryfund.org.uk';
     router.get(routeConfig.path, noCache, (req, res) => {
+        req.query = normaliseQuery(req.query);
+
         if (req.query.q) {
-            trackPageview(req);
+            customEvent('Search', 'Term', req.q);
```
mattandrews: Should this be `req.query.q`?

davidrapson: Yes…

PR #670, opened 2018-02-13
Add querystrings to route config

mattandrews:
... and only generate them if they exist.


controllers/routes.js
```diff
@@ -283,16 +283,15 @@ const otherUrls = [
     }),
     basicRoute({
         path: '/tools/*',
-        isPostable: true,
-        allowQueryStrings: true
```
mattandrews: (this didn't actually use any)

PR #673, opened 2018-02-13
Update how cloudfront configs are generated

davidrapson:
- Safe default config in addition to cache behaviours
- Simplify cloufront generation
- Add white-list route for assets
- Modify update-cloudfront to use new schema

### DIFF
--- a/bin/generate-cloudfront-config
+++ b/bin/generate-cloudfront-config
@@ -57,7 +57,7 @@ forEach(cloudfrontDistributions, (distribution, distributionName) => {
     const behaviours = generateBehaviours(distribution);
 
     const confData = {
-        DefaultBehaviour: defaultBehaviour,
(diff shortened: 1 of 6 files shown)


bin/generate-cloudfront-config
```diff
@@ -2,69 +2,75 @@
...
-    behaviours = sortBy(behaviours, 'PathPattern');
-    return behaviours;
-};
+    const customBehaviours = [
+        makeBehaviourItem({
+            origin: 'legacy',
+            originServer: distribution.origins.legacy,
+            pathPattern: '/~/*',
+            isPostable: true,
+            cookiesInUse: cookiesInUse
```
mattandrews: Wonder if we need this – possibly it should allow any query strings and any cookie?

PR #678, opened 2018-02-15
Redirect regional URLs

davidrapson:
Adds a new redirect middleware which is responsible for redirecting region specific URLs so that for example `/{england,scotland,wales/northernireland}/about-big/publications/corporate-documents` all redirect to `/about-big/publications/corporate-documents`


mattandrews:
Looks good code-wise but maybe let's wait with this one to confirm it's safe to effectively kibosh the country-specific pages.

PR #683, opened 2018-02-16
[DRAFT] Add preview mode

mattandrews:
This is more conceptual than anything else, but worth a look.

This adds the concept of a "preview domain" (eg. `preview.biglotteryfund.org.uk`) which, if used, would attempt to request draft/revisions of CMS content (based on `?draft=$id` or `?version=$id` parameters in the path).

There's a corresponding branch over at https://github.com/biglotteryfund/craft-dev which allows Craft to output this content (for authorised users). This PR passes through all cookies to the API call (when on the preview domain) so we can test this integration out.

There's also a discussion to be had around error handling: if someone is unauthorised, should we tell them, or just send them to the non-preview version of the page? Likewise if someone is successful, how should we indicate that the page is a preview/draft?

There's also some changes here to update the `/robots.txt` handler to block everything when the site is served over `preview.` (which would need IT to create a duplicate CNAME from the `www`) one.

Alternatively we could use the internal domain to serve these requests? 

Thoughts very welcome.


mattandrews:
Craft cookie looks to be called `CraftSessionId`.

Assuming that `preview.` is pointed at Cloudfront too, we'd need to whitelist that cookie for every page, as well as pass it on to the API request.

middleware/preview.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+const { getSecret } = require('../modules/secrets');
+
+// Sets a global parameter to indicate if this request came from a preview domain
+module.exports = (req, res, next) => {
+    let PREVIEW_DOMAIN = process.env.PREVIEW_DOMAIN || getSecret('preview.domain');
```
davidrapson: This could happen outside of the middleware so we're not looking this up on every request.

middleware/preview.js
```diff
@@ -0,0 +1,21 @@
...
+            let queryData = req.query[mode];
+            if (queryData) {
+                previewData.mode = mode;
+                previewData.id = parseInt(queryData);
+            }
+            return queryData;
+        });
+        res.locals.PREVIEW_MODE = previewData;
+    }
+    return next();
```
davidrapson: Pop a test in for this middleware perhaps?

davidrapson:
>  if someone is unauthorised, should we tell them, or just send them to the non-preview version of the page?

I'd rather send people to a non-preview version of the page. Although this feels like something to test with people to get a sense of what the natural expectation is.

> Likewise if someone is successful, how should we indicate that the page is a preview/draft?

A banner a the top of the page would be nice.

middleware/preview.js
mattandrews: Done!

middleware/preview.js
mattandrews: Just added a test for this.

davidrapson:
image

davidrapson:
🚀  👁

PR #687, opened 2018-02-19
Clean up secrets

davidrapson:
Rather than dotting lot's of `process.env.something || getSecret('something')` round the place, I've tried to move all the definitions into `secrets.js` so they can be imported directly where needed.

### DIFF
--- a/modules/secrets.js
+++ b/modules/secrets.js
@@ -1,4 +1,5 @@
 'use strict';
+const config = require('config');
 const fs = require('fs');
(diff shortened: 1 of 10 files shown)


modules/secrets.js
```diff
@@ -36,8 +37,34 @@ function getSecret(name) {
     }
 }
 
+// @TODO: Make all process.env vars consistent case
```
mattandrews: Think this is down to Travis, could change them all there when this goes out?

davidrapson: Good point, will merge this as is and then double check / clean things up soon.

PR #707, opened 2018-02-22
Validation update

davidrapson:
Based off https://github.com/biglotteryfund/blf-alpha/pull/684

Upgrade to `express-validator` 5 and begins to add translations using the new `withMessage` behaviour.


config/locales/en.yml
```diff
@@ -152,6 +152,12 @@ toplevel:
       lastName: Surname
       organisation: Organisation
       emailAddress: Email Address
+    errors:
+      firstName: Please provide your first name
```
mattandrews: Just realised we're missing the Welsh versions of these?

davidrapson:
> Wondering it might make more sense to have a function/placeholder (eg. __('Please provide %s', 'your first name'))? – but this is good for now.

Handled this a little bit differently but the translation function now accepts replacements too.

Worth giving this another look as there's a fair few changes on the materials form. Now that the code is a little less verbose I've inlined all materials controller code into a the one file.

modules/validators.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+function errorTranslator(prefix) {
+    return function(prop, replacementKeys = []) {
+        return function(value, { req }) {
+            const t = `${prefix}.${prop}`;
+            const replacements = replacementKeys.map(_ => req.i18n.__(_));
```
davidrapson: This needs more underscores.

config/locales/cy.yml
```diff
@@ -153,6 +153,12 @@ toplevel:
       lastName: Cyfenw
       organisation: Mudiad
       emailAddress: Cyfeiriad e-bost
+    errors:
+      firstName: Please provide your first name
+      lastName: Please provide your last name
+      emailMissing: Please provide your email address
+      emailInvalid: Please provide a valid email address
+      location: Please choose a country
```
mattandrews: We have most of this already:

`Please provide the '%s' field` = `Darparwch y maes '%s'`
`Please provide a valid email address` = `Rhowch gyfeiriad e-bost dilys`

You could probably translate this yourself by just adding in the field names?

Issue #690, opened 2018-02-20
Link to online form from Scotland and Wales versions of NLAFA pages


TomSteinberg:
50% of traffic

mattandrews:
Now live!

PR #709, opened 2018-02-22
Expand AFA test

mattandrews:
This adds Scotland and Wales into the A/B test for the online form (eg. #690).

One thing to consider: do we need per-route experiment IDs for GA?

### DIFF
--- a/views/pages/toplevel/under10k.njk
+++ b/views/pages/toplevel/under10k.njk
@@ -41,25 +41,25 @@
 
             <ul class="unstyled grid grid--padded grid--equal grid--4-up grid--wide-only">
                <li class="grid__item">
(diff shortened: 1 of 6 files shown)


davidrapson:
> One thing to consider: do we need per-route experiment IDs for GA?

Yep :)

views/pages/toplevel/under10k.njk
```diff
@@ -41,25 +41,25 @@
 
             <ul class="unstyled grid grid--padded grid--equal grid--4-up grid--wide-only">
                <li class="grid__item">
-                   <a href="{{ buildUrl('global-content/programmes/england/awards-for-all-england') }}"
+                   <a href="{{ buildUrl('funding/programmes/national-lottery-awards-for-all-england') }}"
```
davidrapson: Good catch!

PR #714, opened 2018-02-23
Update cache control headers to use s-max-age

davidrapson:
Updates how cache control headers are set to use `s-max-age`. Allows us to set a lower cache time for the browser and a longer (or even custom) cache time for CloudFront.

Defaults to: `max-age=30,s-maxage=300`

Adds a extra middleware and option to the routes config to extent the `s-max-age` value for specific pages. To start with I've set the cache time for `/funding`, `/research`, and `/about` to 30 minutes.

Going forward we could look at setting these much higher and purging routes (e.g. on CMS save or post-deploy) but that's another conversation.


middleware/cached.js
```diff
@@ -1,10 +1,16 @@
 'use strict';
-const config = require('config');
+const ms = require('ms');
 const csurf = require('csurf');
 const vary = require('vary');
 const cacheControl = require('express-cache-controller');
 
-const defaultMaxAge = config.get('viewCacheExpiration');
+const toSeconds = naturalTime => {
+    const parsedMilliseconds = typeof naturalTime === 'string' ? ms(naturalTime) : naturalTime;
```
mattandrews: You can probably do this with `moment` already, like `moment.duration(4, 'weeks').asMilliseconds()` etc? But perhaps doesn't have the natural time support.

davidrapson: Doesn't seem to have the shorthand syntax.

PR #715, opened 2018-02-23
Remove unused gmail script and dependencies

davidrapson:
I'm right that this is OK to go now @mattandrews ?


config/content/sass.json
```diff
@@ -1,13 +0,0 @@
-{
```
davidrapson: The sass sharing stuff went a while ago but this file was still hanging around. Straggler.

PR #720, opened 2018-02-23
Upgrade swiper to version 4

davidrapson:
- Upgrade Swiper to version 4, mostly syntax updates behaviour should be the same
- Import vendor styles directly from `node_modules` rather than copy-pasting to ensure the styles are linked to the version


davidrapson:
I've fixed the build the simplest way, but at the expense of increasing the build size by a few KB. This version has gone all "modular" so might come back to this to see if we can get the size down again.

davidrapson:
@mattandrews Mind giving this another look? I've rejigged this to use ES6 modules where available. Seems…promising so far. Module import of swiper + this line `const { forEach } = require('lodash');` seem to pull the build size down *a lot*. To ~86KB.

Going to test this out on some older devices and browsers first to confirm that nothing is amiss. But looking good.

mattandrews:
Looks good – nice work and good cleanup!

PR #721, opened 2018-02-23
Refactor surveys

mattandrews:
This combines with https://github.com/biglotteryfund/craft-dev/pull/39 to serve surveys from the CMS rather than directly from the database.

Note that when deploying this it'll require some manual database changes which might be a bit fiddly as we don't have migrations for Sequelize. Basically involves deleting the two redundant tables (`surveys` and `survey_choices`) and adding two new `INT` columns to `survey_responses`). I'll export from my DEV environment when we're ready to do this, which preserves existing responses.

Also requires syncing the Craft environment on TEST for the "Survey" section.


services/surveys.js
```diff
@@ -1,53 +1,95 @@
...
-        }
-    });
-}
+    // get user responses from the database
+    const getResponses = SurveyResponse.findAll();
+    const normalisedDateFormat = 'YYYY-MM-DD';
+
+    // combine the votes with the choices
+    return Promise.all([getSurveys, getResponses]).then(responses => {
+        let surveys = responses[0];
```
davidrapson: could use `const [surveys, votes] = responses;` here?

services/surveys.js
```diff
@@ -1,53 +1,95 @@
...
+            // append responses to the relevant choice
+            survey.choices = survey.choices.map(choice => {
+
+                // retrieve the votes for this survey's choices
+                choice.responses = votes.filter(v => {
+                    return v.survey_id === survey.id && v.choice_id === choice.id;
+                });
+
+                // fill in gaps for days without votes
+                let oldestVote = _.minBy(choice.responses, 'createdAt');
```
davidrapson: Maybe import the individual lodash methods as we're doing elsewhere?

views/pages/tools/surveys.njk
```diff
@@ -3,29 +3,146 @@
...
-            {% else %}
-                <p>No additional messages.</p>
+            {% if choice.allowMessage %}
+                <div class="border mb-5" style="max-height: 400px; overflow: auto">
+                    <div class="list-group">
+                        {% for response in choice.responses %}
+                            {% if response.message %}
+                                <div class="list-group-item list-group-item-action flex-column align-items-start">
+                                    <div class="d-flex w-100 justify-content-between">
+                                        <h6 class="mb-3">{{ moment(response.createdAt).format('ddd DD MMM YYYY (hh:mm a)') }}</h6>
```
davidrapson: Not a _huge_ fan of moment in the view, rather than creating a filter or something.

views/pages/tools/surveys.njk
```diff
@@ -3,29 +3,146 @@
...
+                                </div>
+                            {% endif %}
+                        {% endfor %}
+                    </div>
+                </div>
             {% endif %}
         {% endfor %}
     {% endfor %}
+
+    <script>
```
davidrapson: Maybe create a dedicated file for this (and import charts into there) like we have with `styleguide.js`?

views/components/surveys.njk
```diff
@@ -14,7 +14,7 @@
         >
             <h6 class="survey__header"
                 v-on:click="toggleSurvey()">
-                <% localeify(survey, 'question', '{{ locale }}') %><span class="survey__arrow js-survey-arrow">
```
davidrapson: Can `localeify` go from `viewFilters ` too?

services/surveys.js
```diff
@@ -1,53 +1,95 @@
...
+                return {
+                    title: choice.title,
+                    votes: choice.responses.length
+                };
+            });
+
+            // find out the winner's percentage
+            let winner = _.maxBy(voteTotals, 'votes');
+            let totalResponses = _.sumBy(voteTotals, 'votes');
+            let winnerPercentage = Math.round(winner.votes / totalResponses * 100);
```
davidrapson: `winner.votes` rather than `winner`? Does `let winner = _.maxBy(voteTotals, 'votes');` not return a number?

services/surveys.js
```diff
@@ -1,53 +1,95 @@
...
+                let oldestVote = _.minBy(choice.responses, 'createdAt');
+                let newestVote = _.maxBy(choice.responses, 'createdAt');
+
+                // calculate votes per day (or substitute with zeroes)
+                if (newestVote && oldestVote) {
+                    let daysInRange = moment(newestVote.createdAt).diff(moment(oldestVote.createdAt), 'days');
+
+                    let counts = {};
+                    // group votes by day to aid in graphing
+                    choice.responses.forEach(response => {
```
davidrapson: Could do this with `reduce` rather than `forEach` to make it into one statement?

services/surveys.js
mattandrews: Nope, returns the `choice` object with the `title`/`votes` properties.

views/components/surveys.njk
mattandrews: Yep!

PR #725, opened 2018-02-26
Conditional Cymraeg

davidrapson:
This PR makes some updates to how pages without welsh translations are handled:

- Allows custom pages to have a `isBilingual = false` flag to be explicitly marked as monolingual.
- Sets `isBilingual` flag for funding programmes based on `availableLanguages` listed by API 
- Adds a new wildcard handler to redirect any un-handled welsh urls to their english counterpart
- Updates the funding programmes list to allow programmes without welsh translations to be disabled in the CMS

This last point is where most of the change are. In order to be able to disable the welsh language translation in the CMS without having the programme disappear from `/welsh/funding/programmes` the funding programmes list has been updated so that english entries are used by default and welsh translation summaries are used if available when appropriate.

screen shot 2018-02-26 at 16 36 21


services/fixtures/funding-programmes-cy.json
```diff
@@ -0,0 +1,85 @@
+{
+  "data": [
+    {
+      "type": null,
+      "id": "70",
+      "attributes": {
+        "status": "live",
+        "title": "National Lottery Awards for All Wales",
+        "url": "https://test.blf.digital/welsh/funding/programmes/national-lottery-awards-for-all-wales",
```
mattandrews: Can we keep this domain out of git, if possible?

davidrapson: Ack, yep

PR #730, opened 2018-02-28
Client-side build upgrade

davidrapson:
This PR is a reworking of https://github.com/biglotteryfund/blf-alpha/pull/720

Webpack 4 is out so I've upgraded to it and cleaned up the build process as a result. Main optimisations are to:

- Switch to using the built in `mode` parameter to opt-in to optimisations without having to do it ourselves
- Update some related CSS build dependencies, use latest autoprefixer.

**But**

The yak shaving part of this is that upgrade is that Swiper breaks again when using the modular build (can only be fixed by adding `exclude: /node_modules\/(?!(dom7|swiper)\/).*/,` to the webpack config according to their docs which I'm not doing). So I've switched back to the non-modular build. Back to thinking that Swiper is too fussy and I might come-back and replace this.

Upside is that the build is still smaller than it was, just not as big an improvement until Swiper goes.

🐐


davidrapson:
image

Argh `package-lock.json` ruining my otherwise large amount of deletions. JavaScript!

assets/js/modules/materials.js
```diff
@@ -19,6 +19,7 @@ function init() {
     let storeLangPrefInForm = newState => $('#js-language-choice').val(newState);
 
     new Vue({
+        delimiters: ['<%', '%>'],
```
mattandrews: I would prefer to set this globally if possible rather than on each module, otherwise it breaks pages in hilarious ways if we forget it (gets confused with Nunjucks' `{{ syntax }}`. It's a sitewide config rather than a per-instance one, you know?

mattandrews:
> Back to thinking that Swiper is too fussy and I might come-back and replace this.

Yep... thinking this too. I used a library for the off-canvas swipe nav, HammerJS. Wonder if that might be useful for the carousel too?

PR #733, opened 2018-02-28
New senior management team page

davidrapson:
This is still work in progress but there is enough going on here that it's worth merging this in parts and I'll raise a follow up PR to finish this off.

- Adds a new (draft) template for the senior management team which fetches profiles for that section from the CMS
- Does some work to streamline content-api calls so we weren't repeating the same request syntax everyhere
- Updates how images are modelled so there is a single list of available images in `modules/images`. Fixes a bit of duplication and even one broken image path.

smt-page

/cc @mattandrews


davidrapson:
@mattandrews Now with gif! Worked with Chloe on a different pattern for a quick-nav / jump-links. The code for this page is pretty much ready. A few CMS content bits to do (+translations) but this is ready for review now.

PR #735, opened 2018-02-28
Material order stats

mattandrews:
In the same vein as the survey results page, this shows some stats/charts for the material order form.

Did this because I got asked for some stats from the branding team which involved some manual SQL work (plus this was always on the cards when storing order data).

Not going to put screenshots here to keep data confidential but adds a little dashboard for us to keep an eye on.

### DIFF
--- a/services/orders.js
+++ b/services/orders.js
@@ -1,7 +1,8 @@
 const moment = require('moment');
 const { Op } = require('sequelize');
 const { Order, OrderItem } = require('../models');
-const { take, countBy, meanBy, chain, sortBy } = require('lodash');
+const { take, countBy, meanBy, sortBy, flatMap, map, reverse } = require('lodash');
+const { filter } = require('lodash/fp');
 
 function getAllOrders() {
     return Order.findAll({
@@ -17,7 +18,7 @@ function getAllOrders() {
         // the date we added dropdown grant size options rather than free text
         const LAUNCH_DATE = '2018-02-01';
 
-        let items = orders.reduce((acc, cur) => acc.concat(cur.items), []);
+        const items = flatMap(orders, order => order.items);
 
         let totalOrders = orders.length;
         let averageProductsPerOrder = Math.round(meanBy(orders, 'items.length'));
@@ -26,16 +27,8 @@ function getAllOrders() {
         );
 
         let orderByCount = arr => {
-            return chain(arr)
-                .map((count, code) => {
-                    return {
-                        code: code,
-                        count: count
-                    };
-                })
-                .sortBy('count')
-                .reverse()
-                .value();
+            const mapped = map(arr, (count, code) => ({ code, count }));
+            return reverse(sortBy(mapped, 'count'));
         };
 
         // sum items by their quantity ordered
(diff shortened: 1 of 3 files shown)


services/orders.js
```diff
@@ -1,8 +1,96 @@
...
+        // the date we added dropdown grant size options rather than free text
+        const LAUNCH_DATE = '2018-02-01';
+
+        let items = orders.reduce((acc, cur) => acc.concat(cur.items), []);
+
+        let totalOrders = orders.length;
+        let averageProductsPerOrder = Math.round(meanBy(orders, 'items.length'));
+        let averageItemQuantityPerOrder = Math.round(meanBy(orders, o => o.items.reduce((acc, cur) => acc + cur.quantity, 0)));
+
+        let orderByCount = arr => {
```
davidrapson: I'd err towards breaking this out rather than relying on `chain`. Something like:

``` js
const mapped = map(arr, (count, code) => ({ code, count }));
return reverse(sortBy(mapped));
```

services/orders.js
```diff
@@ -1,8 +1,96 @@
...
+            }
+            acc[cur.code] += cur.quantity;
+            return acc;
+        }, {});
+
+        // order items by total ordered
+        let mostPopularItemsByQuantity = orderByCount(itemsByQuantity);
+        let mostPopularItems = take(orderByCount(countBy(items, 'code')), 10);
+
+        let orderReasons = orderByCount(countBy(orders, 'orderReason')).filter(o => {
```
davidrapson: This could possibly be flattened up a bit using `lodash/fp`, something like:

```js
const { filter } = require('lodash/fp');

const filterIsRecentOrder = filter(_ => moment(_.createdAt).isAfter(LAUNCH_DATE));
const filterHasCode = filter(_ => _.code !== '');

const orderReasons = filterHasCode(orderByCount(countBy(orders, 'orderReason')));

const ordersAfterLaunch = filterIsRecentOrder(orders);
const grantAmounts = filterHasCode(orderByCount(countBy(ordersAfterLaunch, 'grantAmount')));
```

services/orders.js
```diff
@@ -1,8 +1,96 @@
...
+        let ordersByDay = orders.reduce((acc, order) => {
+            let normalisedDate = moment(order.createdAt).format(normalisedDateFormat);
+            if (!acc[normalisedDate]) {
+                acc[normalisedDate] = 0;
+            }
+            acc[normalisedDate] = acc[normalisedDate] + 1;
+            return acc;
+        }, {});
+
+        let voteData = [];
```
davidrapson: Maybe something like this?

``` js
const sortedOrdersByDay = sortBy(map(ordersByDay, (order, date) => ({
    x: date,
    y: order
}), 'x'));
```

services/orders.js
```diff
@@ -1,8 +1,96 @@
...
+        include: [
+            {
+                model: OrderItem,
+                as: 'items'
+            }
+        ]
+    }).then(orders => {
+        const normalisedDateFormat = 'YYYY-MM-DD';
+        // the date we added dropdown grant size options rather than free text
+        const LAUNCH_DATE = '2018-02-01';
```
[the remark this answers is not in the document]
mattandrews: Probably not – after this date, values are either picked from a list or entered as "other" (meaning they could look like the older entries which were always manually typed). Bit hacky but in a few months time (eg. 1st July or so) this will stop being a problem as all records before this date will have been deleted.

PR #745, opened 2018-03-02
Update heroes [WIP]

mattandrews:
This removes most of them – a few still to do, and also some discussion needed about if each image works.

This is also @LynseyJReynolds's first contribution to the project! Hooray!
giphy


mattandrews:
Do we still need the logic for the old header? I guess we do for the accessible header, right @davidrapson?

davidrapson:
@mattandrews The accessible header code is separate, so should work regardless of the header image that's used. But the whole first block of CSS in `heroes.scss` can go.

assets/sass/_heroes.config.scss
```diff
@@ -4,51 +4,11 @@ $heroes: (
...
-        (
-            'default': ('width': 644, 'height': 460, 'filename': '../../../images/hero/press--mobile.jpg'),
-            'tablet': ('width': 1364, 'height': 500, 'filename': '../../../images/hero/press--desktop.jpg')
-        ),
-    'contact':
-        (
-            'default': ('width': 644, 'height': 460, 'filename': '../../../images/hero/contact--mobile.jpg'),
-            'tablet': ('width': 1364, 'height': 500, 'filename': '../../../images/hero/contact--desktop.jpg')
-        ),
     'jobs':
```
davidrapson: This can go too, no?

davidrapson:
Looking at this again the `heroes.scss` code can only go once we've replaced the final two images:

- https://www.biglotteryfund.org.uk/404 (which I missed off the list)
- https://www.biglotteryfund.org.uk/empowering-young-people (which @ChloeAlper has already done a new image for but we need to migrate)

This is good to go as I then I think!

views/components/hero.njk
```diff
@@ -110,7 +110,11 @@
 {% macro miniatureHero(titleText, imagePath, imagePathMobile = false, imageCaption = false, linkUrl = false, accent, useRemoteImages = false) %}
     {% if not useRemoteImages %}
         {% set imagePath = imagePath | getImagePath %}
-        {% set imagePathMobile = imagePathMobile | getImagePath %}
+        {% if imagePathMobile %}
```
davidrapson: 👍

davidrapson:
👍 Got a conflict but it's the EYP template so can delete it.

PR #746, opened 2018-03-05
Revert "Client-side build upgrade"

davidrapson:
Reverts biglotteryfund/blf-alpha#730


davidrapson:
@mattandrews Reviewed this on test and it broke the materials form, so reverting this. Not sure why the tests passed though so going to take a look at that.

PR #747, opened 2018-03-05
Build update redo

davidrapson:
Re-do of https://github.com/biglotteryfund/blf-alpha/pull/730 but with a working build…

- All client-side dependency versions are upgraded
- Vue components code is code split so we load the minimum javascript up front
- Swiper is upgraded to v4 and code-split so the code isn't loaded outside of the homepage.


assets/js/modules/carousel.js
```diff
@@ -1,37 +1,17 @@
...
-    const settings = {
-        selector: '.js-carousel',
-        nextSelector: '.js-carousel-next',
-        prevSelector: '.js-carousel-prev'
-    };
+    const carouselElems = document.querySelectorAll('.js-carousel');
+    if (carouselElems.length > 0) {
+        const carouselSplit = () => import(/* webpackChunkName: "carousel" */ './carousel.split');
+        carouselSplit().then(carousel => {
+            console.log('Hello Carousel', carouselElems);
```
mattandrews: 👋

views/includes/metaFooterJS.njk
```diff
@@ -1,9 +1,10 @@
-<script src="https://www.google-analytics.com/cx/api.js"></script>
 <script src="{{ 'javascripts/app.js' | getCachebustedPath }}"
         id="js-script-main"
         data-ga-code="{{ appData.config.get('googleAnalyticsCode') }}"{% if ab %}
         data-ab-id="{{ ab.id }}"
         data-ab-variant="{{ ab.variantId }}"{% endif %}></script>
 
+
+<script async src="https://www.google-analytics.com/cx/api.js"></script>
```
mattandrews: Does this work okay like this with the A/B tests?

davidrapson: Hmm, good point, will check.

PR #749, opened 2018-03-05
Add case studies to funding programme pages

davidrapson:
Add case studies to funding programme pages. Already exists in the API but weren't being displayed on the page yet.


views/pages/funding/programme-detail.njk
```diff
@@ -72,5 +72,11 @@
                 </div>
             {% endif %}
         </div>
+
+        {# Case Studies #}
+        {{ caseStudyCollection(
+            caseStudies = entry.caseStudies,
+            customSectionTitle = __('global.misc.projectExamples'
```
mattandrews: The linebreaks here make this a bit weird, can you put the closing `)` on the same line?

PR #754, opened 2018-03-07
Improve web-font loading

davidrapson:
This PR does some work to improve web-font loading using the [FOUT with a class](https://www.zachleat.com/web/comprehensive-webfonts/#fout-class) technique.

Adds a dedicated `head.js` file to the webpack config which is inlined into the page and uses [`FontFaceObserver`](https://fontfaceobserver.com/) and [`netlify/postcss-fout-with-a-class`](https://github.com/netlify/postcss-fout-with-a-class) for font-loading.

Result is deferred loading of web-fonts on first-load, reducing FOIT.


assets/sass/_vars.scss
```diff
@@ -39,6 +39,16 @@ $accentBorderWidth: 4px;
 // (and also referenced in bordered grid)
 $contentBoxHorizontalPadding: 20px;
 
+/* =========================================================================
+   Global Variables: Font Stacks
+   ========================================================================= */
+
+$font-stacks: (
```
mattandrews: Maybe add a comment here to remind us that changes here should be reflected in `assets/js/head.js` too?

davidrapson:
> Looks good. Any before/after numbers for perf improvements?

Not the most scientific test but capturing timeline screenshots throttled to "Fast 3G" in Chrome show text on the page at 2.11s vs 3.11s, so ~1s improvement. Can probably get some more realistic numbers by doing a comparison of test vs production through webpagetest

**Before**
screen shot 2018-03-07 at 11 06 24

**After**
screen shot 2018-03-07 at 11 07 22

PR #758, opened 2018-03-07
Fix data map

mattandrews:
I just noticed that the [data map](https://www.biglotteryfund.org.uk/data) was broken – this should restore it.


davidrapson:
👍 Good catch (snyk is moaning though 🤷‍♂️ )

PR #760, opened 2018-03-07
Delete outstanding header code

davidrapson:
Thanks to @LynseyJReynolds we can now delete a whole bunch of unused code for the old hero images. 💪

### DIFF
--- a/controllers/funding/programmes.js
+++ b/controllers/funding/programmes.js
@@ -7,7 +7,7 @@ const Raven = require('raven');
 const { assign, find, findIndex, get, last, uniq } = require('lodash');
 
 const { splitPercentages } = require('../../modules/ab');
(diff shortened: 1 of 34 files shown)


controllers/funding/programmes.js
```diff
@@ -140,21 +140,11 @@ function initProgrammesList(router, options) {
 function renderProgrammeDetail({ res, entry }) {
     addPreviewStatus(res, entry);
 
-    /**
-     * Allow for programmes without heroes
-     */
-    const defaultProgrammeHeroImage = createHeroImage({
```
davidrapson: This picture no longer exists, using `heroImages.rathlinIslandDevelopment` as it doesn't have any people in it so is the safest as a fallback. We should never really show this anyway as all programmes should have their own image.

PR #762, opened 2018-03-07
Add a basic sitemap

davidrapson:
This might be a little pre-emptive but this PR reuses the code from `/status/pages` to build a basic `sitemap.xml` file. This can only cover application and CMS pages, not legacy pages so it's not complete. But as there isn't a `/sitemap.xml` on the legacy site and https://www.biglotteryfund.org.uk/sitemap is patchy it might still be worth adding this in as is.

### DIFF
--- a/controllers/route-helpers.js
+++ b/controllers/route-helpers.js
@@ -1,0 +1,85 @@
+const { compact, compose, concat, filter, flatMap, getOr, map, pick, sortBy, uniqBy } = require('lodash/fp');
+const contentApi = require('../services/content-api');
+const routes = require('./routes');
+
+const sortedUniqByPath = compose(sortBy('path'), uniqBy('path'));
+const isLive = route => route.live === true;
+
+/**
+ * Build a flat list of all canonical routes
+ * Combines application routes and routes defined by the CMS
+ */
+function getCanonicalRoutes({ includeDraft = false } = {}) {
+    const routerCanonicalUrls = flatMap(section => {
+        const withoutWildcards = filter(_ => _.path.indexOf('*') === -1);
+        const mapSummary = map((page, key) => {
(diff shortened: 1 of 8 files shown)


controllers/toplevel/index.js
```diff
@@ -208,6 +210,25 @@ module.exports = (pages, sectionPath, sectionId) => {
         res.send(text);
     });
 
+    router.get('/sitemap.xml', sMaxAge('30m'), (req, res) => {
+        contentApi.getCanonicalUrls().then(canonicalUrls => {
+            const sitemapInstance = sitemap.createSitemap({
+                hostname: getBaseUrl(req),
+                urls: canonicalUrls.map(route => ({
```
mattandrews: Does this need a filter to remove non-live URLs?

davidrapson:
@mattandrews Mind reviewing this again, reworked things a fair bit to keep more of the together and to more clearly handle draft/live status. Draft routes now excluded from sitemap.

controllers/route-helpers.js
```diff
@@ -0,0 +1,85 @@
...
+                live: page.live
+            };
+        });
+
+        return compose(mapSummary, withoutWildcards)(section.pages);
+    })(routes.sections);
+
+    return contentApi.getRoutes().then(cmsCanonicalUrls => {
+        const combined = concat(routerCanonicalUrls, cmsCanonicalUrls);
+        const filtered = includeDraft === true ? combined : combined.filter(isLive);
```
mattandrews: Could just do:
`const filtered = (includeDraft) ? combined : combined.filter(isLive);`

davidrapson: Personal preference but I like the explicitness

PR #766, opened 2018-03-08
Move sitemap out of devDependencies and into dependencies

davidrapson:
🙈


mattandrews:
D'oh. Wonder if we could've caught this in Travis?

davidrapson:
> D'oh. Wonder if we could've caught this in Travis?

Possibly by stripping `devDependencies` earlier.

PR #772, opened 2018-03-09
Improve how isBilingual is defined

davidrapson:
Improve how `isBilingual` is defined so that we define an initial `app.locals` value of `true` and then override it as needed on a per route basis. As a result allows us to do fix this TODO: `todo: <link rel="alternate" hreflang="cy" href="#" />`

### DIFF
--- a/modules/viewEngine.js
+++ b/modules/viewEngine.js
@@ -6,15 +6,24 @@ const slugify = require('slugify');
 const assets = require('./assets');
 const appData = require('./appData');
 
-function init(app) {
-    const templateEnv = nunjucks.configure('views', {
-        autoescape: true,
-        express: app,
-        noCache: appData.isDev,
-        watch: appData.isDev
-    });
+/**
+ * Define common app locals
+ * @see https://expressjs.com/en/api.html#app.locals
+ */
+function initAppLocals(app) {
+    /**
+     * Is this page bilingual?
+     * i.e. do we have a Welsh translation
+     * Default to true unless overriden by a route
+     */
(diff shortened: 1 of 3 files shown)


modules/viewEngine.js
```diff
@@ -59,28 +67,6 @@ function init(app) {
         return str.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
     });
 
-    // via http://stackoverflow.com/a/25770787
-    templateEnv.addFilter('splitByCharLength', (str, length) => {
```
davidrapson: This was unused.

mattandrews: 👍

views/includes/metadata.njk
```diff
@@ -49,4 +49,10 @@
 <link rel="icon" type="image/png" sizes="16x16" href="{{ 'favicon/favicon-16x16.png' | getImagePath }}">
 <meta name="msapplication-TileImage" content="{{ 'favicon/ms-icon-144x144.png' | getImagePath }}">
 
-{# todo: <link rel="alternate" hreflang="cy" href="#" /> #}
+{% if isBilingual %}
+    {% if locale == 'en' %}
+        <link rel="alternate" hreflang="cy" href="{{ getCurrentUrl(request, 'cy') }}" />
+    {% else %}
+        <link rel="alternate" hreflang="en" href="{{ getCurrentUrl(request, 'en') }}" />
```
mattandrews: giphy-tumblr

PR #776, opened 2018-03-09
Dynamic case studies

davidrapson:
This PR allows case studies to be fetched dynamically from the CMS. Using the 10k pages as the first use case.

- Adds `getCaseStudies` method, and cleans up language merging in `content-api`
- Moves 10k templates to funding section
- Updates 10k routes to by dynamic and fetch case studies from the CMS

screen shot 2018-03-09 at 12 28 39

### DIFF
--- a/controllers/funding/10k.js
+++ b/controllers/funding/10k.js
@@ -1,0 +1,51 @@
+'use strict';
+
+const contentApi = require('../../services/content-api');
+
+function init10k({ router, routeConfig, caseStudySlugs }) {
+    router.get(routeConfig.path, (req, res) => {
+        const lang = req.i18n.__(routeConfig.lang);
+        const currentLocale = req.i18n.getLocale();
+
+        const servePage = caseStudies => {
+            res.render(routeConfig.template, {
+                title: lang.title,
+                description: lang.description || false,
+                copy: lang,
+                caseStudies: caseStudies || []
+            });
+        };
+
+        contentApi
+            .getCaseStudies({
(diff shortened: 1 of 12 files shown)


controllers/funding/10k.js
```diff
@@ -0,0 +1,51 @@
...
+        contentApi
+            .getCaseStudies({
+                locale: currentLocale,
+                slugs: caseStudySlugs
+            })
+            .then(
+                caseStudies => {
+                    servePage(caseStudies);
+                },
+                function() {
```
mattandrews: What's this, the error handler? Bit confusing to read (or might be because I need a coffee)?

views/pages/funding/under10k.njk
```diff
@@ -19,7 +19,7 @@
         )
     }}
 
-    <article role="main" class="nudge-up">
+    <main class="nudge-up">
```
mattandrews: I think the `role` here does some useful a11y stuff, or does using `<main>` do the same thing?

davidrapson: `<main>` does the same thing afaik. It has the implicit role of `main` whereas `<article role="main">` is being assigned a role of `main`…if that makes sense.

controllers/funding/10k.js
davidrapson: ☕️second argument to `.then` is the error handler ☕️

mattandrews: I'm going back to bed.

PR #783, opened 2018-03-13
Fix Welsh links on legacy pages

mattandrews:
Due to quirks of Sitecore, proxying and cookies, links on legacy pages between English/Welsh are sporadic in appearance. This code tweaks our proxying technique to append these links if found.

Some pages already have Welsh/English links and don't need them to be appended, eg:

- **Press release**: [English](https://www.biglotteryfund.org.uk/global-content/press-releases/wales/060318_wal_a4aw_release_march18) vs [Welsh](https://www.biglotteryfund.org.uk/welsh/global-content/press-releases/wales/060318_wal_a4aw_release_march18)
(eg. this content when proxied already has the links)

... but some content is available in Welsh but has no discernible link to switch language:

- **Research: Open Data**: [English](https://www.biglotteryfund.org.uk/research/open-data) vs [Welsh](https://www.biglotteryfund.org.uk/welsh/research/open-data)
(eg. this content when proxied does not have the links)

This PR checks to see whether the page is missing its opposite language link, and if so, adds it.

### DIFF
--- a/modules/legacy.js
+++ b/modules/legacy.js
@@ -40,19 +40,14 @@ const proxyLegacyPage = ({ req, res, domModifications, followRedirect = true }) 
             dom.window.document.title = '[PROXIED] ' + titleText;
         }
 
+        const regionNav = dom.window.document.getElementById('regionNav');
         const pageIsWelsh = isWelsh(req.path);
-        const hasWelshLink = dom.window.document.querySelectorAll('[hreflang="cy"]').length > 0;
-        const hasEnglishLink = dom.window.document.querySelectorAll('[hreflang="en"]').length > 0;
+        const hasWelshLink = regionNav.querySelectorAll('[hreflang="cy"]').length > 0;
+        const hasEnglishLink = regionNav.querySelectorAll('[hreflang="en"]').length > 0;
 
-        const appendLanguageLink = localeToAppend => {
-            let linkPath, linkText;
-            if (localeToAppend === 'cy') {
-                linkPath = makeWelsh(req.path);
-                linkText = 'Cymraeg';
-            } else {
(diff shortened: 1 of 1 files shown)


mattandrews:
Hmm, confusingly, when I load the press release page above on the live site, I don't see the Cymraeg link, but I do see it on `TEST` and `DEV`. Possibly another cookie quirk / proxying server location issue, but I can confirm that locally, this code doesn't add a second language link when there's an existing one.

modules/legacy.js
```diff
@@ -39,6 +40,41 @@ const proxyLegacyPage = ({ req, res, domModifications, followRedirect = true })
             dom.window.document.title = '[PROXIED] ' + titleText;
         }
 
+        const pageIsWelsh = isWelsh(req.path);
+        const hasWelshLink = dom.window.document.querySelectorAll('[hreflang="cy"]').length > 0;
```
davidrapson: Worth scoping these selectors to `regionNav` too?

modules/legacy.js
```diff
@@ -39,6 +40,41 @@ const proxyLegacyPage = ({ req, res, domModifications, followRedirect = true })
...
+        const hasEnglishLink = dom.window.document.querySelectorAll('[hreflang="en"]').length > 0;
+
+        const appendLanguageLink = localeToAppend => {
+            let linkPath, linkText;
```
davidrapson: Maybe ternary statements for conciseness?

``` js
const linkPath = localeToAppend === 'cy' ? makeWelsh(req.path) : 'English';
const linkText = localeToAppend === 'cy' ? 'Cymraeg' : 'English';
```

davidrapson:
👍

PR #787, opened 2018-03-16
Inject hero images

davidrapson:
Add api method and middleware to allow hero images to be injected from the CMS. Allows hero images to be stored in a single place in the CMS and means hero images are consistently optimised through Imgix. Updates `/funding`, `/research` and `/about` to use images from the CMS.

### DIFF
--- a/middleware/inject-hero.js
+++ b/middleware/inject-hero.js
@@ -1,0 +1,29 @@
+const Raven = require('raven');
+const { get } = require('lodash/fp');
+const contentApi = require('../services/content-api');
+const { defaultHeroImage } = require('../modules/images');
+
+module.exports = function injectHeroImage(page) {
+    const heroSlug = get('heroSlug')(page);
+    return function(req, res, next) {
+        if (heroSlug) {
+            contentApi
+                .getHeroImage({
+                    locale: req.i18n.getLocale(),
+                    slug: heroSlug
+                })
+                .then(heroImage => {
+                    res.locals.heroImage = heroImage;
+                    next();
+                })
+                .catch(err => {
+                    Raven.captureException(err);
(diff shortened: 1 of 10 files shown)


middleware/inject-hero.js
```diff
@@ -0,0 +1,32 @@
...
+const { defaultHeroImage } = require('../modules/images');
+
+module.exports = function injectHeroImage(page) {
+    const heroSlug = get('heroSlug')(page);
+    return function(req, res, next) {
+        if (heroSlug) {
+            contentApi
+                .getHeroImage({
+                    locale: req.i18n.getLocale(),
+                    slug: heroSlug
```
mattandrews: hero-slug

middleware/inject-hero.js
```diff
@@ -0,0 +1,32 @@
...
+                })
+                .then(heroImage => {
+                    res.locals.heroImage = heroImage;
+
+                    console.log(heroImage);
```
mattandrews: Can remove this.

views/pages/toplevel/about.njk
```diff
@@ -5,7 +5,6 @@
 
 {% block title %}{{ title }} | {% endblock %}
 
-{% set heroImage = heroImages.mentalHealthFoundation %}
```
mattandrews: Worth removing this (and the others) from the hero config file (and the images themselves) to avoid confusion?

davidrapson: Yep, but I'll come back and do this afterwards as a bit of cleanup. The images themselves can go then too.

davidrapson:
FYI: this will probably fail until the endpoint exists.

PR #789, opened 2018-03-16
News link

davidrapson:
image

### DIFF
--- a/assets/sass/components/news.scss
+++ b/assets/sass/components/news.scss
@@ -1,0 +1,29 @@
+/* =========================================================================
+   News
+   ========================================================================= */
+
+// .article-teaser {}
+.article-teaser__summary {
+    font-size: 18px;
+}
+
+.latest-news {
+    position: relative;
+}
+.latest-news__footer {
+    margin-bottom: $spacingUnit;
+
+    @include mq('medium') {
+        position: absolute;
+        top: 0;
+        right: 0;
+        margin: 0;
+    }
+}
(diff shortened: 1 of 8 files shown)


views/components/news.njk
```diff
@@ -13,3 +13,28 @@
...
     </article>
 {% endmacro %}
+
+{% macro latestNews(articles) %}
+    <div class="latest-news">
+        <div class="latest-news__header">
+            <h3 class="t1 t--underline a--text accent--blue">
+                <a class="u-link-minimal" href="{{ buildUrl("uk-wide/news-and-events") }}">
+                    {{ __('global.misc.latestNews') }}</h3>
+                </a>
```
mattandrews: Missing a closing `</h3>` tag here?

assets/sass/components/news.scss
```diff
@@ -0,0 +1,29 @@
+/* =========================================================================
+   News
+   ========================================================================= */
+
+// .article-teaser {}
```
mattandrews: Could delete this?

davidrapson: Could, but I kind of like an explicit comment saying what the parent class  is if it doesn't have any styles attached.

views/components/news.njk
davidrapson: Good catch

davidrapson: (This has been corrected, but the comment is on the `</a>` so the comment isn't dismissed.

PR #792, opened 2018-03-19
Update text/design for Reaching Communities emails

mattandrews:
This adds some formatting to the HTML emails and uses newly-supplied text.

One thing to note – I tested this with various encoding challenges (eg. words like `Dharmaśāstra`) and was seeing encoding errors when retrieving them from the session. I traced this to the database encoding itself and subsequently discovered that  `utf8mb4` [seems to be the correct MySQL encoding](https://mathiasbynens.be/notes/mysql-utf8mb4) to preserve these characters when storing. I've changed the `data` rows on the `sessions` table in each database environment to use this encoding, but it's worth remembering for any future data storage fields where we might be needing characters with accents etc.

### DIFF
--- a/assets/sass/emails.scss
+++ b/assets/sass/emails.scss
@@ -53,3 +53,24 @@ blockquote {
         width: 150px;
     }
 }
+
+.data__step {
+    border-bottom: 1px solid palette('pink');
+    padding-bottom: 3px;
+    font-weight: bold;
+}
+
+.data__label {
+    font-weight: bold;
+}
+
+small {
+    display: block;
+    font-size: 11px;
+    color: #666666;
+    margin-bottom: 10px;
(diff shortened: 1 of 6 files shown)


davidrapson:
> I traced this to the database encoding itself and subsequently discovered that utf8mb4 seems to be the correct MySQL encoding to preserve these characters when storing. I've changed the data rows on the sessions table in each database environment to use this encoding, but it's worth remembering for any future data storage fields where we might be needing characters with accents etc.

Is there a way we can codify or document this somewhere? In case we need to spin up a new database or something.

mattandrews:
Good call – have added a default config to use this encoding when initialising a new database (via https://github.com/sequelize/sequelize/issues/7110).

PR #798, opened 2018-03-21
Debug search logs

mattandrews:


davidrapson:
👍 That fix makes sense to me. If we need to do this more then abstracting this might make sense, but looks good for now.

PR #804, opened 2018-03-22
Inject heroes cleanup

davidrapson:
- Fetch region page images from the cms, which removes the concept of `page.heroImage` and fixes fetching of images for these pages. 
- Remove all unused image definitions that are now fetched from the cms


davidrapson:
Will keep an eye on the changed one and see if purging the cache fixes it, thankfully it's a generic image (i.e. not a programme page) so I feel OK about testing how viable it is to purge it.

PR #806, opened 2018-03-22
RC form tweaks

mattandrews:
This still needs the arrow button, but free to review otherwise.

This change contains: 

- tweaks to some Unusual Capitalisation In Forms
- adds RC branding/title to pages
- Adds a few fields (project location and additional organisations)

Have at it!

### DIFF
--- a/assets/sass/components/buttons.scss
+++ b/assets/sass/components/buttons.scss
@@ -249,6 +249,22 @@ $btnConf: (
     }
 }
 
+
+
+.btn--next {
+    background-image: arrow-btn();
+    background-size: contain;
+    background-repeat: no-repeat;
+    background-position: center left;
+    min-width: 158px;
+    background-color: transparent;
+    @include on-interact {
+        background-color: transparent;
+        background-image: arrow-btn(darken(palette('blue'), 15%));
+    }
+}
+
+
 /* =========================================================================
    Button Group
    ========================================================================= */
--- a/controllers/apply/forms/reaching-communities-form.js
+++ b/controllers/apply/forms/reaching-communities-form.js
@@ -218,40 +218,44 @@ formModel.registerSuccessStep({
              * Render a Nunjucks template to a string and
              * convert the string to inline HTML
              */
-            app.render('emails/applicationSummary', {
-                summary: summary,
-                form: formModel,
-                data: flatData
-            }, (errorRenderingTemplate, html) => {
+            app.render(
+                'emails/applicationSummary',
+                {
+                    summary: summary,
+                    form: formModel,
+                    data: flatData
+                },
+                (errorRenderingTemplate, html) => {
+                    if (errorRenderingTemplate) {
+                        return reject(errorRenderingTemplate);
+                    }
 
-                if (errorRenderingTemplate) {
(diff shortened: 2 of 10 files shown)


controllers/apply/forms/reaching-communities-form.js
```diff
@@ -203,40 +225,44 @@ formModel.registerSuccessStep({
...
-            app.render('emails/applicationSummary', {
-                summary: summary,
-                form: formModel,
-                data: flatData
-            }, (errorRenderingTemplate, html) => {
-
-                if (errorRenderingTemplate) {
-                    return reject(errorRenderingTemplate);
-                }
+            app.render(
```
mattandrews: Every change to this file below this line is a `prettier`/whitespace tweak, not a proper change.

controllers/apply/forms/reaching-communities-form.js
```diff
@@ -100,70 +100,92 @@ formModel.registerStep({
                             }
                         });
                     }
+                },
+                {
+                    type: 'text',
+                    name: 'project-location',
+                    label: "In your own words, describe the location(s) that you'll be running your project(s) in",
+                    placeholder: 'eg. "Newtown community centre" or "Alfreton, Derby and Ripley"',
```
davidrapson: It's a little odd to have a Newtown as an example when this programme is for England only.

controllers/apply/forms/reaching-communities-form.js
```diff
@@ -100,70 +100,92 @@ formModel.registerStep({
...
                     validator: function(field) {
                         return check(field.name)
                             .trim()
                             .not()
                             .isEmpty()
                             .withMessage('Organisation must be provided');
                     }
+                },
+                {
+                    type: 'text',
```
davidrapson: Does this do anything yet?

mattandrews: How do you mean? It outputs an `<input type="text" />` field...?

assets/sass/components/buttons.scss
```diff
@@ -249,6 +249,22 @@ $btnConf: (
     }
 }
 
+
+
+.btn--next {
+    background-image: arrow-btn();
```
davidrapson: Thinking about it, maybe css clip-paths are the way to go here. e.g. http://bennettfeely.com/clippy/
That's the way all the wonky stuff on https://membership.theguardian.com/ was done.

davidrapson:
👍 Changes looks good, parameter store stuff makes sense. Would be good to come back and get some more tests in for this code, but good to merge.

assets/sass/components/buttons.scss
mattandrews: Just looking into this. Here's the best I could manage:
image

(using `clip-path: polygon(0% 0%, 80% 0%, 95% 50%, 80% 100%, 0% 100%)`)

The original one, though, has a rounded edge on the arrow:

image

[Apparently](https://stackoverflow.com/questions/31765345/how-to-round-out-corners-when-using-css-clip-path), though, you can't easily round corners using CSS clip-path (other than literally draw each point in the corner). Not sure how much of a compromise we can make here.

PR #811, opened 2018-03-27
Add the "AWSELB" cookie

mattandrews:
We set session stickiness for the load balancer so a repeat visitor gets the same instance (for 5 minutes), but there were two issues:

1. We only set this for the HTTP listener (now rectified so the HTTPS one gets it too)
2. We drop the cookie at Cloudfront so nobody ever got this set

This PR adds back that cookie which might help the (admittedly rare) edge case where someone visits the site during a deploy and gets broken CSS/JS (eg. because the load balancer routes their assets request to the new instance with different hashes).

### DIFF
--- a/config/cloudfront/live.json
+++ b/config/cloudfront/live.json
@@ -49,9 +49,10 @@
             "contrastMode",
             "blf-alpha-session",
             "_csrf",
-            "blf-ab-afa-v2"
+            "blf-ab-afa-v2",
+            "AWSELB"
           ],
-          "Quantity": 4
+          "Quantity": 5
         }
       }
     },
@@ -114,9 +115,10 @@
                 "contrastMode",
                 "blf-alpha-session",
                 "_csrf",
-                "blf-ab-afa-v2"
+                "blf-ab-afa-v2",
+                "AWSELB"
               ],
-              "Quantity": 4
+              "Quantity": 5
             }
           }
         },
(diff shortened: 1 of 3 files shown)


davidrapson:
Oh, interesting. Just want to make sure I'm following. My understanding of why someone might get an old asset hash is something in this process:

- During the period where both the new and old servers are in the load-balancer during deployment
- A page's cache expires, the fresh request has a chance of hitting one of the old servers and a page is cached with the old asset hash
- The old servers get removed and now the assets for the old hash no longer exists but the old page is in the cache so points to broken assets.

I'm not 100% sure that's correct but is something in that period right? The bit I'm not following is:

- If it's a reference to an old hash why don't the old assets stick around in Cloudfront given they have a ~1 year cache
- Why does sticky sessions solve this? Even if it improves it does the problem not still exist for new-visitors?


Keen to dig into this as it would be nice to avoid the need for any statefulness like sticky sessions.

davidrapson:
Scratch that, reverse it. The process is more likely to be:

- During the period where both the new and old servers are in the load-balancer during deployment
- A page's cache expires, the fresh request hits the new servers and caches the page with the new hash
- Whilst the old servers still exist, initial requests for assets with the new hash have a chance of being routed to the old servers which don't have any files for that hash resulting in a 404.

That sounds like a more likely series of steps, right?

If so, is the problem solved by sticky session here too? Sticky sessions will apply to the "page" request where cookies are passed. But the assets are going to be stateless requests, so are they not going to have the same problem?

davidrapson:
Closing in favour of https://github.com/biglotteryfund/blf-alpha/pull/816

PR #812, opened 2018-03-27
Add field explanations for checkboxes

mattandrews:


views/pages/apply/reaching-communities-startpage.njk
```diff
@@ -64,7 +64,7 @@
         <div class="u-constrained-content s-prose">
             <h2 class="t2 t--underline accent--pink">Other ways to apply</h2>
             <p>
-                We’re available every day from 5am to midnight to discuss
+                We’re available every day from 9am to 5pm to discuss
```
davidrapson: Good catch. Should this be something like "We’re available Monday–Friday from 9am to 5pm to discuss". Week days vs. everyday?

mattandrews: Good call, just changed.

mattandrews:
Added some changes to highlight required/optional fields:

image

davidrapson:
👍

PR #817, opened 2018-03-29
Regression testing

davidrapson:
This PR sets up some tests with [cypress](https://www.cypress.io/) which is a testing tool somewhere in-between our current integration test approach and something like Selenium. Lets us write tests in different styles so can mix and match simpler network tests with fuller interaction/e2e tests. These tests do two main things:

1. Adds a new test to interact with the mobile nav and confirm it works
2. Transfers over some of the existing integration tests to cypress as an example


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
mattandrews: Do we need this file?

cypress/integration/interactions_spec.js
```diff
@@ -0,0 +1,8 @@
+describe('Interaction tests', function() {
+    it('should toggle mobile navigation', function() {
+        cy.viewport(320, 480);
+        cy.visit('/');
+        cy.get('#js-mobile-nav-toggle').click();
+        cy.get('#qa-offscreen-navigation').should('be.visible');
```
mattandrews: 💯

cypress/plugins/index.js
```diff
@@ -0,0 +1,17 @@
...
+// https://on.cypress.io/plugins-guide
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+module.exports = (on, config) => {
+    // `on` is used to hook into various events Cypress emits
+    // `config` is the resolved Cypress config
+};
```
mattandrews: Is this whole file a placeholder in case we want a plugin?

davidrapson:
"Not sure we need the unused boilerplate but all good otherwise." Same, but deleting them seemed to re-add them when running the tests so leaving them.

PR #819, opened 2018-03-29
Migrate customer service

mattandrews:
Swaps out a couple of hard-coded static pages into CMS routes, and adds half a dozen more.

Also adds full "segment" support so a CMS-defined segment can output an image:

image

I'm about to import the content itself into the live CMS, but other than that this should be good to go.


views/components/segment.njk
```diff
@@ -2,15 +2,22 @@
...
+                <img src="{% if imageIsAbsolute %}{{ imagePath }}{% else %}{{ imagePath | getImagePath }}{% endif %}" alt="" />
+                {{ headers.overlayText(
+                    tag = 'h4',
+                    text = title,
+                    additionalClasses = 'segment__title grid__item__title grid__item__title--top',
+                    id = id
+                )}}
+            </div>
+        {% endif %}
+        <div class="segment__aside">
```
davidrapson: Could we add the `.s-prose` classes to this.

davidrapson: Worth comparing with https://www.biglotteryfund.org.uk/about/strategic-framework

PR #837, opened 2018-04-05
Improve form submission handling

davidrapson:
When `process.env.DONT_SEND_EMAIL` is set we were returning an object where we normally return a promise. This was breaking the confirmation screen of the reaching communities form when under tests. To fix this, we now always return a promise allowing any dependent steps to happen even if we are skipping sending the actual email.

Edit: Updated the router further to serve a real error screen if there is a problem processing submissions.

screen shot 2018-04-06 at 09 20 58

### DIFF
--- a/controllers/apply/forms/create-form-router.js
+++ b/controllers/apply/forms/create-form-router.js
@@ -3,7 +3,6 @@ const { get, isEmpty, set, unset } = require('lodash');
 const { validationResult } = require('express-validator/check');
 const { matchedData } = require('express-validator/filter');
 const cached = require('../../../middleware/cached');
-const { renderError } = require('../../http-errors');
 
 module.exports = function(router, formModel) {
     const formSteps = formModel.getSteps();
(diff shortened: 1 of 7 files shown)


davidrapson:
@mattandrews Mind giving this another look 👀

controllers/apply/forms/reaching-communities-form.js
```diff
@@ -297,12 +296,6 @@ formModel.registerSuccessStep({
                                 .then(() => resolve(formData));
                         })
                         .catch(err => {
-                            Raven.captureMessage('Error converting template to inline CSS', {
-                                extra: err,
-                                tags: {
-                                    feature: 'reaching-communities'
-                                }
-                            });
                             return reject(err);
```
mattandrews: Does this mean we won't know anymore if the email rendering failed?

controllers/apply/forms/create-form-router.js
```diff
@@ -92,18 +94,23 @@ module.exports = function(router, formModel) {
...
+                        res.render('pages/apply/success', {
+                            form: formModel,
+                            success: successStep
+                        });
                     });
                 })
                 .catch(err => {
-                    // @TODO handle this
-                    res.send(err);
+                    err.friendlyText = 'There was a problem processing your submission';
```
mattandrews: If we get here it's because something in the `processor` failed (eg. it's our fault, not the user's). Maybe we could show some text giving them tips on what to do? eg. their session data should be preserved so we could add a button saying "Try again" or something which sends them back to the review step? No guarantees it'll work a second time but it feels like we could at least offer them some sort of onward step?

controllers/apply/forms/reaching-communities-form.js
davidrapson: It gets captured in the form-router catch step instead.

controllers/apply/forms/create-form-router.js
davidrapson: Good shout

PR #841, opened 2018-04-06
Migrate tools test, add seed endpoints

davidrapson:
Builds off https://github.com/biglotteryfund/blf-alpha/pull/840

Migrates tools tests over to cypress. Adds seed endpoints when using a local database allowing new tests users to be created. As recommended in https://docs.cypress.io/guides/getting-started/testing-your-app.html#Seeding-Data


server.js
```diff
@@ -70,13 +70,14 @@ app.use(
...
+app.get('/status', require('./controllers/toplevel/status'));
+
+// Mount tools controller
+app.use('/tools', require('./controllers/tools'));
 
 // Mount apply controller (forms)
-const applyPath = '/apply';
-app.use(applyPath, require('./controllers/apply'));
-app.use(cymreigio(applyPath), require('./controllers/apply'));
+app.use(cymreigio('/apply'), require('./controllers/apply'));
```
davidrapson: This is me realising that `cymreigio` returns an array of paths, so we never needed the duplication.

cypress/support/commands.js
```diff
@@ -1,5 +1,46 @@
...
+        method: 'POST',
+        url: '/user/register',
+        failOnStatusCode: false,
+        followRedirects: false,
+        form: true,
+        body: formBody
+    });
+});
+
+Cypress.Commands.add('seeduser', () => {
```
mattandrews: Could maybe camel-case this (`seedUser`) to match the other names?

controllers/tools/seed.js
```diff
@@ -0,0 +1,25 @@
+const uuidv4 = require('uuid/v4');
+const userService = require('../../services/user');
+
+function init({ router }) {
+    if (!process.env.USE_LOCAL_DATABASE) {
+        return;
+    }
+
+    router.post('/seed/user', (req, res) => {
```
mattandrews: We could wrap this in an environment check too for safety so we never end up with a public route that creates a privileged user, but perhaps the `USE_LOCAL_DATABASE` flag check is enough?

PR #844, opened 2018-04-09
Update Raven loading

davidrapson:
Tweaks how Sentry is loaded based on the docs. Passing a falsely value for the DSN will still configure the client, but not send any events. Potentially improves workflow in CI/test environments if Sentry DSN is not set but an explicit call to `captureException` is sent. (Could allow us to remove DSN from CI environment)

### DIFF
--- a/server.js
+++ b/server.js
@@ -32,18 +32,25 @@ const localesMiddleware = require('./middleware/locales');
 const { noCache } = require('./middleware/cached');
 const previewMiddleware = require('./middleware/preview');
 
-if (SENTRY_DSN) {
-    Raven.config(SENTRY_DSN, {
-        environment: appData.environment,
-        dataCallback(data) {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -32,18 +32,25 @@ const localesMiddleware = require('./middleware/locales');
...
-    app.use(Raven.requestHandler());
-}
+/**
+ * Configure Sentry client
+ * https://docs.sentry.io/clients/node/config/
+ * https://docs.sentry.io/clients/node/usage/#disable-raven
+ */
+Raven.config(SENTRY_DSN, {
+    environment: appData.environment,
+    autoBreadcrumbs: true,
```
davidrapson: Enabled `autoBreadcrumbs` to the config, notably should log any HTTP requests that the express application makes leading up to an error.

PR #853, opened 2018-04-10
A11y fixes

mattandrews:
Adds some fieldsets/legends, and a few missing bits of missing file link info.


views/includes/surveyPlaceholder.njk
```diff
@@ -0,0 +1,3 @@
+{% from "../components/surveys.njk" import showSurvey with context %}
```
davidrapson: Tiny thing but not sure I follow why you are using the include?

PR #855, opened 2018-04-10
Update ebulltin from to use new form fields

davidrapson:
This started as a PR to add fieldsets to the ebulletin form but ended up converting it to use the new form macros. Visually it remains the same, aside from simplifying the display of radio options (always left-aligned, fight me 😄)

Tried to keep the commits granular enough to show what's going on.

The end-goal of this is to reduce the forked form logic we have and where possible favour the inputs used on the reaching communities form. This will likely mean isolating off the materials form (accepting it's completely custom) and standardising all other forms on the new inputs.

Still needs a little work on error messages, but otherwise there.


assets/sass/globals/forms.scss
```diff
@@ -149,7 +149,19 @@ $errorColor: #b00;
     padding: 0.5em;
 }
 
-.ff-radio,
+.ff-radio {
```
davidrapson: Mirrors checkbox styles, there are no radio options on the RC form so I think these were just never styled.

assets/sass/globals/forms.scss
```diff
@@ -171,6 +183,20 @@ $errorColor: #b00;
     }
 }
 
+.ff-field-pair {
```
davidrapson: Adds the concept of a pair of form fields. Avoids needing to fight against grid class margins.

config/locales/cy.yml
```diff
@@ -231,6 +231,7 @@ toplevel:
       wales: Cymru
       scotland: Yr Alban
       northernIreland: Gogledd Iwerddon
+    yourDetails: Eich Manylion
```
mattandrews: manylions

assets/sass/globals/forms.scss
```diff
@@ -296,99 +335,3 @@ $errorColor: #b00;
         @extend .u-visually-hidden;
     }
 }
-
-.form-ebulletin {
```
davidrapson: This class was never used so these styles never get triggered. Bye 👋

assets/sass/globals/forms.scss
```diff
@@ -218,38 +218,6 @@ $errorColor: #b00;
    Forms: Specific Form Customisations
    ========================================================================= */
 
-.form-fund {
```
davidrapson: This class was never used so these styles never get triggered. Bye 👋

views/components/ebulletin.njk
```diff
@@ -0,0 +1,93 @@
+{% from "components/forms.njk" import textInput, radios with context %}
```
mattandrews: Is this line still needed?

mattandrews:
Design regression here?

image


(before vs after – the pink location text is now black?)

davidrapson:
"Design regression here?"

Kinda intentional :trollface: in that the right aligned labels bug me, but no reason I can't reinstate.

Thinking about this, I think there's some more work to be done to share/reinstate some of the error message logic without peppering flash message stuff into the form. Going to timebox myself on this a bit and if it feels like a rabbit hole I might revisit.

davidrapson:
Ah yeah, colour could come back for now. Although I'd bet we'd need to increase the contrast on this at some-point anyway.

davidrapson:
🐰 🕳 Emerged successfully from the rabbit hole. Fairly happy with the end result.

Now allows form field macros from RC form to be used in a standalone context. Refactored the ebulletin form to avoid the use of `flash` so no session storage needed. Error messages are improved and use the shared style:

image

controllers/about/ebulletin.js
```diff
@@ -2,144 +2,149 @@
...
+                    formErrors: formErrors.array()
+                });
+            } else {
+                const handleSignupSuccess = () => {
+                    customEvent('emailNewsletter', 'signup', formData.location);
+                    renderForm({
+                        res: res,
+                        formStatus: FORM_STATES.SUBMISSION_SUCCCESS,
+                        formData: formData,
+                        formErrors: formErrors.array()
```
mattandrews: Do you still need to pass this (empty) array in the valid state?

modules/forms.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Enum of form submission states
+ */
+const FORM_STATES = {
+    NOT_SUBMITTED: 'NOT_SUBMITTED',
+    VALIDATION_ERROR: 'VALIDATION_ERROR',
+    SUBMISSION_ERROR: 'SUBMISSION_ERROR',
+    SUBMISSION_SUCCCESS: 'SUCCESS'
+};
```
mattandrews: This is nice, good pattern. 💯

views/components/ebulletin.njk
```diff
@@ -0,0 +1,99 @@
...
+                {{ __('toplevel.ebulletin.yourDetails') }}
+            </legend>
+            <div class="form-fieldset__fields">
+                <div class="ff-field-pair">
+                    <div class="ff-field-pair__item">
+                        {{ renderField({
+                            type: 'text',
+                            name: 'firstName',
+                            label: __('toplevel.ebulletin.fields.firstName'),
+                            placeholder: __('toplevel.ebulletin.fields.firstName') + ' *',
```
mattandrews: I wonder if this is better handled by checking `isRequired` in one of the field macros and appending the asterisk there?

davidrapson: Good call

PR #857, opened 2018-04-11
Axe upgrade fixes

davidrapson:
See: https://github.com/biglotteryfund/a11y-tests/pull/1

Fixes warnings reported by latest version of aXe cli.

Only mildly controversial change is that I _slightly_ darkened the pink we use as it was so close to passing at AA+ that it seems silly not to push it over the edge. Pulled the luminance down 5% so shouldn't be noticeable.

Best viewed with https://github.com/biglotteryfund/blf-alpha/pull/857/files?w=1


views/components/tabs.njk
```diff
@@ -1,14 +1,14 @@
 {% from "components/icons.njk" import iconArrowDown %}
 
 {% macro tab(title, paneId, tabClasses = false, setActive = false, id = false) %}
-    <a class="js-tab{% if tabClasses %} {{ tabClasses }}{% else %} tab{% endif %}{% if setActive %} tab--active{% endif %}"
+    <a role="tab" class="js-tab{% if tabClasses %} {{ tabClasses }}{% else %} tab{% endif %}{% if setActive %} tab--active{% endif %}"
```
davidrapson: This ~might~ break on the data page where tabs are used out of context, but improves things overall so will come-back and fix that if needs be.

PR #869, opened 2018-04-13
Apply form refinement

davidrapson:
Makes a few different improvements to the application forms

### Small screen refinements

Update markup and styles to improve display of forms on small screens. Notably excess spacing around content boxes and wrapping of checkbox labels

### Accessibility fixes

Went through all screens manually using aXe and fixes any warnings. Main change was to checkbox markup to explicitly wrap the options in a fieldset. Necessitated a small change in labelling for these fields.

Before:
image

After:
screen shot 2018-04-13 at 09 43 23

Also makes a small change to the "step 1 of n" and "optional" labels to improve text contrast

### No more destructive actions on GET

Noticed that the way we were processing the submit step from the review screen was via a GET request which is not ideal as it means we're performing a destructive action (sending an email) on a GET request. This PR refactors the router to process the form on the *POST* action of the review form.

### DIFF
--- a/views/pages/apply/review.njk
+++ b/views/pages/apply/review.njk
@@ -5,37 +5,48 @@
 {% set bodyClass = 'has-static-header' %}
 
 {% block content %}
+    {% set pageAccent = 'pink' %}
     <main role="main" id="content">
-        <div class="content-box inner inner--wide-only accent--pink a--border-top">
+        <form action="" method="POST"
+            class="js-application-form js-application-form-review
+            content-box u-inner-wide-only accent--{{ pageAccent }} a--border-top">
             <div class="u-constrained-content">
                 <h1 class="t1">{{ review.title }}</h1>
                 {% for step in summary %}
-                    <h2 class="t2 t--underline accent--pink">
+                    <h2 class="t2 t--underline accent--{{ pageAccent }}">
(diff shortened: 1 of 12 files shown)


views/pages/apply/review.njk
```diff
@@ -1,41 +1,56 @@
...
-                                {% if field.value %}
-                                    <dt>{{ field.label }}</dt>
-                                    {% if field.type === 'textarea' %}
-                                        <dd class="s-prose" data-hj-suppress>{{ field.value | striptags(true) | escape | nl2br }}</dd>
-                                    {% elseif field.type === 'checkbox' %}
-                                        <dd data-hj-suppress>{{ field.value | joinIfArray(', ') }}</dd>
-                                    {% else %}
-                                        <dd data-hj-suppress>{{ field.value }}</dd>
-                                    {% endif %}
+                    <dl class="{{ classNames({
```
davidrapson: This was me playing around with using https://www.npmjs.com/package/classnames in templates to clean up conditional class logic. It works, but is a bit of a tangent so will pull it out of this PR.

davidrapson:
@mattandrews Done. Remove the `beforeunload` message from this PR as it needs a little more thought.

PR #873, opened 2018-04-13
Add before unload message to review page

davidrapson:
screen shot 2018-04-12 at 17 59 44

Adds a `beforeunload` message to the review screen. Needs a bit of thought. This approach is a bit heavy handed and has plenty of caveats: can't change the message, need to jump through hoops to only trigger in certain contexts, browsers don't guarantee they will show it.


mattandrews:
Looks good, like the flag system 👍 

If you're not digging the bind/unbind approach, you could check the event target element in the `onbeforeunload` handler and see if it has a class/attribute indicating it shouldn't trigger the message? Then you could just bind the callback once?

PR #876, opened 2018-04-13
Swap out XSS library for DOMpurify

mattandrews:
We had a complaint that a user address for a material order came through like this:

	Address line 1: c&#x2F;o Foo Cottage

(was meant to be `c/o`).

It turned out that the `escape()` filter was adding this, not the `xss` library, but that package felt a little basic so thought I'd upgrade it here to `DOMpurify` which feels a bit better. 

I removed `escape()` and tested it by ordering with these fields:

`c/o Foo Cottage <script src="evil.js"></script><script>alert('lol');</script><marquee>foo</marquee>`

The `<script>`s were removed (though the `<marquee>` wasn't, but we don't display this input anywhre or store it), and the `c/o` remained the same – success.

### DIFF
--- a/controllers/about/ebulletin.js
+++ b/controllers/about/ebulletin.js
@@ -7,8 +7,7 @@ const rp = require('request-promise-native');
 const config = require('config');
 
 const { customEvent } = require('../../modules/analytics');
-const { purifyUserInput } = require('../../modules/validators');
-const { errorTranslator } = require('../../modules/validators');
+const { purifyUserInput, errorTranslator } = require('../../modules/validators');
 const { FORM_STATES } = require('../../modules/forms');
 const { getSecret } = require('../../modules/secrets');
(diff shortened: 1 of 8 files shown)


controllers/about/ebulletin.js
```diff
@@ -5,9 +5,9 @@ const { body, validationResult } = require('express-validator/check');
 const { matchedData } = require('express-validator/filter');
 const rp = require('request-promise-native');
 const config = require('config');
-const xss = require('xss');
 
 const { customEvent } = require('../../modules/analytics');
+const { purifyUserInput } = require('../../modules/validators');
```
davidrapson: This can be a one-liner? `const { purifyUserInput, errorTranslator } = require('../../modules/validators');`

mattandrews: Oops yes, wasn't paying attention! Fixed.

PR #877, opened 2018-04-13
Fix bilingual logo example

mattandrews:
On the [logo](https://www.biglotteryfund.org.uk/funding/funding-guidance/managing-your-funding/grant-acknowledgement-and-logos) page, if you choose Wales as your location you're prompted to download a bilingual logo. If you click "How to use our logo", though, you'll see a monolingual example image.

This PR fixes it so you'll only see a monolingual logo if you're browsing the site in English and have chosen "England, Scotland or Northern Ireland" as your location.


davidrapson:
👍 Nice, much simpler

PR #878, opened 2018-04-13
Reduce reliance on setViewGlobal

davidrapson:
This is an attempt to remove the occasional race condition we see with values set using `setViewGlobal` in middleware. Switches `locale` and `highContrast` to use request locals instead.


davidrapson:
I'll need to confirm once this is live, but at least on the test site I can no longer replicate the race condition.

PR #880, opened 2018-04-17
Split RC emails continued

davidrapson:
Please forgive the PR hijacking but had a stab at trying to simplify some of the promises. In the process realised we could move a fair amount of logic down into the mailer service exposed as a new `generateAndSend` method. 💭 welcome.

### DIFF
--- a/controllers/apply/reaching-communities-form.js
+++ b/controllers/apply/reaching-communities-form.js
@@ -1,11 +1,10 @@
 const { get, isArray } = require('lodash');
 const { check } = require('express-validator/check');
 
-const { createFormModel } = require('./create-form-model');
-const { HUB_EMAILS } = require('../../../modules/secrets');
-const app = require('../../../server');
-const appData = require('../../../modules/appData');
-const mail = require('../../../modules/mail');
+const { createFormModel } = require('./forms/create-form-model');
+const { HUB_EMAILS } = require('../../modules/secrets');
+const appData = require('../../modules/appData');
+const mail = require('../../modules/mail');
 
 const DEFAULT_EMAIL = HUB_EMAILS.england;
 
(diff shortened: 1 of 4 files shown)


controllers/apply/reaching-communities-form.js
```diff
@@ -242,94 +241,56 @@ formModel.registerSuccessStep({
...
-            };
+        /**
+         * Determine which internal address to send to:
+         * - If in test then send to primaryAddress
+         * - If multi-region, send to defailt/england-wide inbox
+         * - Otherwise send to the matching inbox for the selected region
+         */
+        const internalAddress = (function() {
+            if (appData.isNotProduction) {
+                return primaryAddress;
```
davidrapson: Ended up sending to the primary address, rather than needing an env var.

modules/mail.js
```diff
@@ -1,55 +1,97 @@
...
+        );
+    });
+}
+
+/**
+ * genarateHtmlEmail
+ *
+ * Given an email schema generate full email HTML
+ * - Render template through express/template-engine
+ * - Generate full emai HTML and return alongside orignal schema
```
mattandrews: Few typos in here...

modules/mail.js
```diff
@@ -1,55 +1,97 @@
...
+ * ]
+ */
+function genarateHtmlEmail(emailData) {
+    return new Promise((resolve, reject) => {
+        app.render(emailData.templateName, emailData.templateData, (templateErr, html) => {
+            if (templateErr) {
+                reject(templateErr);
+            }
+
+            renderHtmlEmail(html)
```
mattandrews: I know I named this method, but could we call it something a bit clearer like `makeStylesInline` etc? At the mo it's not very obvious what the difference is between `generateHtmlEmail` and `renderHtmlEmail`.

davidrapson: Good point

davidrapson:
Thanks for the notes. Ended up wrapping a few of the functions in `mail` using `util.promisify`, cleans up lots of the nesting.

PR #883, opened 2018-04-19
Refactor legacy proxy code, always use canonical link

davidrapson:
This PR refactors the legacy proxy code to always use canonical link. Moves the `~/link.aspx` / `meta[identifier]` logic up into the main `proxyLegacyPage` function and cleans that up so we always attempt to redirect to the canonical path if it's different to the current path.

Should fix a lot of the issues with vanity URLs. 🤞

### DIFF
--- a/modules/assets.test.js
+++ b/modules/assets.test.js
@@ -5,7 +5,7 @@ const expect = chai.expect;
 
 const { getCachebustedPath, getCachebustedRealPath, getImagePath } = require('./assets');
 
-describe.only('assets', () => {
+describe('assets', () => {
     it('should get cachebusted url path for an asset', () => {
         const result = getCachebustedPath('stylesheets/style.css');
         expect(result).to.match(/^\/assets\/build\/\w+\/stylesheets\/style.css$/);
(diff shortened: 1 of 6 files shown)


modules/assets.test.js
```diff
@@ -5,7 +5,7 @@ const expect = chai.expect;
 
 const { getCachebustedPath, getCachebustedRealPath, getImagePath } = require('./assets');
 
-describe.only('assets', () => {
```
davidrapson: I should stop using `.only` to debug tests…who knows how long this has been the only server test running?? 😓

modules/legacy.js
```diff
@@ -24,112 +72,51 @@ const proxyLegacyPage = ({ req, res, domModifications, followRedirect = true })
...
+            // (currently it's rewritten above to the external one)
+            const form = dom.window.document.getElementById('form1');
+            if (form) {
+                form.setAttribute('action', req.path);
+            }
 
-        /**
-         * Are we in an A/B test and do we have a Google Experiments ID
-         */
-        const experimentId = get(res.locals, 'ab.id');
```
davidrapson: I removed this whole analytics injection block on the basis that we don't have any A/B tests against legacy pages, and we're unlikely to again. AFAIK this was mostly for the homepage rollout right?

mattandrews: Yeah you're right, we shouldn't need to do it again either.

PR #894, opened 2018-04-24
Update timings middleware to split by type of page

davidrapson:
image


middleware/timings.js
```diff
@@ -13,43 +13,24 @@ module.exports = responseTime(function(req, res, time) {
...
-                ],
-                Namespace: 'SITE/TRAFFIC'
-            })
-            .send();
-    } else if (req.method === 'POST') {
-        cloudwatch
-            .putMetricData({
-                MetricData: [
+    const method = req.method.toUpperCase();
+    const isLegacy = res.getHeader('X-BLF-Legacy') === 'true';
```
davidrapson: This callback gets triggered `onHeadersSent` so we can reliably check for the `X-BLF-Legacy` header even though this middleware is mounted earlier.

PR #896, opened 2018-04-25
Improve mail sending from materials form

davidrapson:
A few small tweaks to the materials form:

- Send both emails to "customer" in dev/test environments.
- Split add/remove and main form routes into functions
- Refactor `makeOrderText` function to remove unused return value and to locate all order text in one block.


davidrapson:
For the sake of that one `TODO` I might as well to…do it now.

davidrapson:
@mattandrews Mind giving the last two commits a look?

PR #904, opened 2018-04-26
Blog pages

davidrapson:
image

image

image


views/components/nav.njk
```diff
@@ -64,6 +64,19 @@
     </nav>
 {% endmacro %}
 
+{% macro selectionTrail(trail) %}
+    {% if trail | length > 0 %}
+    <ol class="selection-trail">
+        {% for crumb in trail %}
```
mattandrews: Indentation of HTML here is a bit off (and the `{% if %}` above)

PR #910, opened 2018-04-30
Apply route

davidrapson:
Does a little work to clean up how the apply route is mounted:

- Model `/apply/` in `routes.js` using the same method as all other pages.
- Move the startpage into `createFormRouter`
- Relocate the form helpers into a standalone directory


controllers/routes.js
```diff
@@ -38,6 +38,12 @@ const sections = {
         langTitlePath: 'global.nav.blog',
         controllerPath: path.resolve(__dirname, './blog'),
         showInNavigation: false
+    }),
+    apply: createSection({
+        path: '/apply',
+        langTitlePath: 'global.nav.blog',
```
mattandrews: Is this right?

PR #911, opened 2018-04-30
Refactor view filters to add tests

davidrapson:
🤓


modules/filters.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+/**
+ * Any functions added here will be added as
+ * custom Nunkucks filters
```
mattandrews: Nunkucks eh.

PR #913, opened 2018-04-30
View globals prune

davidrapson:
More pruning of view globals in favour of locals. Main changes are:

- Move a few properties to the global `app.locals` block
- Replace inline makeAbsoluteImage method with identical getAbsoluteUrl
- Rework getCurrentUrl to not need request in template

### DIFF
--- a/views/layouts/profiles.njk
+++ b/views/layouts/profiles.njk
@@ -31,7 +31,7 @@
             </div>
             <div class="page-section__supplementary accent--{{ pageAccent }} a--border-top">
                 <aside class="content-box">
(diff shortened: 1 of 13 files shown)


views/layouts/profiles.njk
```diff
@@ -31,7 +31,7 @@
             </div>
             <div class="page-section__supplementary accent--{{ pageAccent }} a--border-top">
                 <aside class="content-box">
-                    {{ nestedMenu(navigation, getCurrentUrl(request)) }}
+                    {{ nestedMenu(navigation, getCurrentUrl()) }}
```
mattandrews: 💯

PR #929, opened 2018-05-04
Remove session cookie from default cloudfront config

davidrapson:
Building on from https://github.com/biglotteryfund/blf-alpha/pull/928 this PR removes the session cookie from the default cloudfront config.

In place it adds a new `sessionRoute` route-type which allows us to whitelist certain routes for use with session cookies. 

The one part that needs some extra handling is the `/item/:id` route in the materials form as there is currently not a wildcard route to cover it. Needs a bit of adjustment to work.

💭Thoughts very welcome.

### DIFF
--- a/config/cloudfront/live.json
+++ b/config/cloudfront/live.json
@@ -603,6 +603,71 @@
         "AllowedMethods": {
           "Items": [
             "HEAD",
+            "DELETE",
+            "POST",
+            "GET",
+            "OPTIONS",
+            "PUT",
+            "PATCH"
+          ],
+          "Quantity": 7,
+          "CachedMethods": {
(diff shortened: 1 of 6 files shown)


mattandrews:
So.. there's no reason the `/item/:id` route needs to be structured that way – it could just take an `itemId` POST parameter instead, and just live under `/modify-item` or something. Would that help?

davidrapson:
Ah I mostly just meant that there was no wildcard route for cloudfront to cover it, added one in `https://github.com/biglotteryfund/blf-alpha/pull/929/commits/f098bb9add6a0e46f074a82ff0da7e9fffbc151c`

PR #930, opened 2018-05-04
AFA portal downtime page

mattandrews:
This is an attempt at formalising this a bit more as it's becoming more common for scheduled downtime/maintenance to happen on the AFA portal.

IT can point `apply.biglotteryfund.org.uk` at AWS (I've already added the domain to Cloudfront), so when they do, this PR will intercept any request on that hostname and serve a downtime page with a customisable date (if known) for the end of the maintenance period. 

Once IT change the CNAME this page will stop serving.


middleware/portal.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+const moment = require('moment');
+
+/**
+ * Serves a holding/downtime page for the AFA portal if redirected here
+ */
+module.exports = (req, res, next) => {
+    const downtimeEndDate = moment('2018-05-29');
```
mattandrews: This will change (there's an upcoming maintenance period in the next week or two but I'm trying to get ahead of it), but illustrates the premise.

PR #931, opened 2018-05-04
Centralise common date formats

davidrapson:
Prompted by my own comment on https://github.com/biglotteryfund/blf-alpha/pull/930

There's clearly a need for some kind of format sharing as I'd already managed to use different date formats for the blog!

Not sure if this the right _place_ for this stuff to live, does this feel OK? Too verbose?

### DIFF
--- a/middleware/locals.js
+++ b/middleware/locals.js
@@ -69,8 +69,9 @@ module.exports = {
          * @see https://momentjs.com/docs/#/displaying/format/
          */
         res.locals.formatDate = function(dateString, format) {
-            moment.locale(locale);
-            return moment(dateString).format(format);
+            return moment(dateString)
+                .locale(locale)
+                .format(format);
         };
 
         /**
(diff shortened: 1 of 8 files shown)


middleware/locals.js
```diff
@@ -73,6 +73,15 @@ module.exports = {
             return moment(dateString).format(format);
         };
 
+        /**
+         * View helper to represent date as relative time
+         * @param {String} dateString
+         */
+        res.locals.timeFromNow = function(dateString) {
+            moment.locale(locale);
```
mattandrews: One thing I noticed when using this on the other PR – this function sets global locale, so possibly each time it's called here we change the locale of any later calls to `moment()`. You can chain it with the line below like `moment(dateString).locale(locale).fromNow()`.

server.js
```diff
@@ -103,6 +103,15 @@ function initAppLocals() {
      */
     app.locals.metadata = config.get('meta');
 
+    /**
+     * Common date formats
+     */
+    app.locals.DATE_FORMATS = {
+        short: 'D MMMM, YYYY',
```
mattandrews: The actual formats feel like they should live in a config file? `default.json`?

middleware/locals.js
davidrapson: Oh, good call.

PR #936, opened 2018-05-08
Autocomplete attributes

davidrapson:
Improvements to fix these warnings in Chrome:

screen shot 2018-05-08 at 11 16 18


davidrapson:
> Worth doing for the free materials form, too?

Yes, but I'd like to migrate those form fields to use the same helpers first

PR #943, opened 2018-05-08
Port remaining integration tests to cypress

davidrapson:
This PR ports the two remaining integration tests over to Cypress (users and free materials). Switches tests to be mostly UI driven rather than making discreet network requests.

Commented on the one mildly contentions change.


controllers/funding/materials.js
```diff
@@ -394,58 +394,50 @@ function initForm({ router, routeConfig }) {
             const errors = validationResult(req);
 
             if (errors.isEmpty()) {
-                /**
-                 * Allow tests to run without sending email
-                 * this is only used in tests, so we confirm the form data was correct
-                 */
-                if (req.body.skipEmail) {
```
davidrapson: Removing this check as a) the `mail` function already skips sending email when in test mode centrally and b) it feels better to test against the real user flow, even at the expense of asserting against the mail response.

mattandrews: My only concern is that last year I introduced a bug where the material email sent out the names of the fields rather than their values (eg. `First name: yourName`) – I added the tests to at least verify the data being sent out was what we expected. Not sure if there's a way we could handle this here?

controllers/funding/materials-helpers.test.js
```diff
@@ -3,9 +3,54 @@
...
+            yourPostcode: 'EC4A 1DE',
+            yourProjectName: '',
+            yourReason: 'projectOpening',
+            yourReasonOther: '',
+            yourGrantAmount: 'over10k',
+            yourGrantAmountOther: ''
+        };
+
+        const orderText = makeOrderText(items, details);
+        console.log(orderText);
```
mattandrews: 🥇

PR #946, opened 2018-05-09
Add the commit hash to Sentry to group release errors together

mattandrews:
From the Sentry docs:

> This will annotate each event with the version of your application, as well as automatically create a release entity in the system the first time it's seen.


davidrapson:
image

👍 Great minds think alike

PR #950, opened 2018-05-09
Allow aliases to be fetched from the CMS

davidrapson:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/68

Allows aliases to fetched from the CMS:

- Add a fallback middleware to lookup aliases from the CMS (tries `en` first and then `cy` to avoid always making two requests)
- Migrates a batch of vanity URLs to the CMS
- Flattens remaining vanity URLs into a JSON config, goal is to move all redirects that remain in the app into this simpler format (h/t @mattandrews for the suggestion)


davidrapson:
Note: This will fail until https://github.com/biglotteryfund/craft-dev/pull/68 is merged and aliases made live

PR #951, opened 2018-05-09
Aliases from cms

davidrapson:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/68

Allows aliases to fetched from the CMS:

- Add a fallback middleware to lookup aliases from the CMS (tries `en` first and then `cy` to avoid always making two requests)
- Migrates a batch of vanity URLs to the CMS
- Flattens remaining vanity URLs into a JSON config, goal is to move all redirects that remain in the app into this simpler format (h/t @mattandrews for the suggestion)


config/app/vanity-redirects.json
```diff
@@ -0,0 +1,35 @@
+{
+  "/about-big/customer-service/fraud": "/contact#fraud",
+  "/about-big/customer-service/making-a-complaint": "/contact#complaints",
```
mattandrews: One downside of this being a JSON file (not JS) means we lose the `${anchors.contactComplaints}` config item here (eg. there's a hard dependency between the anchor string used here and the ID in the Contact template). Worth moving this file to JS to allow that syntax? JSON means we can't even leave a comment here to explain that that string is significant. On the other hand we could add that comment in the Contact page template? Dunno.

middleware/redirects.js
```diff
@@ -60,11 +62,43 @@ function redirectNoWelsh(req, res, next) {
...
+/**
+ * Vanity URL lookup
+ * - First look up global/en alias from the CMS
+ * - If that fails check for a welsh specific alias
+ * - Call next() and passthrough on any failures
+ */
+async function vanityLookup(req, res, next) {
+    const findAlias = find(alias => alias.from === req.path);
+    try {
+        const enAliases = await contentApi.getAliases({ locale: 'en' });
```
mattandrews: 💯 for use of `await`, I'm definitely guilty of littering code with Promise callbacks.

config/app/vanity-redirects.json
davidrapson: I'm OK with loosing the `${anchors.contactComplaints}` config as we're about to move the contact page into the CMS anyway so becomes moot at that point.

Good point about the comments though, will move it back into JS, but keep the simpler k/v format.

middleware/redirects.js
davidrapson: Eh…wouldn't go that far. This is cascade-y enough that the async/await code isn't much more concise.

PR #962, opened 2018-05-11
Strawman: Add inline feedback to form success pages

davidrapson:
Strawman PR for discussion but it might be interesting to put the inline feedback form on success pages given we've seen a reasonable take up for it.

screen shot 2018-05-11 at 16 02 42


davidrapson:
Added a flag to this to only show it on test, so merging in so we can try this out.

PR #969, opened 2018-05-14
Fixup proxied manage funding

davidrapson:
https://sentry.io/big-lottery-fund/biglotteryfund/issues/555790658/

Fixes links under `/manage-your-funding` which aren't handled by the CMS. Notably this section `/funding/funding-guidance/managing-your-funding/about-equalities/evidence-collection-tools`.

As we pass anything under `/funding-guidance` to the CMS we need to guard against missing content for routes that are not found in the CMS. By passing to next we can make sure to pass down to later route handlers, notably `proxyLegacy`.

### DIFF
--- a/controllers/common.js
+++ b/controllers/common.js
@@ -46,32 +46,35 @@ function handleStaticPage(page) {
 }
 
 function handleCmsPage(page) {
-    return (req, res) => {
+    return (req, res, next) => {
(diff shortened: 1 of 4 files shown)


davidrapson:
> Nice catch 👍

Thank Sentry

PR #970, opened 2018-05-14
Move free materials to the CMS

mattandrews:
Quite a lot going on here – this code was ripe for refactoring so hopefully it's improved somewhat now.

This change means we fetch product data from the CMS, and it simplifies how we store data in sessions (mainly: migrating the data structure to be an array of items instead of an object, which simplifies some of the client-side Vue logic as well as the server-side logic).

It also attempts to ease (some of) the pain of the progressively-enhanced frontend, including fixing a long-term bug where reloading the page and reducing an item's quantity to 0 would leave the item visually selected. Also removes some ancient weirdness where we pass around the server-side session quantity as a fallback to the client-side one. 

Finally it updates the internal stats tool to get its product data from the API (including all products).


controllers/funding/materials-helpers.js
```diff
@@ -239,8 +240,31 @@ If you have feedback, please contact matt.andrews@biglotteryfund.org.uk.`;
...
+async function injectMerchandise(req, res, next) {
+    try {
+        res.locals.availableItems = await contentApi.getMerchandise(req.i18n.getLocale());
+        next();
+    } catch (error) {
+        next(error);
+    }
+}
+
+function injectMerchandiseCustom({ locale = false, showAll = false }) {
```
davidrapson: Can `injectMerchandiseCustom` and `injectMerchandise` be merged into a single method?

controllers/funding/materials.js
```diff
@@ -10,60 +10,78 @@ const Raven = require('raven');
...
-            } else {
-                // Alert the user that they're blocked from adding this item
-                set(req.session, [orderKey, 'itemBlocked'], true);
-            }
-        } else if (currentItemQuantity > 1 && action === 'decrease') {
-            set(req.session, [orderKey, code, 'id'], id);
-            set(req.session, [orderKey, code, 'quantity'], currentItemQuantity - 1);
-        } else if (action === 'remove' || (action === 'decrease' && currentItemQuantity === 1)) {
-            set(req.session, [orderKey, code, 'quantity'], 0);
+        if (action === 'increase' && (hasBlockerItem || noSpaceLeft)) {
```
davidrapson: `!` conditions always read a little weirdly to me and the `action === 'increase' && (hasBlockerItem || noSpaceLeft)` check might need a name. Does this feel any simpler:

```js
const shouldBlock = action === 'increase' && (hasBlockerItem || noSpaceLeft);
if (existingProduct) {
    …
} else if (shouldBlock) {
    …
} else {
    …
}
```

controllers/funding/materials.js
```diff
@@ -142,10 +157,10 @@ function initForm({ router, routeConfig }) {
...
         const content = res.locals.content;
         const lang = req.i18n.__(routeConfig.lang);
-        const orders = get(req.session, materialsOrderKey, {});
-        const numOrders = sumBy(values(orders), order => {
-            return order.quantity;
-        });
+        const availableItems = res.locals.availableItems;
+        const orders = req.session[sessionOrderKey] || [];
+
+        const formActionBase = '/' + res.locals.sectionId + routeConfig.path;
```
davidrapson: IIRC this can be written using only the request as `const formActionBase = req.baseUrl + req.path`;

davidrapson: Silly follow up question, but do we need to construct this? Are we in the right context for the form to post to itself using an empty `action=''`?

controllers/funding/materials.js
```diff
@@ -191,7 +217,9 @@ function initForm({ router, routeConfig }) {
                                 sendTo: customerSendTo,
                                 subject: 'Thank you for your Big Lottery Fund order',
                                 templateName: 'emails/newMaterialOrder',
-                                templateData: {}
+                                templateData: {
+                                    locale: req.i18n.getLocale()
```
davidrapson: Maybe add a TODO here to try and work out why we need this?

controllers/funding/materials-helpers.js
mattandrews: Just did this!

controllers/funding/materials.js
mattandrews: The `shouldBlock` name works, but the logic is in this sequence for a reason – the `if (existingProduct)` check happens second because by that point we've ruled out that it's not in invalid increase (eg. they're at maximum, or blocked). If we put that check first we'd have to have nested/repeated checks for quantity/blocker status again.

controllers/funding/materials.js
mattandrews: Done

PR #974, opened 2018-05-15
Config prune

davidrapson:
Prunes a bit of config cruft. 🗑


config/app/userlevels.json
```diff
@@ -1,18 +0,0 @@
-[
```
mattandrews: Thinking out loud here: Perhaps eventually we should switch this whole logic to use Craft to log in / manage auth. The login system used for the tools etc is technically part of a potential future user registration system, which shouldn't really be tied up with internal/staff accounts (which should be removed after they leave).

davidrapson: Yep, totally agree with this. Decoupling public auth from staff tools would be a goood idea.

PR #979, opened 2018-05-16
Migrate contact page to the CMS

davidrapson:
Mostly deletions here, with one extra check to not output primary content if we don't have it.


davidrapson:
> I can't remember if the CMS template outputs IDs for these sections but we should try to preserve this behaviour if possible before we merge this.

Yep 👍 We output generated IDs so can update them to use that format.

davidrapson:
I'm up for an explicit `anchor` or `slug` field, I think we can come up with some wording like "A short ID to use if you'd like to anchor to a section, e.g. #my-section`. Will merge this in, but happy to add that to the CMS as a follow up bit of work.

PR #980, opened 2018-05-16
Remove old images

davidrapson:
Fairly sure these can go now:

- Under/over 10k images were case studies that are powered by the CMS
- Materials products also now live in the CMS


davidrapson:
image

Lame.

PR #985, opened 2018-05-16
Model about page in the CMS

davidrapson:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/72

Models the about page in the CMS. Has a couple of key trade-offs:

- Uses the same _content model_ as information pages, but has a custom template. As a result thumbnails added to segments on the about page are ignored. 
- ~Relies on a new `pullQuote` content type. This has been modelled in the shared `contentSegments` matrix field but we don't currently support pull quotes on general information pages; this might get confusing.~ Added support for them on information pages…they look OK! Still something to consider.

### DIFF
--- a/assets/sass/components/media.scss
+++ b/assets/sass/components/media.scss
@@ -61,10 +61,6 @@
         position: relative;
         top: -$spacingUnit; // Optical alignment
     }
-}
-.media-aside__text {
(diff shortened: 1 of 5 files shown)


davidrapson:
Closing this as I'm going to try a slightly different direction for this.

PR #996, opened 2018-05-17
Remove remaining awards for all A/B test code

davidrapson:
The awards for all scotland team were meeting last week to decide if they are happy to switch to 100% of traffic. I need to follow up to confirm but raising this PR now to remove the code once they are ready.

_Note: The copy will need to be updated in the CMS when this goes live._


davidrapson:
This has been approved to go out from 1st June. Will mostly likely need to be on Monday 4th June.

PR #998, opened 2018-05-17
Update blog posts to use new flexible content

davidrapson:
Extends upon https://github.com/biglotteryfund/blf-alpha/pull/997 as a way of confirming that the new `flexibleContent` fields are, well, flexible.

Updates the blog post template to use a flexible content field and adds a content type of `inlineFigure` for adding a photo with a caption as a dedicated content block.

### DIFF
--- a/assets/sass/scaffolding/scope-prose.scss
+++ b/assets/sass/scaffolding/scope-prose.scss
@@ -258,12 +258,7 @@ $constrainedWide: 42em;
     }
 
     figure {
-        margin-left: -($spacingUnit);
-        margin-right: -($spacingUnit);
-
-        @include mq(tablet) {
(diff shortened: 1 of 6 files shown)


assets/sass/scaffolding/scope-prose.scss
```diff
@@ -265,15 +265,14 @@ $constrainedWide: 42em;
             margin: 1em 0 2em;
         }
 
-        &.auto-width {
-            img {
-                width: auto;
-            }
+        img {
```
davidrapson: Need to test this a bit more, but I _think_ we're OK to make all images/figures in content areas have an auto width rather than relying on the `auto-width` class.

PR #1002, opened 2018-05-17
Remove (most) overrides from ebulletin form

davidrapson:
Makes the e-bulletin form fields look like the rest of our form fields. No more placeholders-as-labels!

Makes some concessions to the fact that the form is on a tinted background and in a constrained space. There is probably a case to turn this into a more general modifier or two but for now I'm just happy to make this form less special.

screen shot 2018-05-17 at 15 50 46

image

### DIFF
--- a/assets/sass/components/forms.scss
+++ b/assets/sass/components/forms.scss
@@ -232,44 +232,25 @@
 /* =========================================================================
    Custom Form: Ebulletin
    ========================================================================= */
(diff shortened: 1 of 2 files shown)


mattandrews:
FYI this build is failing with the same test that was failing for me (and I couldn't fix). Looks like a possible timeout/race condition – maybe we should mock the API data for this test?

PR #1003, opened 2018-05-18
[Experimental] Refactor materials form to use newer form model

davidrapson:
This is a bit experimental at the moment. But this is a working version of the materials form using the newer form fields.

image

Things left to do if we go ahead with this:

- [ ] Confirm using RC-style form design works
- [ ] Refine translation logic, add translation code for any new labels
- [ ] Fix a bug where "other" field is not toggled when reloading the form with a value
- [ ] Test, test, test


controllers/materials/index.js
```diff
@@ -239,17 +206,17 @@ function initForm({ router, routeConfig }) {
...
                     })
                     .catch(err => {
                         Raven.captureException(err);
-                        renderForm(req, res, FORM_STATES.SUBMISSION_ERROR);
+                        renderForm({ req, res, status: FORM_STATES.SUBMISSION_ERROR });
                     });
             } else {
-                req.flash('formErrors', errors.array());
-                req.flash('formValues', req.body);
+                req.session['materials.formData'] = matchedData(req, { locations: ['body'] });
```
mattandrews: Can't remember for certain but I think this will only return fields that were required?

davidrapson: Seems to work as I'd expect it to here, I'm only using it to repopulate form fields, same way as `create-form-router`. Will double check though.

PR #1036, opened 2018-06-08
Survey prompt

davidrapson:
This PR makes some changes to the micro surveys and introduces a new prompt style for one-off messages.

- Reworks the original micro-surveys to be displayed inline on the page.
- Re-appropriates the pop-over message style for a new weighted prompt message e.g. show to 20% of people.

Currently weighting of the prompt message is fairly simple, and only supports a single active message for now. Mostly here so we can ask people to complete an external survey.

screen shot 2018-06-08 at 15 35 54

Also removes the hacky iOS 11 workaround as a) the "did you find?" survey is no longer position fixed and b) the bug is fixed in iOS 11.3 anyway. Hoorah.


controllers/toplevel/index.js
```diff
@@ -88,13 +107,17 @@ module.exports = ({ router, pages }) => {
 
                 res.send({
                     status: surveyToShow ? 'success' : 'error',
+                    lang: {
```
davidrapson: Including any global translations here means we can make the whole survey component client-side. Avoids the half-vue / half-nunjucks interpolation issue.

PR #1044, opened 2018-06-12
Rework survey results

davidrapson:
Now that we show the "did you find?" survey as a fixed bar on all pages the concept of having multiple surveys with different wording no longer matches with what we are actually asking people; which is a single question across the site.

This PR adds a new stripped down data model for survey answers along with a new tools screen with a single site-wide graph. In order to maintain the ability to have per page graphs you can pass a `?path` query parameter to see results for a single page.

Keeping the old screen and data around for now but there is a task to write a migration script to clean things up afterwards.

It does mean removing the need for the CMS endpoint for managing the surveys (sorry matt 😢) but it doesn't feel like we need the complexity in practice.

### DIFF
--- a/assets/js/vue-components/surveys.js
+++ b/assets/js/vue-components/surveys.js
@@ -1,14 +1,20 @@
 import $ from 'jquery';
 import Vue from 'vue';
 
+const statuses = {
+    NOT_ASKED: 'NOT_ASKED',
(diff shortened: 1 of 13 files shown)


mattandrews:
Think this makes sense – would it be worth adding some navigation to the page which lists each unique page path along with a link to `?path=$path` for ease of use?

Also, this looks a bit broken...

image

PR #1064, opened 2018-06-19
Update fields based on initial feedback

davidrapson:
Update fields based on initial feedback and separates out UI/interaction explanation text from guidance help text to make sure questions make more sense when shown on the preview screen.

### DIFF
--- a/controllers/apply/building-connections/fields.js
+++ b/controllers/apply/building-connections/fields.js
@@ -1,35 +1,23 @@
 'use strict';
 const { check } = require('express-validator/check');
 
-// @TODO: Are these the same as Reaching Communities? Can we share then
-const PROJECT_LOCATIONS = [
+const currentWork = [
     {
-        label: 'North East & Cumbria',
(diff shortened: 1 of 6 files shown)


controllers/apply/building-connections/fields.js
```diff
@@ -49,54 +37,77 @@ const yourIdea = [
...
+                    'North East',
+                    'North West',
+                    'Yorkshire and the Humber',
+                    'East Midlands',
+                    'West Midlands',
+                    'East of England',
+                    'Greater London',
+                    'South East',
+                    'South West'
+                ].map(_ => ({ label: _, value: _ })),
```
davidrapson: This `.map` feels a little silly, might see if I can handle this logic in the form model, i.e. allow either an array (where labels + values are the same) or an object. Maybe overthinking this.

mattandrews: Works for me really.

Issue #1021, opened 2018-05-30
Cached Webpack bundles are served on the wrong domain


mattandrews:
![image](https://user-images.githubusercontent.com/394376/40716488-19065ea6-6401-11e8-8075-c3ca9a1d9883.png)

After a deploy, a stale page (eg. https://www.biglotteryfund.org.uk/wales which has a long `s-max-age`) will continue to refer to static files hosted on `www.` instead of `media.` (eg. bundle files like [this](https://www.biglotteryfund.org.uk/assets/build/7851ef2665f3/javascripts/vendors~vue-components.bundle.js) which are deleted when a new app instance is deployed).

These Webpack bundles should be served via `media.biglotteryfund.org.uk` which is hosted in S3 and will always be valid.

I don't fully understand how/why this bundling works so not sure where to begin configuring how they're served – thoughts very welcome @davidrapson.

PR #1067, opened 2018-06-20
Dynamically set webpack public path

davidrapson:
This is supposedly the documented way to avoid https://github.com/biglotteryfund/blf-alpha/issues/1021, either that or set the CDN domain statically at build time. Either way feels a bit…hacky so going to take a look at alternatives. Raising this PR for reference.

See https://webpack.js.org/guides/public-path/


davidrapson:
Feeling slightly warmer to this approach after reading https://medium.com/@aviv.rosental/portable-bundle-with-webpack-d2eed216cd4c

Think it echoes my concerns with the statically defined approach.

PR #1070, opened 2018-06-20
Store application responses

davidrapson:
First stab at storing application submissions. Had to rejig where the reference ID was defined as it was being created too early, so not unique to the submission. Involves a little session wrangling but is now unique.

Thoughts very welcome 💭

### DIFF
--- a/controllers/helpers/create-form-router.js
+++ b/controllers/helpers/create-form-router.js
@@ -1,6 +1,5 @@
 'use strict';
 const Raven = require('raven');
-const shortid = require('shortid');
 const { get, isEmpty, set, unset } = require('lodash');
 const { validationResult } = require('express-validator/check');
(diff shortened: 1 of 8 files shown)


controllers/helpers/create-form-router.js
```diff
@@ -134,27 +134,25 @@ function createFormRouter({ router, formModel }) {
...
-                        res.redirect(`${req.baseUrl}/success`);
-                    })
-                    .catch(err => {
-                        Raven.captureException(err);
-                        res.render('pages/apply/error', {
-                            form: formModel,
-                            stepConfig: errorStep,
-                            returnUrl: `${req.baseUrl}/review`
-                        });
+                try {
```
davidrapson: Didn't actually change anything here in the end other than switch to async/await.

PR #1079, opened 2018-06-21
Use hash for reference code

davidrapson:
The reference codes we generate are currently a little awkward to read out, plus I feel a little gross about essentially having two IDs in the database.

What we are really doing is looking for a human readable ID that isn't just the raw database ID, a bit like a youtube video reference. Turns out https://hashids.org/ does that.


davidrapson:
@mattandrews Had second thoughts about this and ended up using a hash of the submission as you suggested and storing explicitly in the database again.

services/applications.js
```diff
@@ -1,14 +1,17 @@
 'use strict';
-const Hashids = require('hashids');
+var hash = require('object-hash');
```
mattandrews: Should be `const`?

PR #1080, opened 2018-06-22
Convert cookie warning to vue component

davidrapson:
Converted the cookie consent banner to a vue component and moved text to translation files.


assets/js/vue-apps/components/cookie-consent.vue
```diff
@@ -0,0 +1,37 @@
...
+import { storageAvailable } from '../../helpers/storage';
+const { trackEvent } = require('../../helpers/metrics');
+
+const canStore = storageAvailable('localStorage');
+const STORAGE_KEY = 'biglotteryfund:cookie-consent';
+
+export default {
+    props: ['lang'],
+    data() {
+        const hasAccepted = canStore && window.localStorage.getItem(STORAGE_KEY) === 'true';
```
mattandrews: I was thinking about `localStorage` yesterday – in some places we don't test for `canStore` (we just use `try`/`catch` blocks), and in others like this we test for it. 

We have a `isModernBrowser` test (and corresponding class) but don't use it for anything. Do you think we should change anything here (eg. do we need to test for `localStorage` support, or should we just use try/catch etc)?

davidrapson:
Good shout about local storage, will come back to it in another PR.

PR #1095, opened 2018-06-29
Add a dashboard for viewing online application data

mattandrews:
This is a work in progress but adds a basic way to view application data. Open for discussion on how to approach this!

### DIFF
--- a/controllers/apply/building-connections/processor.js
+++ b/controllers/apply/building-connections/processor.js
@@ -26,7 +26,7 @@ module.exports = async function processor(formModel, formData) {
     const stepsWithValues = formModel.getStepsWithValues(formData);
 
     const dataToStore = formatDataForStorage(stepsWithValues);
-    const record = await applicationService.storeApplication(formModel.shortCode, dataToStore);
+    const record = await applicationService.storeApplication(formModel.id, formModel.shortCode, dataToStore);
 
     const primaryAddress = flatData['email'];
     // @TODO determine an internal email address to send to for production environments
--- a/controllers/tools/applications.js
+++ b/controllers/tools/applications.js
@@ -1,0 +1,47 @@
+'use strict';
+const applicationService = require('../../services/applications');
+const { purifyUserInput } = require('../../modules/validators');
+
+function init({ router }) {
+    router.route('/applications/:formId/:applicationId?').get(async (req, res, next) => {
+        try {
+            const formId = purifyUserInput(req.params.formId);
+
+            const applications = await applicationService.getApplicationsByForm(formId);
+            if (!applications) {
+                return next();
+            }
+            // via https://stackoverflow.com/questions/196972/convert-string-to-title-case-with-javascript#comment85199999_46959528
+            const titleCase = (str) => str.replace(/\b\S/g, t => t.toUpperCase());
+            let formTitle = titleCase(applications[0].form_id.replace(/-/g, ' '));
+
+            let applicationData;
+            if (req.params.applicationId) {
+                applicationData = await applicationService.getApplicationsById(purifyUserInput(req.params.applicationId));
+                if (!applicationData) {
+                    return next();
+                }
+            }
+
+            res.render('tools/applications', {
+                formTitle,
+                applications,
+                formId,
+                applicationData
+            });
+        } catch (error) {
+            next(error);
+        }
+    });
+
+    return router;
+}
+
+module.exports = {
+    init
+};
+
+
+
+
+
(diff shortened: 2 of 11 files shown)


controllers/apply/building-connections/processor.js
```diff
@@ -26,7 +26,7 @@ module.exports = async function processor(formModel, formData) {
     const stepsWithValues = formModel.getStepsWithValues(formData);
 
     const dataToStore = formatDataForStorage(stepsWithValues);
-    const record = await applicationService.storeApplication(formModel.shortCode, dataToStore);
+    const record = await applicationService.storeApplication(formModel.id, formModel.shortCode, dataToStore);
```
mattandrews: (I've pre-emptively added this field manually to the existing dev/prod databases so we can merge this without issues)

models/application.js
```diff
@@ -2,6 +2,10 @@
 
 module.exports = function(sequelize, DataTypes) {
     return sequelize.define('application', {
+        form_id: {
+            type: DataTypes.STRING,
+            allowNull: false
+        },
```
mattandrews: (I've pre-emptively added this field manually to the existing dev/prod databases so we can merge this without issues)

controllers/tools/index.js
```diff
@@ -7,12 +7,18 @@ const { toolsSecurityHeaders } = require('../../middleware/securityHeaders');
...
 const surveysService = require('../../services/surveys');
 
 const pagelistRouter = require('./pagelist');
 const seedRouter = require('./seed');
 
+const availableApplicationModels = [
+    require('../apply/building-connections/form-model'),
+    require('../apply/reaching-communities/form-model')
+];
+
```
mattandrews: I feel sliiiiiiiightly weird about this, but I couldn't think of a better way to retrieve the details of the form and validate that it really exists. Suggestions welcome!

davidrapson: Hmm yeah, what do we really need from the form model? Looks like just the title. Could we store that as an additional field so we're only querying against the applications table and don't need to worry too much about the original form model?

davidrapson:
What happened with the `package-lock` file?

controllers/tools/index.js
```diff
@@ -103,4 +109,30 @@ router.route('/order-stats').get(injectMerchandise({ locale: 'en', showAll: true
     }
 });
 
+router.route('/applications/:formId/:applicationId?').get(async (req, res, next) => {
```
davidrapson: I know not everything is broken in this controller, but maybe a candidate to live in a separate file?

views/components/applicationSummary.njk
```diff
@@ -0,0 +1,19 @@
+{% macro readOnlySummary(summary) %}
+    {% for step in summary %}
+        <h4 class="data__step">{{ step.name }}</h4>
```
davidrapson: These `data__` classes look like they are being used out of context. I can't see a parent `data` class anywhere? Do these need to be different classes?

views/tools/applications.njk
```diff
@@ -0,0 +1,43 @@
...
+            </div>
+        </div>
+    {% else %}
+        <h2>All applications ({{ applications.length }})</h2>
+        <p>Click an application below to view all its data.</p>
+        <div class="border mb-5">
+            <div class="list-group">
+                {% for application in applications %}
+                    <a class="list-group-item list-group-item-action flex-column align-items-start"
+                       href="/tools/applications/{{ form.id }}/{{ application.reference_id }}">
```
davidrapson: Are we able to avoid hard-coding the route here? Does a relative link work e.g. `./{{ form.id }}/{{ application.reference_id }}`

controllers/tools/applications.js
```diff
@@ -0,0 +1,47 @@
...
+function init({ router }) {
+    router.route('/applications/:formId/:applicationId?').get(async (req, res, next) => {
+        try {
+            const formId = purifyUserInput(req.params.formId);
+
+            const applications = await applicationService.getApplicationsByForm(formId);
+            if (!applications) {
+                return next();
+            }
+            // via https://stackoverflow.com/questions/196972/convert-string-to-title-case-with-javascript#comment85199999_46959528
```
davidrapson: Oh, I thought you might end up storing the title of the form in the database when saving an application. But this works too.

davidrapson:
One more thought: Do we want to put this page behind an IP whitelist or are we happy with just the current auth?

mattandrews:
Re the `package-lock.json` thing – no idea, ran `npm install` on my home laptop (not updated this app there in a while) and this was the result. Might see if I can clean it up on my work machine now...

PR #1096, opened 2018-06-29
Hero credit

davidrapson:
We've been given some stricter requirements for crediting photos used on funding programme pages (i.e. to credit the specific grant programme the image is from).

image

There is a companion PR to the CMS which adds a per-programme (rather than per-image) credit field for adding these longer credits.


views/pages/funding/programme-detail.njk
```diff
@@ -33,6 +33,9 @@
                         fields = entry.summary,
                         isCompact = true
                     ) }}
+                    {% if entry.heroCredit %}
+                        <p class="u-credit u-no-margin"><strong>Image credit:</strong> {{ entry.heroCredit }}</p>
```
mattandrews: Can we get "Image credit" translated?

davidrapson: Yep, will sort that now

PR #1099, opened 2018-07-03
Dashboard tweaks

mattandrews:
- Adds an endpoint for `/tools/applications` (just for completeness' sake – lists all available forms)
- Adds a download/Excel export button to application pages
- Adds an instance method to the model for converting form slugs to titles


controllers/tools/applications.js
```diff
@@ -1,22 +1,63 @@
...
+                    return next();
+                }
+
+                formTitle = applications[0].formTitle;
+
+                if (req.query.download) {
+
+                    // build a link to the application page
+                    let urlBase = getFullUrl(req);
+                    urlBase = urlBase.split('?')[0]; // remove querystring
```
davidrapson: Minor thing but you can maybe use the bultin `url.parse` to get the part of the URL you need. Don't worry about it for this PR

PR #1100, opened 2018-07-03
[Experiment] Colocate apply app

davidrapson:
This PR is mostly an experiment to get a feel for if this is a better way to structure the apply section.

Realised that nunjucks can take an array of paths to look for views in, not just a single path. If we include the root in the list we can co-locate views for certain sections whilst still sharing common paths for components and layouts.

This PR tries out colocating all files related to applying into a single directory. I can see this being useful for other sections like tools, or login/register, where there's a fair amount of jumping around the codebase and it might be nice for all related files to live together.

🤔


davidrapson:
Going to come back to this once we've extracted out some of the application dashboard stuff to avoid too much churn.

Issue #1103, opened 2018-07-04
Spin out application form storage / dashboard into a separate app


mattandrews:
- [x] Create a new Express app containing the `Application` model which exposes an endpoint for storing application data and serves the existing dashboard
- [x] Configure Microsoft Azure Active Directory for secure staff authentication
- [x] Deploy this app to a new Elastic Beanstalk environment
- [x] Delete test database and reconfigure with correct VPC config
- [x] Protect dashboard screen with auth flow
- [x] Create production environment in Elastic Beanstalk
- [x] Set up deployment pipeline and tests
- [x] Add a status endpoint
- [x] Test with a large number of mock apps (pagination)
- [x] Add Sentry
- [x] Reinstate spreadsheet functionality
- [x] Update main app to `POST` applications to this new service

mattandrews:
The most important question: what should we call this new app? Suggestions:

- APPLICATRON 3000
- applyboard
- dashlication
- applicable
- not applicable

mattandrews:
There's also a question here about whether this whole thing runs in a VPC to further aid security.

PR #1104, opened 2018-07-04
Prevent direct access to the dashboard

mattandrews:
This is a bit of a sticking plaster until we move to VPC / complete #1103 but it's still worth doing.

This change means the app will no longer serve the application dashboard if accessed via anything other than the two Cloudfront-powered (and therefore WAF-protected) domains (or in DEV mode). This means direct access via IP/EC2 hostnames will no longer work, and therefore access is only available for an approved IP whitelist.

### DIFF
--- a/controllers/tools/applications.js
+++ b/controllers/tools/applications.js
@@ -19,7 +19,7 @@ const restrictAccess = (req, res, next) => {
     const shouldServe = appData.isDev || isCloudfront;
 
     if (!shouldServe) {
-        return res.redirect('/');
+        return res.redirect('/error-unauthorised');
     } else {
         return next();
     }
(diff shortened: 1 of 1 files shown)


controllers/tools/applications.js
```diff
@@ -1,15 +1,32 @@
...
 const XLSX = require('xlsx');
+const config = require('config');
 
 const applicationService = require('../../services/applications');
 const { purifyUserInput } = require('../../modules/validators');
 const { getFullUrl } = require('../../modules/urls');
+const appData = require('../../modules/appData');
+const { getSecret } = require('../../modules/secrets');
+
+const TEST_DOMAIN = process.env.TEST_DOMAIN || getSecret('test.domain');
```
mattandrews: (I added the `test.domain` secret yesterday)

PR #1115, opened 2018-07-13
Launch Building Connections [PENDING]

mattandrews:
We should merge/deploy this by 9.30 on Monday, then mark the programme page as active. 

We should also verify the live `applications` app is ready (eg. the test data is cleared) and all users who need immediate access are in IT's staff group.


mattandrews:
Realised we can merge this now and deploy on Monday.

PR #1119, opened 2018-07-17
Streamline database connection uris

davidrapson:
Moves towards using a single `DB_CONNECTION_URI` connection string instead of multiple parts. Maintains support for individual parts until parameter store has been updated.

### DIFF
--- a/controllers/tools/seed.js
+++ b/controllers/tools/seed.js
@@ -1,12 +1,12 @@
 'use strict';
-const { sampleSize } = require('lodash');
+const { sampleSize, startsWith } = require('lodash');
 const uuidv4 = require('uuid/v4');
 const userService = require('../../services/user');
-const appData = require('../../modules/appData');
 const aliases = require('../../controllers/aliases');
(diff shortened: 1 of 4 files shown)


controllers/tools/seed.js
```diff
@@ -1,12 +1,12 @@
...
+const { sampleSize, startsWith } = require('lodash');
 const uuidv4 = require('uuid/v4');
 const userService = require('../../services/user');
-const appData = require('../../modules/appData');
 const aliases = require('../../controllers/aliases');
+const { DB_CONNECTION_URI } = require('../../modules/secrets');
 
 function shouldMount() {
-    return appData.isNotProduction && process.env.USE_LOCAL_DATABASE;
+    return process.env.NODE_ENV !== 'production' && startsWith(DB_CONNECTION_URI, 'sqlite://');
```
davidrapson: This check is repeated both here and in the main model file. I'm _OK_ with this as a belt and braces thing rather than trying to share the logic.

PR #1124, opened 2018-07-18
[WIP] Strategic programmes

davidrapson:
This is a very rough prototype version of a strategic programmes pages. Mostly focussed on preparatory work to define structure and a few more generalised design patterns.

There's quite a lot going on here so will try and break it down by topic

## Colocating programme page templates

Moves programme templates around to colocate them together with controller code, for both regular programmes and strategic programmes.

## Tab updates

Updates tab macros to include pagination within the macro so we can reuse it on these new pages and is a more natural place for the logic to live.

## Media object cards

We have the canonical use for the "media object" pattern on these new pages so have created a general pattern for them:

image

## Promo cards

This attempts to better define the design pattern we use for promo/summary cards. Currently only used for case studies but these pages expand and establish that pattern as being used in more general terms.

image

> As an aside I realised that Nunjucks has a `call` method https://mozilla.github.io/nunjucks/templating.html#call which allows content to be passed into a macro as a block which helps a lot here.

## Strategic programmes

The rest of the changes are concerned with sketching out the structure of a strategic programme pages. I'm doing this by mocking the content API response we want inline in the controller. Still needs a bit of thought but once this feels more settled there will be a companion PR to add the fields we need to the content API.

image

image


controllers/programmes/index.js
```diff
@@ -0,0 +1,310 @@
...
+    });
+}
+
+/**
+ * Route Strategic Programme Detail
+ */
+function initStrategicProgrammeDetail(router) {
+    router.get('/strategic/headstart', function(req, res) {
+        // Mock entry
+        const entry = {
```
mattandrews: Could maybe store this mock data in a JSON file somewhere (in tests folder?) rather than having it in the history of this file etc?

davidrapson:
This PR is doing a **lot** now, and I've had a bit of a breakthrough on how to model and document common design patterns (like the media object and promo card) so I'm going to pull parts of this out into a fresh PR.

davidrapson:
Closing in favour of https://github.com/biglotteryfund/blf-alpha/pull/1131

PR #1125, opened 2018-07-19
Promo card and media object patterns

davidrapson:
More clearly defines a promo card and media object patterns. https://github.com/biglotteryfund/blf-alpha/pull/1124/files

Makes use of Nunjucks `call` block https://mozilla.github.io/nunjucks/templating.html#call which is amazing. It allows us to pass a block of custom content into a macro. I'm also using it as a way of improving how we document components. Each of these new components has a `macro` file which defines the template and a `examples` file which defines examples for use in the styleguide. Partly inspired by the structure used in https://github.com/alphagov/govuk-frontend

screen shot 2018-07-19 at 11 33 17

Updates case studies to use this new pattern, will also be used for research cards.


assets/sass/scaffolding/grid.scss
```diff
@@ -224,3 +224,40 @@ $num_columns_mobile: 3;
         }
     }
 }
+
+/* =========================================================================
+   Flex Grid
+   ========================================================================= */
+
+.flex-grid {
```
mattandrews: Is this similar to `.grid {}`?

davidrapson: Hmm…yeah not sure why I thought it was useful to have a different one. Let me take another look.

davidrapson: Looking at this again, the reason I went down this route is because the current grid has some issues with grid items that take up an equal height. This is probably fixable in general terms, but the number of generated classes and modifiers of the current grid system gives me pause as it feels a bit too magic each time I  use it. I think this is mostly a documentation thing, but for now I'd prefer to use something simpler here and coming back to do some work to make sense of the existing grid styles.

PR #1126, opened 2018-07-19
Add support for a number input field with currency label

mattandrews:
Looks like this:

image

I didn't add any further validation on the server, not sure if we massively need to enforce that at this stage, but thoughts welcome.


davidrapson:
Nice work

PR #1127, opened 2018-07-19
Add a basic demo form for Digital Funding (which copies RC)

mattandrews:
By request from Stuart – just to illustrate a form.


controllers/apply/digital-funding-demo/form-model.js
```diff
@@ -0,0 +1,234 @@
+'use strict';
+const { check } = require('express-validator/check');
+const { PROJECT_LOCATIONS } = require('../reaching-communities/constants');
```
davidrapson: Maybe copy-paste these into a new file too, just so we're not using any RC code (even if small)

PR #1128, opened 2018-07-19
Add project name to BC form

mattandrews:
Adds a missing field to the form. Tested and confirmed works properly on dashboard.


mattandrews:
True – we can keep an eye on it.

PR #1129, opened 2018-07-20
More design pattern prep

davidrapson:
More prep to share design patterns for strategic programmes pages. This PR pulls out the tab related changes from https://github.com/biglotteryfund/blf-alpha/pull/1124 and updates the pattern pages to auto generate the components list based on example files.

image


assets/js/modules/tabs.js
```diff
@@ -12,19 +11,6 @@ function scrollIntoView(el) {
     el.get(0).scrollIntoView(true);
 }
 
-function trackTabClick(label, trackTabClicksAsPageviews) {
```
davidrapson: To the comment on the previous PR. I'm removing this because we never enable it anywhere (as tracking separate pageviews was confusing analytics reports). Plus removing it simplifies the tab macros which are a little complicated. Happy to leave this in you have objections.

assets/sass/styleguide.scss
```diff
@@ -1,88 +0,0 @@
-/* =========================================================================
```
davidrapson: Inlining this into the layout as it's plain CSS and saves building another file in sass + keeps things in the same place.

assets/js/modules/tabs.js
mattandrews: Ah I forgot it was no longer enabled. Kill it then!

controllers/pattern-library/index.js
```diff
@@ -31,18 +25,26 @@ router.get('/components', (req, res, next) => {
...
+        try {
+            const matches = await globby(path.resolve(__dirname, '../../views/components') + '/**/examples.njk');
+            if (matches.length > 0) {
+                const componentSlugs = matches.map(match => path.basename(path.dirname(match)));
+                res.render(path.resolve(__dirname, './views/components'), {
+                    title: 'Components',
+                    componentSlugs
+                });
+            } else {
+                next(new Error('No components fount'));
```
mattandrews: I quite like the dialect here ("fount") but might be a typo...

PR #1135, opened 2018-07-23
Clean up components list

davidrapson:
Cleans up the component list templates to remove some old macros and imports and to add a skip-link nav to the listing page:

screen shot 2018-07-23 at 10 13 00

### DIFF
--- a/assets/sass/_mixins.scss
+++ b/assets/sass/_mixins.scss
@@ -98,7 +98,6 @@
         // eg. 80% of the max
         width: #{$widestScreen / 100 * 80};
         @if not $skipMargins {
-            height: 100%;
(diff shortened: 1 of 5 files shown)


assets/sass/_mixins.scss
```diff
@@ -98,7 +98,6 @@
         // eg. 80% of the max
         width: #{$widestScreen / 100 * 80};
         @if not $skipMargins {
-            height: 100%;
```
davidrapson: Noticed this was causing some display bugs with overflowing content when this class is used. Removing it fixes those bugs and afaict doesn't cause any regressions elsewhere.

PR #1136, opened 2018-07-23
Layout cleanup

davidrapson:
A few changes to clean up some layout classes. Particularly focussed on removing redundant code or single use classes that should be components.

The biggest change is refactoring the `responsiveGutter`. I spotted in https://github.com/biglotteryfund/blf-alpha/pull/1135 that the mixin has a `skipMargins` flag which is never called and we only use the mixin on `inner` classes. This PR simplifies these styles a lot, removing unused logic and inlining the styles into the `inner` classes themselves.


views/layouts/profiles.njk
```diff
@@ -13,18 +12,30 @@
...
-        <div class="page-section inner--wide-only accent--{{ pageAccent }} a--border-top">
-            <div class="page-section__content accent--{{ pageAccent }} a--border-top">
+        <div class="page-section u-inner-wide-only u-accent-border-top-{{ pageAccent }}">
+            <div class="page-section__content u-accent-border-top-{{ pageAccent }}">
                 {% for profile in profiles %}
-                    <section id="profile-{{ profile.title | slugify }}"
-                        class="content-box {% if loop.index > 1 %} accent--{{ pageAccent }} a--border-top{% endif %}">
-                        {{ profileSummary(profile) }}
-                    </section>
+                    <article id="profile-{{ profile.title | slugify }}"
```
davidrapson: This was a component but it's just text content. There were no styles associated with it,  and it's only used here, so inlining this.

assets/sass/_vars.scss
```diff
@@ -14,22 +14,15 @@ $breakpoints: (
     'medium-major': 840px,
     'large': 980px,
     'xlarge': 1200px,
+    // @TODO: Find-and replace these
```
mattandrews: Worth doing this as part of this change?

assets/sass/globals/layout.scss
```diff
@@ -2,38 +2,42 @@
...
-    max-width: $constrained;
-}
-
-.u-constrained-wide {
-    max-width: $constrainedWide;
+/**
+ * Inner/constrained layout classes
+ * .u-inner defines a constrained layout at all breakpoints
+ * .u-inner-wide-only defines a constrained layout at wider breakpoints only
+ * @TODO: Rename all uses of .inner* to u-inner*
```
mattandrews: Likewise above, might as well do this while we're fixing this here? Happy to help out!

views/layouts/profiles.njk
mattandrews: 👍

PR #1139, opened 2018-07-23
Proof of concept to replace homepage carousel with promo cards

davidrapson:
We know that nobody uses the carousel controls  and carousels in general aren't great. So with that in mind I wanted to see what the gain would be if we replaced the homepage case studies with promo-cards. Ignore the awkward image aspect ratios, mostly seeing how much we'd get to delete.

image


davidrapson:
🚀 Cleaned things up, let's do this thing!

PR #1142, opened 2018-07-24
Component macros

davidrapson:
Refactors up a few more components and adds them to the components list. I've tried to make each commit self contained enough so you can see what's going on.

Also updates the listing page to refine how the navigation works:

compnents


assets/sass/components/overlay-text.scss
```diff
@@ -1,63 +1,57 @@
...
+            @include on-interact {
+                .overlay-text__inner {
+                    background-color: darken($colour, 10%);
+                }
+            }
+        }
 
-    > span {
-        padding: 5px;
+        &.overlay-text--transluscent {
```
davidrapson: I pulled the accent classes into this file rather than having them in accents in order to make it clearer which component they relate to. Also previously this was called `is-opaque` but the work opaque describes the opposite of what these styles did 🤷‍♂️

views/components/overlay-text/macro.njk
```diff
@@ -0,0 +1,7 @@
+{% macro overlayText(text, linkUrl = false, accent = 'pink', isTransluscent = false) %}
+    <span class="overlay-text overlay-text--accent-{{ accent }}{% if isTransluscent %} overlay-text--transluscent{% endif %}">
```
davidrapson: The biggest change here was to refactor this so that it doesn't require a heading level (this macro is referenced with heading instead).

views/pages/funding/over10k.njk
```diff
@@ -5,7 +5,7 @@
 {% extends "layouts/main.njk" %}
 
 {% block content %}
-    {% set pageAccent = 'turquoise' %}
+    {% set pageAccent = 'blue' %}
```
davidrapson: This is not meant to be here…

assets/sass/components/overlay-text.scss
mattandrews: One of those words I always confuse with its antonym.

mattandrews: LIKE IMPOSSIBLE.

PR #1143, opened 2018-07-24
Migrated lambda function to a standalone repo

davidrapson:
Booped this over into a new repo https://github.com/biglotteryfund/deploy-notifier with a new name (as it's not just Slack anymore). This thing is deploy separately anyway so makes sense to separate it out.


mattandrews:
👍

PR #1144, opened 2018-07-25
Fix some Snyk warnings

mattandrews:


modules/appData.js
```diff
@@ -18,6 +19,7 @@ function getDeploymentData() {
 
 const deploymentData = getDeploymentData();
 const environment = config.util.getEnv('NODE_ENV'); // Defaults to 'development'
+debug(`App env is ${environment}`);
```
davidrapson: I think this is worth leaving in.

PR #1148, opened 2018-07-25
Shhh! Secrets

davidrapson:
This PR does some work to clean up our secrets to:

- remove unused secrets
- reduce console noise
- fix inconsistencies in camelCase vs WHATEVER_CASE_THIS_IS for environment variables

It flattens the DB connection credentials into a single `db.connection-uri` secret and adds a flag to mark if a secret is a hard requirements (i.e. we should throw a fatal error) and no longer logs out a warning in CI for optional secrets.

I've already updated parameter store with the new values so this is safe to merge assuming the changes look OK.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -45,7 +45,5 @@
   "googleAnalyticsCode": "UA-98908627-1",
   "legacyDomain": "https://wwwlegacy.biglotteryfund.org.uk",
   "ebulletinApiEndpoint": "https://apiconnector.com/v2",
-  "database": "website-dev",
-  "database-test": "website-test",
   "emailSender": "noreply@biglotteryfund.org.uk"
 }
--- a/controllers/tools/index.js
+++ b/controllers/tools/index.js
@@ -2,7 +2,6 @@
 const express = require('express');
 
 const { injectMerchandise } = require('../../controllers/funding/materials-helpers');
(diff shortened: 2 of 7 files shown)


config/default.json
```diff
@@ -45,7 +45,5 @@
   "googleAnalyticsCode": "UA-98908627-1",
   "legacyDomain": "https://wwwlegacy.biglotteryfund.org.uk",
   "ebulletinApiEndpoint": "https://apiconnector.com/v2",
-  "database": "website-dev",
```
davidrapson: These now live in the `db.connection-uri` secret in parameter store.

controllers/tools/index.js
```diff
@@ -40,10 +39,7 @@ router.route('/').get((req, res) => {
         { label: 'View a list of all published pages', href: '/tools/pages' },
         { label: 'View micro-survey results', href: '/tools/survey-results' },
         { label: 'View feedback results', href: '/tools/feedback-results' },
-        { label: 'View recent materials order stats', href: '/tools/order-stats' },
-        { label: 'View the Reaching Communities & Partnerships data dashboard', href: TOOLS_DATASTUDIO_URL },
```
davidrapson: These are overkill and are a bit of scope creep on what this index page should do, which is just list internal tools URLs.

mattandrews: Fair.

PR #1151, opened 2018-07-26
Component cleanup

davidrapson:
Some more component cleanup. Tidies up a couple more components and updates the documentation screens to include colour swatches and icons.

screen shot 2018-07-26 at 12 02 33
screen shot 2018-07-26 at 12 02 25


controllers/pattern-library/views/styles.njk
```diff
@@ -0,0 +1,105 @@
+{% extends "./layout.njk" %}
+{% from "components/styleguide.njk" import sgSectionHeader, sgExample %}
+{% import "components/icons.njk" as allIcons %}
+
+{% set colours %}
+    {{ sgSectionHeader('Colours') }}
+    {# Keep in sync with vars.scss #}
```
mattandrews: Wonder if there's a way to automate this?

PR #1156, opened 2018-07-27
Prototype navigation

davidrapson:
Work in progress but I'd like to get this merge in pieces. Adds a prototype version of the new header. The main thing this PR does so far is to add a way of toggling work in progress features using query strings which sets a cookie.

- Enable with: http://localhost:3000/?enable-feature=use-new-header
- Disable with: http://localhost:3000/?disable-feature=use-new-header

image


middleware/locals.js
```diff
@@ -20,6 +22,49 @@ module.exports = {
     middleware: function(req, res, next) {
         const locale = req.i18n.getLocale();
 
+        /**
+         * Allows feature flags to be passed through as query strings
+         * e.g. ?feature=use-new-header
+         * Useful for testing new features
+         */
+        res.locals.queryFeature = function(name) {
+            const featureNames = ['use-new-header'];
```
mattandrews: Maybe this should live in a config file so we don't have this string in multiple places?

PR #1157, opened 2018-07-30
Add FieldLevelEncryptionId to cloudfront config

davidrapson:
Just ran the update cloudfront script on TEST after too long and got this error `The parameter Field level encryption id is required`.

Looks like this is a recent schema change on the CloudFront end:

- https://github.com/boto/boto3/issues/1558
- https://github.com/ansible/ansible/issues/40724#issuecomment-397672401

This PR adds the missing field, which can be blank, which fixes the update script. I've already run this against test and confirmed it works.

### DIFF
--- a/config/cloudfront/live.json
+++ b/config/cloudfront/live.json
@@ -5,6 +5,7 @@
     "MinTTL": 0,
     "MaxTTL": 31536000,
     "DefaultTTL": 86400,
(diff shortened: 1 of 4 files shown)


mattandrews:
Good catch. The original CF config file we're basing this on is over a year old at this point so unsurprising a few things have changed. 👍

PR #1160, opened 2018-07-31
Improve build speeds for pull requests

davidrapson:
This PR does a couple of things to improve build speeds for pull requests:

- Updates unit tests and cypress tests to run in parallel
- Merges cypress tests suites to into either common or e2e to avoid extra boot up time between specs
- Switch to use travis build stages in order to avoid unnecessary steps when not deploying (particularly the `npm prune` stage

In good cases testing down from 7-8 minutes to 5-6 minutes. I do still see travis take ~2minutes to even reach the install stage so I'm not convinced we can shave _that_ much time off overall but when the wind is behind the build these changes should help.


.travis.yml
```diff
@@ -1,65 +1,77 @@
...
-node_js:
-- '8'
-before_install:
-- npm i -g npm@6
-install:
-- npm ci
+node_js: '8'
+
+# Install latest npm & dependencies
+before_install: npm i -g npm@6
```
mattandrews: Does the 6 here conflict with the 8 above?

cypress/integration/integration_spec.js
```diff
@@ -62,10 +242,10 @@ describe('e2e', function() {
         cy.get('#js-mobile-nav-toggle').as('navToggle');
         cy.get('#qa-offscreen-navigation').as('nav');
 
-        cy.get('@navToggle').click();
+        cy.get('@navToggle').click({ force: true });
```
mattandrews: What's this doing (and why)?

davidrapson: Oh good point, there was something funny going on with the scroll position when I upgraded Cypress, so this bypasses the visibility check. This was a workaround. It's not a great one though, so let me double check.

.travis.yml
davidrapson: Nope, this is **npm** 6. Node 8 comes with npm 5 but we want 6 for `npm ci`. This hasn't changed from the previous config. 

This is all moot if we upgrade to Node 10 as in this PR https://github.com/biglotteryfund/blf-alpha/pull/1162 as Node 10 is bundled with npm 6.

mattandrews: This is what I get for doing code reviews before coffee. Carry on!

cypress/integration/integration_spec.js
mattandrews: I don't mind the code, just wasn't clear what it was doing. Add a comment maybe?

davidrapson: If I can't find a better solution then will do

PR #1162, opened 2018-07-31
Node 10 and further build speed improvements

davidrapson:
Builds on from https://github.com/biglotteryfund/blf-alpha/pull/1160 and upgrades to node 10. There's a few dependency updates for one of two goals:

- Update necessary to support node 10
- Cleaning dependencies to further improve build times

For this PR build times have been consistently around the 5m30 mark which I'm pretty happy with.

screen shot 2018-07-31 at 16 35 09

Needs a fair bit of testing before merging.


package.json
```diff
@@ -142,7 +142,6 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-preset-env": "^1.7.0",
-    "common-tags": "^1.8.0",
```
davidrapson: I only added this in https://github.com/biglotteryfund/blf-alpha/pull/1123 but in retrospect this was a whole extra dependency for something that could be solved by indenting some code differently.

package.json
```diff
@@ -146,8 +146,7 @@
     "del": "^3.0.0",
     "dotenv": "6.0.0",
     "eslint": "5.2.0",
-    "eslint-plugin-cypress": "2.0.1",
```
davidrapson: Gets rid of `eslint-plugin-cypress` which was doing nothing more than setting a few globals we can do in a few lines of config.

package.json
```diff
@@ -48,8 +46,9 @@
           "modules": false,
           "targets": {
             "browsers": [
-              "last 2 versions",
-              "ie >= 10"
+              ">0.25% in GB",
```
davidrapson: See https://jamie.build/last-2-versions 

(Although not quite as aggressive as that suggestion in the article, still supports IE 10+)

assets/js/vue-apps/components/prompt.vue
```diff
@@ -1,7 +1,7 @@
 <script>
 const { trackEvent } = require('../../helpers/metrics');
 import $ from 'jquery';
-import { includes } from 'lodash';
+import includes from 'lodash/includes';
```
davidrapson: This one change allows us to remove the need for `babel-plugin-lodash` which basically just does this.

package.json
```diff
@@ -37,8 +37,6 @@
   },
   "babel": {
     "plugins": [
-      "lodash",
-      "transform-object-assign",
```
davidrapson: We don't use Object.assign anywhere in our client-side code so we can remove this babel plugin too.

.eslintrc.js
```diff
@@ -8,10 +8,7 @@ module.exports = {
         node: true
     },
     parserOptions: {
-        ecmaVersion: 8,
-        ecmaFeatures: {
-            experimentalObjectRestSpread: true
-        }
+        ecmaVersion: 9
```
davidrapson: This change is from upgrading to eslint 5

package.json
```diff
@@ -154,14 +154,13 @@
...
     "gulp-postcss": "^7.0.1",
-    "gulp-sass": "^3.1.0",
+    "gulp-sass": "4.0.1",
     "gulp-sourcemaps": "^2.6.1",
     "husky": "^0.14.3",
     "jest": "23.4.1",
     "lint-staged": "7.2.0",
     "livereload": "^0.7.0",
     "node-mocks-http": "^1.7.0",
-    "node-sass": "^4.5.3",
```
davidrapson: Removing this because `gulp-sass` already install a version of this as a dependency so avoids doubling up.

package.json
```diff
@@ -12,7 +12,7 @@
     "test-ci": "./bin/start-test-server & concurrently --success --raw 'npm test' 'npm run test-cypress'",
     "build": "gulp clean && gulp css && webpack",
     "build-production": "export NODE_ENV=production && npm run build && gulp manifest",
-    "build-autotrack": "autotrack -o public/autotrack/autotrack.js -p outboundLinkTracker,eventTracker,maxScrollTracker",
+    "build-autotrack": "npx autotrack -o public/autotrack/autotrack.js -p outboundLinkTracker,eventTracker,maxScrollTracker",
```
davidrapson: We commit the `public/autotrack/autotrack.js` file to the repo so we don't need to install `autotracker` every time. Instead we can use `npx` to install it on demand if/when we need to rebuild this file.

package.json
```diff
@@ -10,7 +10,7 @@
     "test-cypress": "wait-on http://localhost:8090 && cypress run",
     "pretest-ci": "npm run build",
     "test-ci": "./bin/start-test-server & concurrently --success --raw 'npm test' 'npm run test-cypress'",
-    "build": "gulp build && webpack",
+    "build": "gulp clean && gulp css && webpack",
```
davidrapson: Being explicit with these tasks means we can remove some code from the gulpfile.

davidrapson:
Added comments to each commit to explain what I'm changing.

davidrapson:
👀 

image

package.json
mattandrews: I remember upgrading this caused some build issues a few months ago (something to do with how the new `node-sass` interpreted our media queries) so let's keep an eye on this when we deploy it, though it's probably fine.

PR #1165, opened 2018-08-01
Revert build changes

davidrapson:
Don't want to faff around with build condition logic so reverting the travis build stages changes but keeping the concurrency improvements to tests. Means we still have the `npm prune` step but should still be faster relatively.


davidrapson:
Will probably come back to it once I've had time to read the docs a little more closely. I think I'm just misunderstanding the conditions.

davidrapson:
Edit: Worked out the condition based on https://github.com/travis-ci/travis-ci/issues/9614#issuecomment-388798396

PR #1168, opened 2018-08-01
Save BC application data for longer

mattandrews:
This change means that when a user submits a page of the BC form, their cookie is extended to last for one week, which effectively means their session data persists beyond the closure of the browser. They get a message notifying them of this:

image

It also tweaks the start button on the form depending on whether they've already started the form:

image

(top: user who has previously submitted a page, bottom: a new user)

@LynseyJReynolds had a few other feature suggestions which we should use for future forms:

- Collapse the save message after a few views, eg:
image

- Clarify which data has been restored when the user returns:
image
(this was tricky because it's not clear *when* they should see this)

- Allow the user to clear their data and start again


assets/sass/utilities/utilities-text.scss
```diff
@@ -56,3 +56,20 @@
         text-decoration: none;
     }
 }
+
+.icon--tick {
```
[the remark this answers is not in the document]
davidrapson: I'd maybe make this in an inline icon in `components/icons.njk` or failing that we have an `icons.scss` file too. If it stays as a CSS class maybe name it `o-icon-tick` or `icon-tick`. Not a modifier, as it's not modifying anything.

controllers/helpers/create-form-router.js
```diff
@@ -59,11 +71,24 @@ function createFormRouter({ router, formModel }) {
             }
         }
 
+        // for users submitting a step, increase their session expiry
+        // so they can save progress beyond a browser session
+        function extendSessionDuration(req) {
+            req.session.cookie.maxAge = moment()
+                .add(EXTENDED_SESSION_DURATION, 'days')
+                .toDate();
+            req.flash('progressSaved', {
```
davidrapson: Don't know if we need to use `flash` here, you could probably assign this to `res.locals.progressSaved` and use it directly in the template at that point, or at least from `renderStep`

config/default.json
```diff
@@ -43,6 +43,7 @@
     "features": "blf-features",
     "session": "blf-alpha-session"
   },
+  "extendedCookieDurationInDays": 7,
```
davidrapson: This feels more like a constant in that it's across all environments + specific to one controller so I'd be inclined to inline it.

views/pages/apply/form.njk
```diff
@@ -16,6 +16,20 @@
 
             {{ formErrors(errors = errors, title = "There was a problem with your submission") }}
 
+            {% set progressSaved = getFlash(request, 'progressSaved') %}
+            {% if progressSaved %}
+                <div class="u-constrained-wide u-constrained-content">
+                    <div class="message message--info">
+                        <h1 class="t2 icon--tick">Application progress saved</h1>
```
davidrapson: Is this the `h1` for the page at this point?

controllers/helpers/create-form-router.js
mattandrews: The challenge there is that `renderStep` is only called if there's a validation error, otherwise they're redirected to the next form step (which obviously loses the `res.locals` data), that's why I used `flash` initially...

davidrapson: Ah! missed that. Carry on.

views/components/icons.njk
```diff
@@ -104,3 +104,15 @@
...
     </svg>
 {% endmacro %}
+
+{% macro iconTick() %}
+    <svg viewbox="0 0 100 96"
+        width="20"
+        height="20"
+        xmlns="http://www.w3.org/2000/svg"
+    >
+        <title>tick by mikicon from the Noun Project</title>
```
davidrapson: Sorry to ⛏ but this title could do with updating.

mattandrews: I got the icon from the Noun Project which requires some attribution, couldn't think how else to do it..?

davidrapson: Fair enough, go for it then.

Issue #1130, opened 2018-07-20
Make postcode validation inline rather than after page submission


mattandrews:
Suggested by @LynseyJReynolds

mattandrews:
Done!

PR #1179, opened 2018-08-02
Check postcodes against the same pattern on client/server

mattandrews:
@LynseyJReynolds suggested this (in #1130) and it was easy enough to add.

Turns out the check we were using before (`isPostalCode()`) is defined [here](https://github.com/chriso/validator.js/blob/master/lib/isPostalCode.js#L54) so I took that regex directly and used it as a constant so we can pass it to the templates to validate inline, too. 

Was a bit fiddly passing it around as the HTML field needs it as a raw string whereas the validator needs it as a `RegExp` but this wasn't too painful either.

Minor tweak but means we can keep all of the BC form's validation on the client-side now.

postcode

### DIFF
--- a/controllers/apply/building-connections/fields.js
+++ b/controllers/apply/building-connections/fields.js
@@ -29,6 +29,13 @@ const LENGTH_HINTS = {
     }
 };
 
+// Allows us to share postcode validation on server and client-side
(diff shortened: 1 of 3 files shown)


mattandrews:
Sure, have a look now.

davidrapson:
💯 Nice

PR #1180, opened 2018-08-02
Buttons cleanup

davidrapson:
This PR is mostly focussed on cleaning up button styles. Particularly getting rid of some awkward modifiers.

The most controversial change is updating the 10k pages to use `section-links` rather than buttons. I've done this because a) it's more honest, these are links after all, and b) it allows us to avoid some awkward balancing of buttons for the uk wide option.

Before:

screen shot 2018-08-02 at 16 25 30

After:

screen shot 2018-08-02 at 16 11 22

screen shot 2018-08-02 at 16 14 49

Obviously warrants a take from @LynseyJReynolds + then a heads up on Slack for the rest of the team but I think it's worth the change.


mattandrews:
I'm not super keen on this, I think the bigger buttons were a bit friendlier / more obvious, though the horizontal list is easier to scan. Would defer to @LynseyJReynolds on this question though!

PR #1181, opened 2018-08-02
Survey cleanup

davidrapson:
Removes legacy responses. It's been long enough now that these don't have much value to us and are near the point where we'd want to delete them anyway due to age.

Also sorts "Pages with feedback" by count so we can more easily see which pages have the most responses.


views/tools/survey.njk
```diff
@@ -52,33 +52,14 @@
...
-        {% endif %}
     </section>
 
     {% if not pathQuery %}
         <section>
             <h2>Pages with feedback</h2>
             <div class="mb-5" style="max-height: 350px; overflow: auto">
                 <div class="list-group">
-                    {% for path, count in survey.groupedPaths %}
+                    {% for path, count in survey.groupedPaths | dictsort %}
```
davidrapson: Wrangled around for too long with combinations of lodash methods to do this when it turns out nunjucks is magic 🤷‍♂️

PR #1182, opened 2018-08-03
Add hover state and inbetween breakpoint to new nav

davidrapson:
This PR adds hover styles for the drop-down nav in the new header and adds "inbetween" breakpoint styles as well as generally refining things.

hover

screen shot 2018-08-03 at 11 06 04

This brings us to the point where the navigation prototype is largely "done" in that it's inline with design. The css and templates behind this could definitely stand to be cleaned up but that can wait until we've run some tests with this.

### DIFF
--- a/views/includes/global-header-next.njk
+++ b/views/includes/global-header-next.njk
@@ -150,3 +150,60 @@
         </div>
     </div>
 </header>
+
+<script>
+/**
+ * Initialise event listeners on header navigation early
+ * (e.g. search and menu toggle)
+ *
+ * We do this early so that users can interact
+ * with the navigation before the main script bundle has loaded
+ * increasing time-to-interactive
+ */
+(function() {
+    const elSelector = '.js-global-header';
+    var stateClassNames = {
+        nav: 'has-toggled-navigation',
+        search: 'has-toggled-search'
+    };
+
+    var body = document.querySelector('body');
+    var html = document.querySelector('html');
+    var el = document.querySelector(elSelector);
+    var searchInput = el.querySelector('input[type=search]');
+
+    if (!el) {
+        return;
+    }
+
+    var toggleNav = el.querySelector('.js-toggle-nav');
+    var toggleSearch = el.querySelector('.js-toggle-search');
+
+    toggleNav.addEventListener('click', function(e) {
+        e.preventDefault();
+        html.classList.remove(stateClassNames.search);
+        html.classList.toggle(stateClassNames.nav);
+    });
+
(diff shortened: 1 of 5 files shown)


views/includes/metaHead.njk
```diff
@@ -3,7 +3,7 @@
 <link rel="stylesheet" type="text/css" href="{{ 'stylesheets/style.css' | getCachebustedPath }}">
 <link href="https://fonts.googleapis.com/css?family=Poppins:300,500,600|Roboto:400,700" rel="stylesheet">
 
-<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=HTMLPictureElement,Promise,Array.prototype.find"></script>
+<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=HTMLPictureElement,Promise,Array.prototype.find,Element.prototype.closest"></script>
```
davidrapson: I've added `Element.prototype.closest` but all but IE11 supports it, and it's unlikely IE11 users would trigger this so we might be able to get away without this.

views/includes/global-header-next.njk
```diff
@@ -150,3 +150,60 @@
         </div>
     </div>
 </header>
+
+<script>
```
davidrapson: Slightly controversial but this is the quickest way to add event listeners for these elements, by triggering the script here the part of the DOM we care about is ready and it means users can click the navigation links before the rest of the page has finished resulting in a snappy time to interactive for common navigation tasks. 💭 thought's welcome on if there is a nicer way to approach this.

views/includes/global-header-next.njk
```diff
@@ -54,6 +66,28 @@
     </ul>
 {% endmacro %}
 
+{% macro globalSearch() %}
+    <form class="global-search-next" role="search" action="/search">
+        <input type="hidden" name="lang" value="en-GB">
```
mattandrews: Just looking at this... wonder if we still need these parameters now the search goes to Google?

views/includes/global-header-next.njk
```diff
@@ -103,8 +145,65 @@
...
+        nav: 'has-toggled-navigation',
+        search: 'has-toggled-search'
+    };
+
+    var body = document.querySelector('body');
+    var html = document.querySelector('html');
+    var el = document.querySelector(elSelector);
+    var searchInput = el.querySelector('input[type=search]');
+
+    if (!el) {
```
mattandrews: You could maybe move this up a bit earlier and bail out before doing the other queries?

PR #1192, opened 2018-08-14
Grid refactor

davidrapson:
Started this PR with the aim of pulling out one-off styles from the `.grid` classes into components. By the time I was finished all the previous grid styles were gone so we're left with the simpler flex-grid and everything else is in component styles.

Main changes are to:

- Introduce a `.o-button-group-flex` helper in addition to existing button groups to cover a common grid use case where buttons are aligned together. Avoids some awkward classes on 10k pages
- Extracts one-off grid styles for the materials form
- Extracts one-off grid styles for logos, mainly the overlay style. Also fixes a bug on the close button where the button styles had been removed by mistake
- Extracts out some one-off grid styles for the homepage social content
- Replaces remaining uses of the grid with the simpler `flex-grid` classes


assets/sass/components/forms.scss
```diff
@@ -409,12 +409,6 @@
         border: none;
     }
 
-    input[type='submit'] {
```
davidrapson: Removing this because this was a copy-paste error from the lines above so these styles are already set.

PR #1196, opened 2018-08-14
Prep for VPC move

mattandrews:
Updates deploys to fetch their secrets from `eu-west-2` (which has the same content as `eu-west-1`), meaning when we cut over to a new set of servers we don't need to do anything.

Also updates the bootstrap script to allow for pattern matches on the CodeDeploy group names so that the new VPC groups (with slightly different names) will still count as TEST/PROD deploys.

### DIFF
--- a/bin/get-secrets
+++ b/bin/get-secrets
@@ -5,7 +5,7 @@ const assert = require('assert');
 const AWS = require('aws-sdk');
 const { chunk, flatten, partition, unionBy } = require('lodash');
 
-AWS.config.update({ region: 'eu-west-1' });
+AWS.config.update({ region: 'eu-west-2' });
 
 const ssm = new AWS.SSM();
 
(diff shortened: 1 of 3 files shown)


bin/get-secrets
```diff
@@ -21,7 +21,7 @@ const argv = require('yargs').option('environment', {
  * Get Secrets
  * `describeParameters` returns up to **50** items but `getParameters`
  * will only return **10** items at a time so to get everything we
- * need we'll must chunk up our list of parameters and fetch them
+ * need we must chunk up our list of parameters and fetch them
```
davidrapson: lol, I need to turn spell checking on in my editor…

mattandrews:
Just done a final confirmation that `get-secrets` returns the same values for both `eu-west-1` and `eu-west-2` so will merge this now.

PR #1200, opened 2018-08-15
Navigation test tweaks

davidrapson:
A few minor tweaks to the dummy pages for the navigation test.

Also adds a prompt to ask people to take part in a version of the test using chalkmark.

Simplifies the prompt code as part of this to not require and external xhr call when we can just pass the values in. I think my original thinking was that we might need more than one prompt at a time with weighting, but really we just want to have one running at once so we can simplify.

### DIFF
--- a/assets/js/vue-apps/index.js
+++ b/assets/js/vue-apps/index.js
@@ -39,8 +39,7 @@ function initPrompts() {
     if (el) {
         new Vue({
             el: el,
-            components: { Prompt },
-            template: `<Prompt />`
+            components: { prompt: Prompt }
         });
     }
 }
--- a/views/pages/funding/thinking-of-applying.njk
+++ b/views/pages/funding/thinking-of-applying.njk
@@ -30,7 +30,7 @@
                         "title": "Is the Big Lottery Fund right for you?",
                         "imageUrl": "placeholders/placeholder-3.jpg"
                     }, {
-                        "title": "Find the right funding?",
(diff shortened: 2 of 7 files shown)


assets/js/vue-apps/index.js
```diff
@@ -39,8 +39,7 @@ function initPrompts() {
     if (el) {
         new Vue({
             el: el,
-            components: { Prompt },
```
davidrapson: Prompt Prompt Prompt Prompt

views/pages/funding/thinking-of-applying.njk
```diff
@@ -13,23 +14,26 @@
...
                 <ul class="flex-grid flex-grid--3up">
-                    {% set themes = cycler('blue', 'pink-dark', 'orange', 'green', 'turquoise', 'pink', 'cyan') %}
+                    {% set themes = cycler('blue', 'pink-dark', 'orange') %}
                     {% for page in [{
-                        "title": "Thinking of applying",
+                        "title": "Is the Big Lottery Fund right for you?",
                         "imageUrl": "placeholders/placeholder-3.jpg"
                     }, {
-                        "title": "What to expect if you apply",
+                        "title": "Find the right funding?",
```
mattandrews: Is this meant to be a question?

davidrapson: Is _this_ meant to be a question? (good catch!)

PR #1212, opened 2018-08-17
Remove tabs from logo downloads page

davidrapson:
Paired with @LynseyJReynolds on updating the logos page to avoid nested tabs. In fact it  now uses no tabs at all and instead shows all the content upfront. We're also merging the two versions of the "how to use our logo" into one as the only difference was the sample image which can be handled by showing both next to each other.

This plus the end result of https://github.com/biglotteryfund/blf-alpha/pull/1211 will allow us to completely remove the old tab macro.

localhost_3000_funding_funding-guidance_managing-your-funding_grant-acknowledgement-and-logos

### DIFF
--- a/views/pages/funding/logos.njk
+++ b/views/pages/funding/logos.njk
@@ -90,9 +90,21 @@
     {{ super() }}
 
     <div class="content-box u-inner-wide-only accent--{{ pageAccent }} a--border-top" id="segment-1">
+        <p><strong>{{ copy.locationPrompt }}</strong></p>
+        <div class="o-button-group-flex">
+                <a href="#region1" class="btn btn--small accent--{{ pageAccent }} a--btn js-tab">
+                    {{ copy.locations.monolingual }}
+                </a>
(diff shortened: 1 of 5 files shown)


davidrapson:
@mattandrews 

Added prompt back in as discussed:

image

views/pages/funding/logos.njk
```diff
@@ -89,10 +89,22 @@
...
+    <div class="content-box u-inner-wide-only accent--{{ pageAccent }} a--border-top" id="segment-1">
+        <p><strong>{{ copy.locationPrompt }}</strong></p>
+        <div class="o-button-group-flex">
+                <a href="#region1" class="btn btn--small accent--{{ pageAccent }} a--btn js-tab">
+                    {{ copy.locations.monolingual }}
+                </a>
+                <a href="#region2" class="btn btn--small accent--{{ pageAccent }} a--btn js-tab">
+                    {{ copy.locations.bilingual }}
+                </a>
+            </div>
```
mattandrews: Indentation a bit broken here?

PR #1225, opened 2018-08-22
Experimental: past grants prototype

davidrapson:
Raising for visibility. Requires https://github.com/biglotteryfund/grants-service to be set up and running


mattandrews:
Do you think we should run the grants service as:

- a standalone app
- a lambda function
- directly inside this app?

davidrapson:
Options 2 or 3 for me. 1 feels overkill given it's a single endpoint. I prefer the lambda function as we only need mongo for this one bit of functionality so I'm keen to keep that isolated. Appreciate that adds some complexity to deployments but we can solve that I think. Should it feel to granular falling back to bringing that code into the main app would be fine too.

mattandrews:
@davidrapson FYI I've just tidied this up a bit and added EC2 secrets for the two Lambda endpoints, and made the standalone controller into a page within Funding. Could you give it another look when you get time?

bin/import-grants
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+mongoimport -d grants -c blf --type csv --file $1 --headerline
```
davidrapson: We can remove this from here now right?

mattandrews: Yeah true, will do.

PR #1228, opened 2018-08-23
Remove closed apply routes

davidrapson:
Builds on https://github.com/biglotteryfund/blf-alpha/pull/1227

Removes the closed building connections fund code. Colocates remaining apply code and moves `route-helpers` into modules.


controllers/apply/index.js
```diff
@@ -1,11 +1,21 @@
...
 
 const appData = require('../../modules/appData');
-const { createFormRouter } = require('../helpers/create-form-router');
+const cached = require('../../middleware/cached');
+
 const reachingCommunitiesForm = require('./reaching-communities/form-model');
 const digitalFundingDemoForm = require('./digital-funding-demo/form-model');
 
+// How many days to extend the user's session by
+const EXTENDED_SESSION_DURATION_IN_DAYS = 7;
```
mattandrews: Possibly we can get rid of this as the session is now 7 days by default rather than a session cookie?

PR #1234, opened 2018-08-29
Debug step value mutation

davidrapson:
Found a function where step data is mutated where we don't want it to be. Starting with a failing test and working out from there.


davidrapson:
Fixed the site of the original bug so tests should now pass. Continuing to work through the form model to see if there are any other cases where data could be mutated and to simplify the shared code.

controllers/apply/__tests__/create-form-model.test.js
```diff
@@ -0,0 +1,101 @@
...
+            },
+            'step-4': {
+                'first-name': 'Example',
+                'last-name': 'Name',
+                email: 'example@example.com',
+                'phone-number': '03454102030'
+            }
+        };
+
+        const flattenedExpectation = {
```
mattandrews: Stealing this for a band name

PR #1245, opened 2018-08-30
Update mail service to determine appropriate send from address

### LINKED ISSUE #1239: Update email sending to use biglotteryfund.org.uk domain by default, unless internal email

davidrapson:
Adds a new `getSendAddress` which determines the best send from address to use based on the sender. i.e. if we're sending **to** and internal email we need to send from the internal domain, otherwise we can use the default.

The reason it's in a single `mail-helpers.js` file is because loading `mail.js` in tests is slow as it requires `server.js` to render emails and create nodemailer/cloudfront instances so ends up with lots of side effects that are not good for a unit test. Hints that there is some work to do to make the wider code more testable, but this suits for now.

Closes #1239

### DIFF
--- a/modules/__tests__/mail.test.js
+++ b/modules/__tests__/mail.test.js
@@ -1,0 +1,21 @@
+/* eslint-env jest */
+'use strict';
+const { getSendAddress } = require('../mail-helpers');
+
+describe('mail', () => {
+    const expectedDefault = `noreply@biglotteryfund.org.uk`;
+    const expectedInternal = `noreply@blf.digital`;
+
+    it('it should return default send from address for external send to addresses', () => {
+        expect(getSendAddress('example@example.com')).toBe(expectedDefault);
+        expect(getSendAddress('example@gmail.com')).toBe(expectedDefault);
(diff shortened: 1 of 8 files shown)


modules/__tests__/mail.test.js
```diff
@@ -0,0 +1,21 @@
...
+    it('it should return default send from address for external send to addresses', () => {
+        expect(getSendAddress('example@example.com')).toBe(expectedDefault);
+        expect(getSendAddress('example@gmail.com')).toBe(expectedDefault);
+    });
+
+    it('should return internal send from address for internal send to addresses', () => {
+        expect(getSendAddress('example@biglotteryfund.org.uk')).toBe(expectedInternal);
+        // Assert against similar looking but incorrect emails to test for false positives
+        expect(getSendAddress('example@biggerlotteryfund.org.uk')).toBe(expectedDefault);
+        expect(getSendAddress('example@biggestlotteryfund.org.uk')).toBe(expectedDefault);
```
mattandrews: 🤣

PR #1250, opened 2018-09-03
Refactor mail service, add more tests

davidrapson:
This PR refactors the mail service with a few goals:

- Reinstate name option from https://github.com/biglotteryfund/blf-alpha/pull/1249
- Simplify options passed through
- Add more unit tests

The main change is to extract out a `buildMailOptions` function which can be unit tested. Also makes use of JSDOC typedefs to enable some basic type-checking in VSCode (and presumably other editors) . Gave more confidence that the call sites have been updated correctly.


modules/__tests__/mail-helpers.test.js
```diff
@@ -0,0 +1,89 @@
...
+    it('should return internal send from address for internal send to addresses', () => {
+        expect(getSendAddress('example@biglotteryfund.org.uk')).toBe(expectedInternal);
+        // Assert against similar looking but incorrect emails to test for false positives
+        expect(getSendAddress('example@biggerlotteryfund.org.uk')).toBe(expectedDefault);
+        expect(getSendAddress('example@biggestlotteryfund.org.uk')).toBe(expectedDefault);
+        expect(getSendAddress('biglotteryfund.org.uk@example.com')).toBe(expectedDefault);
+    });
+});
+
+// function buildMailOptions({ subject, type = 'text', content, sendTo, sendMode = 'to', customSendFrom = null })
```
mattandrews: Can probably delete this?

davidrapson:
@mattandrews Just updated the tests to also mock out the nodemailer responses. The nodemailer docs had a pretty good guide on doing this https://nodemailer.com/transports/stream/

modules/__tests__/mail-helpers.test.js
```diff
@@ -0,0 +1,96 @@
...
+async function mockEmail(mailOptions) {
+    const transport = nodemailer.createTransport({
+        streamTransport: true,
+        newline: 'unix',
+        buffer: true
+    });
+
+    const info = await transport.sendMail(mailOptions);
+    // Remove generated properties from message
+    info.message = info.message
```
davidrapson: @mattandrews Switched to using snapshot tests as it allows us to assert against a real email response. Would appreciate thoughts. Particularly around this hackery which removes generated content from the responses so we can consider the snapshots static.

PR #1251, opened 2018-09-03
Mail refactor

davidrapson:
This is an attempt to make the `processor` for the reaching communities form unit tested. I think I have an approach that works but I'll add a few comments to point out some tradeoffs.


controllers/apply/reaching-communities/__tests__/processor.test.js
```diff
@@ -0,0 +1,62 @@
...
+            buffer: true
+        });
+
+        const result = await processor(
+            {
+                form: form,
+                data: flattenFormData(mockFormData),
+                stepsWithValues: stepsWithValues(form.steps, mockFormData)
+            },
+            transport
```
davidrapson: In order for this approach to work we have to pass a custom transport down through the different functions. This is the most awkward feeling part as we have to pass it down three levels. There might be a way around this if we pull the transport part up but at the expense of more boilerplate elsewhere. Conflicted.

package.json
```diff
@@ -13,7 +13,8 @@
   "scripts": {
     "start": "node ./bin/www",
     "startDev": "export DEBUG=biglotteryfund:*,express-session,connect:session-sequelize; nodemon --inspect ./bin/www",
-    "test": "npm audit && eslint . && jest --env=node --maxWorkers=4",
+    "test-unit": "export DB_CONNECTION_URI=sqlite://:memory && jest --env=node --maxWorkers=4",
```
davidrapson: The mail service relies on the main server file in order to render html templates. The upshot of this is that if required in unit tests it tries to connect to a database, setting this is the only way of making the tests run reliably. Not a bad thing to set but feels…awkward. Not sure we can get around this unless we rendered the emails using nunjucks directly but this would prevent us from using filters.

PR #1253, opened 2018-09-04
Digital fund prototype

davidrapson:


controllers/apply/digital-funding-demo/form-model.js
```diff
@@ -106,12 +106,11 @@ module.exports = strandNumber => {
         title: `Digital Funding Strand ${strandNumber} (Demo)`,
         pageAccent: 'blue',
         heroSlug: 'whizz-kidz',
-        hidePagination: true,
         shortCode: `DF${strandNumber}-ALPHA`,
         steps: strandSteps[`strand${strandNumber}`],
         processor: processor,
         startPage: {
-            template: path.resolve(__dirname, './startpage')
+            path: `/funding/programmes/digital-funding-demo/strand-${strandNumber}`
```
davidrapson: As this is the one part of "real" code we're touching, could we name this more explicitly, maybe `urlPath`. The `template` property is a path too, albeit a file path, but still keen to make the distinction clear.

assets/js/modules/forms.js
```diff
@@ -88,9 +88,21 @@ function initApplicationForms() {
...
+            e.preventDefault();
+            mockFormEl.classList.toggle('is-submitted');
+        });
+    }
+}
+
 function init() {
     initLegacyForms();
     initApplicationForms();
+    initMockEmailSignup();
```
davidrapson: Given how throwaway this is could we pop this in an inline `<script>` on the page?

assets/sass/components/_forms.scss
```diff
@@ -482,3 +494,43 @@
...
+.email-signup {
+    background-color: palette('pale-grey');
+    padding: $spacingUnit;
+
+    input[type='submit'] {
+        display: inline-block;
+        vertical-align: middle;
+    }
+
+    .post-submit {
```
davidrapson: :neckbeard: Mixing BEM classes makes me twitch a little…but this is fine for now.

controllers/apply/digital-funding-demo/form-model.js
mattandrews: Good call – done.

PR #1260, opened 2018-09-05
Inline images

davidrapson:
Having a little blitz of removing now-unneeded abstractions. This PR simplifies fallback images by inlining `fallbackSuperheroImage` into the one file it's used in, and setting the main `fallbackHeroImage` in `locals`. 

Means we can get rid of `createHeroImage` and `modules/images` completely.

🖼


davidrapson:
❤️

PR #1263, opened 2018-09-07
Unit test services & models

davidrapson:
Was looking for a way to unit test sequelize models. Realised focussing on testing the services that use them directly is the best direction as it handles coverage for both.

These test make use of the in-memory sqlite database in order to work so there's a little boilerplate on each one to se this up.

Starting with the feedback and survey services as they are easiest but if this feels OK we can add tests for the (more valuable) order and user services.

### DIFF
--- a/services/__tests__/surveys.test.js
+++ b/services/__tests__/surveys.test.js
@@ -1,0 +1,48 @@
+/* eslint-env jest */
+'use strict';
+
+describe('createResponse', () => {
+    let surveys;
+
+    beforeEach(async () => {
+        process.env.DB_CONNECTION_URI = 'sqlite://:memory';
+        const models = require('../../models');
+        await models.sequelize.sync();
+        await models.SurveyAnswer.destroy({ where: {} });
+        surveys = require('../surveys');
+    });
+
+    it('should create a survey response', async () => {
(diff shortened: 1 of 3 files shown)


services/__tests__/surveys.test.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    it('should create a survey response', async () => {
+        const yes = await surveys.createResponse({ choice: 'yes', path: '/' });
+        expect(yes.id).toBe(1);
+        const no = await surveys.createResponse({ choice: 'no', path: '/about', message: 'Test message' });
+        expect(no.id).toBe(2);
+    });
+
+    it('should get all respones', async () => {
```
mattandrews: Typo in "respones"

services/__tests__/surveys.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+        const responses = await surveys.getAllResponses();
+
+        expect(responses.totals.totalResponses).toBe(3);
+        expect(responses.totals.percentageYes).toBe(67);
+        expect(responses.totals.percentageNo).toBe(33);
+        expect(Object.keys(responses.groupedPaths).length).toBe(3);
+    });
+
+    it('should get respones for a given path', async () => {
```
mattandrews: And here! Sorry.

PR #1270, opened 2018-09-10
Update data page [PENDING]

mattandrews:
This changes a few things – adds a new map (UK countries only) and fetches data from the [new API endpoint](https://github.com/biglotteryfund/craft-dev/pull/90).

The map now looks like this:
image

In terms of deploying this: the CMS/API change needs to be live first, but the stats themselves aren't fully signed off yet (or translated) so we're waiting on those.


assets/sass/components/_data.scss
```diff
@@ -283,28 +300,8 @@ $mapRegions: (
     'yorkshire-and-the-humber': #ef7a70
 );
 
-.region {
-    fill: map-get($mapRegions, 'default');
-}
-
 @each $region, $colour in $mapRegions {
-    @if ($region != 'default') {
```
davidrapson: Nice 😎

mattandrews:
Yep, the local map should be fine (I renamed the original map to reflect the regional aspect).

Just updated this to add preview params. Should now be safe to merge before the new stats are added – I've migrated the existing (old) stats into this new form, and added the proposed new ones as a draft.

PR #1276, opened 2018-09-11
No top level controllers

davidrapson:
Builds on https://github.com/biglotteryfund/blf-alpha/pull/1274.

This PR removes any remaining controller code to standalone routers meaning we can drop the concept of a section-level controller. As a side-effect this also now means we can make `pages` an array.

FYI: For the time being I've removed `pagelist`, mostly just to let me get out of this refactoring rabbit hole alive. We can almost certainly reinstate it soon if it's still useful, but largely the CMS is becoming the best source of published pages anyway so not sure how essential it is. /post-rationalisation.


davidrapson:
Just done a bit of a pass to update any comments that are now incorrect to make sure everything is accurate.

davidrapson:
Waaay too much going on here so going to break things down a bit.

PR #1296, opened 2018-09-19
Add staff authentication

mattandrews:
Quite a lot going on here so would appreciate some testing / tyre-kicking.

This adds support for Azure authentication as a secondary Passport strategy (eg. alongside the existing `local` one). 

When a user signs in, we serialize them into an object like this:

        {
            userType: 'staff' || 'user',
            userData: { 
                username: 'foo'
            }
        };

Based on this, we can restrict access to different bits of the site by using new middleware like `ensureUserOnly` and `ensureStaffOnly`.

There's some hackiness in here: see the comments, but basically there's a longstanding bug with sessions/redirects and race conditions which mean I had to insert an interstitial redirect page to work around things. It's not too bad, though.

There's a huge amount of environment variables / secrets required here (which I haven't put into Parameter Store yet) – quite possibly this could be simplified. 

Otherwise – have at it.


config/test.json
```diff
@@ -1,5 +1,6 @@
 {
   "features": {
+    "azureAuthEnabled": false,
```
davidrapson: How come this is disabled in test?

mattandrews: Because the config isn't there yet (in parameter store) so it'll just fail. If we can reach consensus on how the secrets should be configured here, I'll store them there and update this.

davidrapson: 👍

middleware/authed.js
```diff
@@ -32,8 +26,28 @@ const requireUnauthed = (req, res, next) => {
     }
 };
 
+// Middleware for pages only authenticated users should see
+const ensureUserOnly = (req, res, next) => {
```
davidrapson: Not sure about this name? `userIsAuthed`, `ensureUserAuthed` or something?

middleware/authed.js
```diff
@@ -32,8 +26,28 @@ const requireUnauthed = (req, res, next) => {
...
+    if (req.isAuthenticated() && get(req, ['user', 'userType'], false) === 'user') {
```
davidrapson: Is `get(req, ['user', 'userType']` the same as `get(req, user.userType'`?

PR #1299, opened 2018-09-21
Add sort filters

mattandrews:


controllers/past-grants/index.js
```diff
@@ -51,20 +51,34 @@ router.get('/', async (req, res) => {
...
+    const queryWithPage = Object.assign({}, facetParams, sortParams, { page: req.query.page || 1 });
 
     const data = await request({
         url: PAST_GRANTS_API_URI,
         json: true,
         qs: queryWithPage
     });
 
+    // @TODO should we define this in the API?
+    const sortableFields = [
```
davidrapson: I think this makes sense to be here, makes translation easier at this end.

controllers/past-grants/views/past-grants.njk
```diff
@@ -10,7 +10,7 @@
             <h3 class="grant-result__title"><a href="#">{{ grant.title }}</a></h3>
             <p class="grant-result__summary">
                 £{{ grant.amountAwarded | numberWithCommas}} to
-                {{ grant.recipientOrganization[0].name }}
+                <strong>{{ grant.recipientOrganization[0].name }}</strong>
```
davidrapson: 💪

PR #1325, opened 2018-10-02
Digital fund preview feature

mattandrews:
Adds a preview flag so test users can complete the full user journey.

### DIFF
--- a/middleware/locals.js
+++ b/middleware/locals.js
@@ -172,8 +172,8 @@ module.exports = {
         }
 
         /* Configure per-feature queries for use in templates */
-        res.locals.featureUseNewHeader = queryFeature('use-new-header');
-        res.locals.featurePreviewDigitalFund = queryFeature('preview-digital-fund');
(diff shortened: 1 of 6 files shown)


middleware/locals.js
```diff
@@ -169,7 +169,11 @@ module.exports = {
             } else {
                 return false;
             }
-        };
+        }
+
+        /* Configure per-feature queries for use in templates */
+        res.locals.featureUseNewHeader = queryFeature('use-new-header');
+        res.locals.featurePreviewDigitalFund = queryFeature('preview-digital-fund');
```
davidrapson: I probably would've put the `appData.isNotProduction` line here so you don't need to repeat it in templates

mattandrews: Yeah I wondered about that – possibly there's a situation where we want a feature to be usable even in production?

davidrapson: Putting that check here wouldn't prevent that use-case would it? We're just moving the check up.

davidrapson:
This was a much bigger chef than anticipated

PR #1326, opened 2018-10-02
Todos

davidrapson:
To-did some to-dos.

- Simplify `navigationSections` logic so we don't have to construct so many things in the templates
- Remove request from locals, no longer used in templates
- Resolved a todo for injectFundingProgramme (used req.slug instead of manipulating url)
- Removed unhelpful analytics plugins maxScrollTracker is better served by hotjar and navigator.connection.effectiveType doesn't tell us much of use.


assets/js/bootstraps/analytics.js
```diff
@@ -1,11 +1,12 @@
 /* global ga, cxApi */
 
+const { forEach } = require('lodash');
 const { isDownloadLink } = require('../helpers/urls');
 const { trackEvent } = require('../helpers/metrics');
 
 function trackDocumentDownloads() {
     const links = document.querySelectorAll('.content-box a[href]');
-    [].forEach.call(links, function(link) {
+    forEach(links, function(link) {
```
mattandrews: Ah, this works on NodeLists?

PR #1327, opened 2018-10-02
Past grants AJAX support

mattandrews:
This adds support for `POST` requests to the past grants search endpoint.

These requests will use the same query object as the `GET` requests, but will instead return an object like this:

```
{
    status: 'success',
    meta: ...,
    facets: ...,
    resultsHtml: ...
}
```

(the `resultsHtml` is the output of rendering the results list to a template – we can then replace the in-page HTML with this, and use the facets/metadata to re-render the UI using VueJS).

### DIFF
--- a/controllers/past-grants/index.js
+++ b/controllers/past-grants/index.js
@@ -54,16 +54,7 @@ function buildPagination(paginationMeta, currentQuery = {}) {
  * @type {object}
  */
 function buildAllowedParams(queryParams) {
-    const allowedParams = [
-        'q',
-        'amount',
-        'postcode',
-        'programme',
-        'year',
-        'orgType',
-        'sort',
-        'country'
-    ];
+    const allowedParams = ['q', 'amount', 'postcode', 'programme', 'year', 'orgType', 'sort', 'country'];
     return pick(queryParams, allowedParams);
 }
 
--- a/controllers/past-grants/index.js
+++ b/controllers/past-grants/index.js
@@ -47,6 +47,41 @@ function buildPagination(paginationMeta, currentQuery = {}) {
     }
 }
 
+/**
+ * Pick out an allowed list of query parameters to forward on to the grants API
+ * @type {object}
+ */
+function buildAllowedParams(queryParams) {
+    const allowedParams = [
+        'q',
+        'amount',
+        'postcode',
+        'programme',
+        'year',
+        'orgType',
+        'sort',
+        'country'
+    ];
+    return pick(queryParams, allowedParams);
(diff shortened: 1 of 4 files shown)


controllers/past-grants/index.js
```diff
@@ -55,72 +83,94 @@ router.use(sMaxAge('1d'), injectBreadcrumbs, (req, res, next) => {
...
+    .route('/')
+    .post(async (req, res) => {
+        // @TODO how can this handle page numbers / sort options?
+        // Do we need hidden input fields?
+        const queryWithPage = addPaginationParameters(buildAllowedParams(req.body), req.body.page);
+        const grantData = await queryGrantsApi(queryWithPage);
+
+        // Repopulate existing app globals so Nunjucks can read them
+        // outside of Express's view engine context
+        const context = Object.assign({}, res.locals, { grants: grantData.results });
```
mattandrews: I'm actually fairly pleased with this solution – it was complaining before I put this in about some global functions (eg. `formatDate`). This way, we won't have silently-failing AJAX calls if we use a new global in the template or one of the macros it includes and forget to pass it in specifically.

controllers/past-grants/index.js
```diff
@@ -55,72 +83,94 @@ router.use(sMaxAge('1d'), injectBreadcrumbs, (req, res, next) => {
...
+
+        nunjucks.render(template, context, (renderErr, html) => {
+            if (renderErr) {
+                Raven.captureException(renderErr);
+                res.send({
+                    status: 'error'
+                });
+            } else {
+                res.send(
+                    JSON.stringify({
```
davidrapson: Why the json stringify?

mattandrews: Think I was worried about escaping HTML but looks like it's fine without – will remove.

controllers/past-grants/index.js
```diff
@@ -55,72 +83,92 @@ router.use(sMaxAge('1d'), injectBreadcrumbs, (req, res, next) => {
...
+
+        // Repopulate existing app globals so Nunjucks can read them
+        // outside of Express's view engine context
+        const context = Object.assign({}, res.locals, { grants: grantData.results });
+        const template = path.resolve(__dirname, './views/ajax-results.njk');
+
+        nunjucks.render(template, context, (renderErr, html) => {
+            if (renderErr) {
+                Raven.captureException(renderErr);
+                res.send({
```
davidrapson: Thinking about it you can use `res.json` to make sure you have an explicit content-type for this too.

PR #1328, opened 2018-10-02
Refactor sort controls UI

davidrapson:
There's definitely a way to make the markup less verbose but I've not quite worked that out yet. This works as a first pass though.

image

### DIFF
--- a/controllers/past-grants/index.js
+++ b/controllers/past-grants/index.js
@@ -117,46 +117,15 @@ router
         const queryWithPage = addPaginationParameters(facetParams, req.query.page);
         const data = await queryGrantsApi(queryWithPage);
 
-        const commonSortOptions = [
-            {
-                label: 'Oldest first',
-                value: 'awardDate|desc'
-            },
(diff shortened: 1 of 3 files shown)


controllers/past-grants/index.js
```diff
@@ -117,46 +117,15 @@ router
...
-            {
-                label: 'Lowest amount first',
-                value: 'amountAwarded|asc'
-            },
-            {
-                label: 'Highest amount first',
-                value: 'amountAwarded|desc'
-            }
-        ];
+        let currentSort = { type: 'awardDate', direction: 'desc' };
```
mattandrews: I feel like this should come from the API rather than here? Or could we send this parameter as an actual `?sort` option, so we know it's definitely the default (in case we change this at the API end)?

davidrapson:
Depends on https://github.com/biglotteryfund/grants-service/pull/13

mattandrews:
Nice – :shipit:

PR #1336, opened 2018-10-03
Add Vue components for past grants search [WIP]

mattandrews:
This probably shouldn't be merged until it's "done" but would be good to get an early review from @davidrapson to ensure it's going in the right direction.

Main takeaway – we should move some more responsibilities into the API:

- Add a "country" facet (which will require adding a field to the grant data)
- Do the "amount awarded" bucketing on the API end (eg. `0|10000` etc)

### DIFF
--- a/assets/js/vue-apps/grant-filters.js
+++ b/assets/js/vue-apps/grant-filters.js
@@ -9,28 +9,59 @@ function init() {
     new Vue({
         el: mountEl,
         delimiters: ['<%', '%>'],
-        data: {
-            facets: {}
+        data() {
+            let facets = {};
+            let defaultFilters = {
+                amount: '',
+                year: '',
+                programme: '',
+                country: '',
+                localAuthority: '',
+            };
+            let filters = Object.assign({}, defaultFilters);
+            return {
+                facets,
+                defaultFilters,
+                filters
+            }
         },
         mounted: function() {
+            // Enable inputs (they're disabled by default to avoid double inputs for non-JS users)
+            $(this.$el).find('[disabled]').removeAttr('disabled');
+            // Populate the facets object
             let facets = this.$el.getAttribute('data-facets');
             if (facets) {
                 this.facets = JSON.parse(facets);
             }
+            // Populate any existing query parameters
+            let queryParams = this.$el.getAttribute('data-query-params');
+            if (queryParams) {
+                this.filters = JSON.parse(queryParams);
+            }
         },
         methods: {
+            sort: function(sortKey, direction) {
+                this.filters.sort = `${sortKey}|${direction}`;
+                this.filterResults();
+            },
+            filtersToString: function() {
+                return Object.keys(this.filters).map(key => {
+                    return `${encodeURIComponent(key)}=${encodeURIComponent(this.filters[key])}`
+                }).join('&');
+            },
+            clearFilters: function() {
+                this.filters = Object.assign({}, this.defaultFilters);
+                this.filterResults();
+            },
             filterResults: function() {
-                const $results = $('#js-grant-results');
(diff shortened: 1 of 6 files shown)


mattandrews:
Pairs with https://github.com/biglotteryfund/grants-service/pull/17

assets/js/vue-apps/grant-filters.js
```diff
@@ -0,0 +1,93 @@
...
+                this.filterResults();
+            },
+            filterResults: function() {
+                const $form = $(this.$el);
+                const url = $form.attr('action');
+                const $results = $('#js-grant-results');
+                $results.text('Please wait, updating results...');
+                $.ajax({
+                    url: url,
+                    type: 'POST',
```
davidrapson: Looking at this now, I'd be strongly tempted to make the action here a separate GET endpoint for a couple of reasons.

- The regular/noscript version of the form doesn't use a post, it's a get action
- The post action makes client-side queries un-cacheable. Using a get endpoint makes sure these queries can be cached

davidrapson: If we really wanted to keep the same express handler / endpoint for both then http://expressjs.com/en/4x/api.html#res.format might help with this.

assets/js/vue-apps/grant-filters.js
```diff
@@ -0,0 +1,93 @@
...
+            };
+        },
+        mounted: function() {
+            // Enable inputs (they're disabled by default to avoid double inputs for non-JS users)
+            $(this.$el)
+                .find('[disabled]')
+                .removeAttr('disabled');
+            // Populate the facets object
+            let facets = this.$el.getAttribute('data-facets');
+            if (facets) {
```
davidrapson: I'm a little wary of passing JSON objects through as data params and re-parsing. I wonder if exposing this data on the page as global window variables might work? It's still a hack, but maybe avoids parsing a big object through an attribute.

controllers/past-grants/views/index.njk
```diff
@@ -83,45 +101,69 @@
...
+                                    }) }}
+
+                                    {% if queryParams.amount %}
+                                        <a class="search-filters__clear-facet" href="?{{ queryParams | removeQueryParam('amount') }}">
+                                            Clear amount
+                                        </a>
+                                    {% endif %}
+                                </noscript>
+
+                                <div class="js-only" v-cloak>
```
davidrapson: We could probably create a view component (`input-choice`) to mirror the macro, and avoid needing to repeat form markup everywhere.

davidrapson:
This approach feels really promising. Nice work.

I've left some comments covering my main concerns, mostly around using the cache and avoiding too much duplication. All solvable I think.

assets/js/vue-apps/grant-filters.js
mattandrews: This is done – think we should be okay using the same URL for both endpoints as we vary on `Accepts` in Cloudfront. We can test this though.

assets/js/vue-apps/grant-filters.js
mattandrews: Yeah, fair point. Have moved it to a global object which saves re-parsing JSON.

controllers/past-grants/views/index.njk
mattandrews: Will do this when we're a bit further along with the UI – suspect it'll change a bit more before we're "done" so no point optimising too early.

mattandrews:
Now pairs with https://github.com/biglotteryfund/grants-service/pull/18

controllers/past-grants/views/index.njk
davidrapson: Aggreed

PR #1338, opened 2018-10-03
Flip order of query feature flag

davidrapson:
Flip order of query feature flag so that programme appears _open_ by default. This is to avoid some issues with running remote user tests using the query string version of the URL. We'll want to reverse this again before we launch.


davidrapson:
This might've been a red herring / issue with the test rather than us so going to double check with a new participant now.

PR #1344, opened 2018-10-04
Digital fund translation

davidrapson:
- Extract out some small bits of copy in forms into translation files
- Rework strand templates to improve markup sharing, use a single template and defines a new "action-card"  component
- Adds translation copy for the key fund pages required for launch


davidrapson:
@mattandrews Bit of a big PR in the end. Added translations for the key pages required for launch (I have the rest already, but going to handle them in another PR). Also managed to clean up the templates a bit more so there's better markup sharing between the two strand pages.

config/locales/en.yml
```diff
@@ -894,10 +900,10 @@ research:
     contact: Contact
 apply:
   error:
-    title: There was an problem submitting your idea
```
davidrapson: "an" problem 💀

PR #1346, opened 2018-10-04
Youth capacity storage

davidrapson:
This PR reinstates application saving for the new youth capacity fund. Ran into the (known) issue that labels are missing by default as they come from translation copy so I've done some work to mix those in before storing. Includes tests and logic for determining the correct email based on region (although they are placeholders for now).

Thoughts very welcome.


controllers/apply/__tests__/Untitled-1
```diff
@@ -1,38 +1,39 @@
 /* eslint-env jest */
```
mattandrews: Is this filename right?

davidrapson: Um, nope that's not meant to be here at all!

PR #1347, opened 2018-10-05
Past grants improvements

mattandrews:
Quite a few tweaks in here:

Error handling:

image

Amount awarded facet now comes from the API:
image

Typed search works: 
past-grants-search

Pagination and sorting links are now dynamic:
past-grants-search-sort

Individual filters can be cleared:
past-grants-search-clear

And noscript support offers all the same fields (just not dynamic). Bingo!


assets/js/vue-apps/grant-filters.js
```diff
@@ -10,80 +11,145 @@ function init() {
...
                 this.filters.sort = `${sortKey}|${direction}`;
                 this.filterResults();
             },
+
+            // Generate CSS classes for sort links
+            sortLinkClasses: function(sortKey) {
+                const type = this.sort.type;
+                const dir = this.sort.direction;
+                return {
+                    'sort-controls__item--active': type === sortKey,
```
mattandrews: It might be a bit ugly having CSS classnames in JS, but on the other hand it looked quite ugly having chained JS logic in the HTML, so...

controllers/past-grants/views/index.njk
```diff
@@ -37,15 +49,22 @@
                                name="q"
                                value="{{ queryParams.q }}"
                                placeholder="Enter a search term, location, or postcode"
-                               v-on:change="filterResults"
+                               autocomplete="nope"
```
mattandrews: This is weird but [recommended by MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion).

davidrapson: Love it

assets/js/vue-apps/grant-filters.js
```diff
@@ -10,80 +11,145 @@ function init() {
...
                 country: '',
                 localAuthority: ''
             };
-            let filters = Object.assign({}, defaultFilters);
             return {
-                facets,
                 defaultFilters,
-                filters
+                sort: Object.assign({}, existingSort),
+                facets: Object.assign({}, existingFacets),
```
davidrapson: Do we need to assign when there is only one object?

assets/js/vue-apps/grant-filters.js
```diff
@@ -10,80 +11,145 @@ function init() {
...
+                searchError: false
             };
         },
+        watch: {
+            // Watch for changes to filters then make AJAX call
+            filters: {
+                handler: function() {
+                    this.filterResults();
+                },
+                deep: true
```
davidrapson: But how deep are you really :neckbeard:

assets/js/vue-apps/grant-filters.js
davidrapson: In which case let's go for a "state" class like `is-active`, `is-descending`, `is-ascending`, rather than using component-specific classes.

assets/sass/utilities/_utilities-display.scss
```diff
@@ -114,6 +114,10 @@
     @include visually-hidden();
 }
 
+.u-invisible {
```
davidrapson: Nothing to see here

assets/js/vue-apps/grant-filters.js
mattandrews: This was to clone them rather than extend anything, have been bitten in the past by mutation when just doing plain `facets = existingFacets`.

assets/js/vue-apps/grant-filters.js
mattandrews: 2deep4u

assets/js/vue-apps/grant-filters.js
mattandrews: Done!

mattandrews:
Made a few further tweaks:

Added a constituency filter (for the MPs):
image

Tweaked the detail page:
image

Added a read-only filter for recipient:
image

Added a read-only filter for org type:
image

(these two are read-only because we don't want a dropdown facet of recipients, and possibly org type is a bit pointless until we get more fine-grained results, but for now you can click through from the detail page to either of those things)

assets/js/vue-apps/grant-filters.js
```diff
@@ -1,5 +1,6 @@
 import $ from 'jquery';
 import Vue from 'vue';
+import _ from 'lodash';
```
davidrapson: Import the specific method otherwise this will pull in all of lodash

assets/js/vue-apps/grant-filters.js
```diff
@@ -10,80 +11,146 @@ function init() {
...
+                            dataType: 'json',
+                            success: response => {
+                                if (response.status === 'success') {
+                                    // @TODO vue-ize this
+                                    $('#js-grant-results').html(response.resultsHtml);
+                                    this.totalResults = response.meta.totalResults;
+                                    this.facets = response.facets;
+                                    this.updateUrl();
+                                } else {
+                                    if (response.error.error.error && response.error.error.error.code) {
```
davidrapson: error
error?
error!

PR #1351, opened 2018-10-08
Staff auth preview

davidrapson:
Switches preview functionality to use staff auth. Seems like this is simpler to do than I'd expected.


server.js
```diff
@@ -168,15 +167,14 @@ app.use(sessionMiddleware(app));
 app.use(passportMiddleware());
 app.use(redirectsMiddleware.common);
 app.use(localsMiddleware.middleware);
+app.use(previewMiddleware);
```
davidrapson: This was the key change, preview middleware needs to come _after_ passport middleware otherwise you get infinite redirects.

middleware/preview.js
```diff
@@ -1,23 +1,33 @@
...
-            const queryData = req.query[mode];
-            if (queryData) {
-                previewData.mode = mode;
-                previewData.id = parseInt(queryData);
-            }
-            return queryData;
-        });
-        res.locals.PREVIEW_MODE = previewData;
-    }
+        let previewMode = null;
```
davidrapson: This block is just cleanup. Making the logic more explicit.

PR #1352, opened 2018-10-08
More SPG tweaks

mattandrews:
This pairs with https://github.com/biglotteryfund/grants-service/pull/20.

I noticed that when you apply a sort to your results, it effectively means any future text search will be sorted by whichever sort parameter you applied (eg. not the `textScore`). This change means whenever the text query search is modified, it resets/ignores the previous sort value (eg. so the API knows to order by score).

Also in here: a pairing change to show the total awarded (adds a slightly delay onto the results but not really noticeable, and worth it IMO):

image

There was also a change from last week which sets the `isCalculating` toggle as soon as input is made which means the UI doesn't flicker so much:

### Before:
before

### After:
after

Minor downside is it may look like the search has started as soon as you press a key (whereas in fact the search is still debounced and doesn't start until 500ms after typing):

search

Finally we don't bother outputting default (eg. empty) filters for URLs / AJAX calls, eg:

### Before:

`?amount=&year=&programme=&country=&localAuthority=&constituency=&q=cats`

### After:

`?q=cats`

### DIFF
--- a/assets/js/vue-apps/grant-filters.js
+++ b/assets/js/vue-apps/grant-filters.js
@@ -39,11 +39,8 @@ function init() {
             // Watch for changes to filters then make AJAX call
             filters: {
                 handler: function() {
-                    if (this.filters.q) {
-                        this.ignoreSort = true;
-                    } else {
-                        this.ignoreSort = false;
-                    }
+                    this.ignoreSort = !!this.filters.q;
+                    this.isCalculating = true;
                     this.filterResults();
                 },
                 deep: true
@@ -66,6 +63,7 @@ function init() {
                     direction: direction
                 };
                 this.filters.sort = `${sortKey}|${direction}`;
+                this.isCalculating = true;
                 this.filterResults();
             },
 
@@ -124,7 +122,6 @@ function init() {
 
             // Send the data to the AJAX endpoint and output the results to the page
             filterResults: debounce(function() {
-                this.isCalculating = true;
                 this.searchError = false;
 
                 setTimeout(() => {
--- a/assets/js/vue-apps/grant-filters.js
+++ b/assets/js/vue-apps/grant-filters.js
@@ -1,6 +1,8 @@
 import $ from 'jquery';
 import Vue from 'vue';
 import debounce from 'lodash/debounce';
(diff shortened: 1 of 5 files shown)


assets/sass/components/_past-grants.scss
```diff
@@ -60,9 +60,12 @@
         }
 
         &.is-ascending::before,
-        &.is-ascending::after {
+        &.is-descending::before {
```
mattandrews: I think this was a typo in the original code?

assets/sass/components/_past-grants.scss
```diff
@@ -60,9 +60,12 @@
...
             position: relative;
             top: 2px; // Optically align
+            margin-right: 4px;
+            display: inline-block;
```
mattandrews: Added this to fix this:

image

controllers/past-grants/views/grant-results.njk
```diff
@@ -2,7 +2,7 @@
 {% from "components/split-nav/macro.njk" import splitNav with context %}
 
 {% macro grantResultItem(grant) %}
-    <div class="u-margin-bottom">
+    <div class="u-margin-bottom" data-score="{{ grant._textScore }}">
```
mattandrews: (Useful for debugging)

assets/js/vue-apps/grant-filters.js
```diff
@@ -82,9 +91,14 @@ function init() {
 
             // Convert filters into URL-friendly state
             filtersToString: function() {
-                return Object.keys(this.filters)
+                let filterClone = Object.assign({}, this.filters);
```
davidrapson: Maybe use `lodash/cloneDeep` or `lodash/clone` here?

davidrapson: Reading this back this sentence is confusing. I mean something like this:

```js
const filterClone = cloneDeep(this.filters);
const cleanFilters = this.ignoreSort && filterClone.sort ?  omit(filterClone, 'sort') : filterClone;
```

mattandrews: Good shout, give it a look now.

PR #1355, opened 2018-10-08
Clean up routes to improve legacy / archival behaviour

davidrapson:
Does a little work to improve a few things around legacy site and archival behaviour:

- Removes `removeCsp` in favour of removing directly in proxy code
- Separates out aliases from routes (import directly) and splits out vanity urls from more general redirect handlers

The biggest change is adding an interstitial screen to archived pages which a) improves the experience a bit by adding some context to why you are being redirected and b) has the added benefit of removing the need for server-side google analytics calls as we can capture archived pages as regular page views.

screen shot 2018-10-08 at 15 29 23

Appears in analytics as:

image


server.js
```diff
@@ -156,6 +159,8 @@ initViewEngine();
 /**
  * Register global middlewares
  */
+app.use(slashes(false));
+app.use(redirectsMiddleware);
```
davidrapson: Lifting the redirects middleware up to avoid initialising sessions / passport etc. before we redirect

views/static-pages/archived.njk
```diff
@@ -0,0 +1,27 @@
+{% from "components/hero.njk" import hero %}
+{% from "components/content-box/macro.njk" import contentBox %}
+
+{% extends "layouts/main.njk" %}
+
+{% block extraHead %}
+    <meta http-equiv="refresh" content="2;url={{ archiveUrl }}" />
```
mattandrews: Forgot this tag existed. Would you mind replacing the one in `user/interstitial.njk` with this same technique (and make it 2 seconds rather than 3 there)?

PR #1356, opened 2018-10-08
Youth Capacity tweaks [WIP]

mattandrews:
Might be a couple of final changes here but this is almost good to go so would appreciate a review. Mostly just copy tweaks.

### DIFF
--- a/views/components/form-fields/macros.njk
+++ b/views/components/form-fields/macros.njk
@@ -59,7 +59,6 @@
     />
 {% endmacro %}
 
-{# @TODO: Extracted currency field. Hook this up properly #}
 {% macro inputCurrency(field, fieldCopy) %}
     <label class="ff-label" for="field-{{ field.name }}">
         {{ fieldLabelText(field, fieldCopy) }}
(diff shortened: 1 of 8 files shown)


views/components/form-fields/macros.njk
```diff
@@ -168,6 +168,8 @@
                     {{ inputSelect(field, fieldCopy) }}
                 {% elseif field.type === 'checkbox' or field.type === 'radio' %}
                     {{ inputChoice(field, fieldCopy) }}
+                {% elseif field.type === 'currency' %}
+                    {{ inputCurrency(field, fieldCopy) }}
```
davidrapson: 👏  FYI There is a @TODO next to the macro for this that can be deleted now.

mattandrews:
Could you give this one more look @davidrapson?

views/components/programmes.njk
```diff
@@ -29,7 +29,7 @@
         {% endif %}
         {% if fields.applicationDeadline %}
             <dt>{{ __('global.programmes.programmeLabels.applicationDeadline') }}</dt>
-            <dd>{{ fields.applicationDeadline }}</dd>
+            <dd>{{ fields.applicationDeadline | safe }}</dd>
```
mattandrews: This is because they want a linebreak in that field but I can't change this in Craft until this fix goes out.

PR #1357, opened 2018-10-09
No top level routers

davidrapson:
This is a re-do of some of the parts in https://github.com/biglotteryfund/blf-alpha/pull/1276

- Moves API-related stuff to an `api` controller
- Removes need for section-level controllers (only page/path-level ones)
- Remove some unused locals and methods
- Converts pages in routes to an array, we don't use the key/id


modules/cloudfront.js
```diff
@@ -4,6 +4,38 @@ const { assign, compact, concat, flatten, flatMap, get, sortBy, uniq } = require
...
+ * or custom cookies to be whitelisted you must define those rules here.
+ */
+const CLOUDFRONT_PATHS = [
+    { path: '*~/link.aspx', isPostable: true, allowAllQueryStrings: true },
+    { path: '/api/*', isPostable: true, allowAllQueryStrings: true },
+    { path: '/funding/funding-finder', isPostable: true, allowAllQueryStrings: true, isBilingual: true },
+    { path: '/funding/programmes', queryStrings: ['location', 'amount', 'min', 'max'], isBilingual: true },
+    { path: '/funding/search-past-grants-alpha', isPostable: true, allowAllQueryStrings: true, isBilingual: true },
+    { path: '/search', allowAllQueryStrings: true, isBilingual: true },
+    { path: '/user/*', isPostable: true, queryStrings: ['token'] }
```
mattandrews: For the user endpoint, there are also these query parameters: `s` and `redirectUrl`.

davidrapson: These haven't changed so TBH we might not even need it in here as those pages are all uncached but will add it for completeness.

PR #1367, opened 2018-10-12
Grant detail improvements [WIP]

mattandrews:
This isn't done yet but worth flagging up.

Pairs with https://github.com/biglotteryfund/grants-service/pull/24

This makes an API call to fetch related grant data, and adds some skeleton page layout to match the new design (unfinished, but all the data should be in place now). Also passes some render options through to the AJAX-powered grant listing.

Needs work yet but would appreciate early thoughts.

### DIFF
--- a/assets/js/vue-apps/grant-detail.js
+++ b/assets/js/vue-apps/grant-detail.js
@@ -1,0 +1,42 @@
+import $ from 'jquery';
+import Vue from 'vue';
+
+function init() {
+    const mountEl = document.getElementById('js-grant-related-projects');
+    if (!mountEl) {
+        return;
+    }
+    new Vue({
+        el: mountEl,
+        delimiters: ['<%', '%>'],
+        data() {
+            return {
+                relatedGrants: false
+            };
+        },
+        mounted: function() {
+            $.ajax({
+                url: '/funding/search-past-grants-alpha/',
+                dataType: 'json',
+                data: {
+                    limit: 3,
+                    programme: 'Reaching Communities',
+                    localAuthority: 'E08000012',
+                    exclude: '0031045037',
+                    fuzzy: true,
+                    skipFacets: true,
+                    related: true
+                },
+                timeout: 20000
+            })
+            .then(response => {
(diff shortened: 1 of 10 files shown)


assets/js/vue-apps/grant-detail.js
```diff
@@ -0,0 +1,42 @@
...
+        data() {
+            return {
+                relatedGrants: false
+            };
+        },
+        mounted: function() {
+            $.ajax({
+                url: '/funding/search-past-grants-alpha/',
+                dataType: 'json',
+                data: {
```
mattandrews: Clearly this is a massive `@TODO` – eventually this will be read out from a global object like with the main grant search. For now just getting something on the page was the goal.

assets/js/vue-apps/grant-detail.js
```diff
@@ -0,0 +1,42 @@
...
+                    limit: 3,
+                    programme: 'Reaching Communities',
+                    localAuthority: 'E08000012',
+                    exclude: '0031045037',
+                    fuzzy: true,
+                    skipFacets: true,
+                    related: true
```
mattandrews: See thoughts on the linked PR with regards to these last 4 params.

mattandrews:
Have updated the styling here and tidied things up, also added a funding programme if we have it.

assets/sass/components/_past-grants.scss
```diff
@@ -128,3 +128,37 @@
         font-size: 15px;
     }
 }
+
+/* =========================================================================
+   Grant details
+   ========================================================================= */
+
+.is-active-project {
```
davidrapson: This should probably have a parent class associated with it. i.e. `grant-programme.is-active-programme` or `grant-programme--active`.

config/default.json
```diff
@@ -22,6 +22,7 @@
   "dateFormats": {
     "month": "MMMM YYYY",
     "short": "D MMMM, YYYY",
+    "slashed": "D/MM/YYYY",
```
davidrapson: Do we want a new date format?

PR #1374, opened 2018-10-17
Grants ui continued

davidrapson:
Refines sort behaviour, pulls sort options from the API reducing the amount of logic needed in the frontend.


davidrapson:
Depends on https://github.com/biglotteryfund/grants-service/pull/30

PR #1378, opened 2018-10-18
Add grant recipient page

mattandrews:
image

One slight quirk of this is organisations with one ID but multiple names: `GB-CHC-219279` is the Royal British Legion, but each grant has a different org name (eg. `Royal British Legion Tamworth Branch`). Not much we can do here as the original data doesn't give an overall/parent org title, so we just display the first (wrong) one. Something to ask the Data team about...


controllers/past-grants/views/grant-results.njk
```diff
@@ -1,40 +1,37 @@
 {% from "components/promo-card/macro.njk" import promoCard %}
 {% from "components/split-nav/macro.njk" import splitNav with context %}
+{% from './helpers.njk' import pgLink, buildLocationList with context %}
 
 {% macro grantResultItem(grant, options) %}
     <div class="u-margin-bottom-s{% if options.wrapperClass %} {{ options.wrapperClass }}{% endif %}" data-score="{{ grant._textScore }}">
         {% set subtitle %}
             £{{ grant.amountAwarded | numberWithCommas}} to
-            <strong>{{ grant.recipientOrganization[0].name }}</strong>
+            <strong>{{ grant.recipientOrganization[0].name | striptags }}</strong>
```
mattandrews: Not really sure why this was needed but Vue seemed to be adding `&nbsp;` in place of spaces (I checked, this isn't in the original data) on this page: `/funding/search-past-grants-alpha/grant/0030008252`:
image

This was breaking the grid layout.

controllers/past-grants/views/recipient-detail.njk
```diff
@@ -0,0 +1,116 @@
...
+                    <h2>Awarding programmes</h2>
+                        <ul>
+                            {% for programme in recipientProgrammes %}
+                                <li class="u-margin-bottom t5">
+                                    <a class="accent--pink a--text"
+                                       href="{{ pgLink('programme', programme.label) }}">
+                                        {{ programme.label }}</a>
+                                    {% if recipientProgrammes.length > 1 %}
+                                        <br />
+                                        <small>
```
davidrapson: Not sure about this award count? It is not going to say "1" a lot?

controllers/past-grants/views/recipient-detail.njk
```diff
@@ -0,0 +1,116 @@
...
+                                <a href="https://www.google.co.uk/search?q={{ orgNameSafe }}">
+                                    Search for {{ organisation.name }} with Google
+                                </a>
+                            </strong>
+                        </p>
+
+                    </div>
+                </div>
+
+                <div class="content-sidebar__secondary u-tone-background-tint u-padded">
```
davidrapson: Rather than adding `u-tone-background-tint u-padded` probably wrap this in a `card`. See related programmes on the research page for this same pattern.

controllers/past-grants/views/recipient-detail.njk
```diff
@@ -0,0 +1,116 @@
...
+
+            </div>
+        </div>
+
+        <div class="u-inner-wide-only">
+            <h2>{{ totalResults }} projects funded by Big Lottery Fund (£{{ totalAwarded }} awarded in total)</h2>
+            <div class="flex-grid">
+                {% for grant in recipientGrants %}
+                    <div class="flex-grid__item">
+                        <div class="content-box content-box--borderless u-tone-background-pale-grey u-block-full u-no-margin">
```
davidrapson: This feels like accidental styling, i.e. chaining a bunch of classes until we get what we want. Is there a pattern for this already, if not should there be?

controllers/past-grants/views/recipient-detail.njk
mattandrews: Possibly – it'll only appear for orgs who've had grants in multiple programmes, though.

controllers/past-grants/views/recipient-detail.njk
mattandrews: 👍

controllers/past-grants/views/recipient-detail.njk
mattandrews: Good call, the action card does this. Have a look now!

controllers/past-grants/views/recipient-detail.njk
davidrapson: Oh yeah, was mostly going on the original screenshot tbh and didn't spot that.

davidrapson:
👍

PR #1383, opened 2018-10-19
Past grants ui fixes

mattandrews:
Few things in here, mostly just enabling non-JS search etc.

### DIFF
--- a/assets/js/vue-apps/past-grants.js
+++ b/assets/js/vue-apps/past-grants.js
@@ -59,6 +59,11 @@ function init() {
             }
         },
         mounted() {
+            // Enable inputs (they're disabled by default to avoid double inputs for non-JS users)
+            $(this.$el)
+                .find('.js-only[disabled]')
+                .removeAttr('disabled');
+
             window.onpopstate = event => {
                 const historyUrlPath = get(event, 'state.urlPath');
                 historyUrlPath && this.updateResults(historyUrlPath);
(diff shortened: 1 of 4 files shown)


assets/js/vue-apps/past-grants.js
```diff
@@ -59,6 +59,11 @@ function init() {
             }
         },
         mounted() {
+            // Enable inputs (they're disabled by default to avoid double inputs for non-JS users)
```
mattandrews: This really needed to be here otherwise for non-JS users the form has two `q` inputs which both get sent (and the API fails)

controllers/past-grants/views/index.njk
```diff
@@ -42,34 +42,25 @@
...
-                            :value="filters.q"
-                            autocomplete="no"
-                            @input="debounceQuery">
-
+                               disabled="disabled"
+                               type="search"
+                               id="search-query"
+                               name="q"
+                               :value="filters.q"
+                               autocomplete="off"
```
mattandrews: Weirdly "no" seems to have changed and shows suggestions (at least, for me) so `off` stops it again.

assets/js/vue-apps/past-grants.js
davidrapson: Ah, thought I'd resolved this by virtue of the fact Vue now replaces the markup for all the inputs, but forgot about the search form! Good catch

controllers/past-grants/views/index.njk
```diff
@@ -78,7 +69,7 @@
                     </p>
                 </div>
 
-                <div class="grants-search{% if meta.totalResults === 0 %} has-no-results{% endif %}" v-if="totalResults > 0">
```
mattandrews: This class meant that if you reloaded a page with 0 results (with JS on) and cleared the filters, you'd still see no results as this class hid them.

PR #1387, opened 2018-10-19
Past grants landing page

mattandrews:
Depends on https://github.com/biglotteryfund/craft-dev/pull/94

Tweaks the past grants landing page (`/funding/past-grants`) to serve this if the `enableNewPastGrantsSearch` flag (added in https://github.com/biglotteryfund/blf-alpha/pull/1386) is enabled:

image

Ideally the featured grants would be enhanced with the grant details, eg:

image

This would require an API method to lookup grant IDs in bulk, though, so for now I think this is an acceptable halfway house.


mattandrews:
(on Prod I've added `grantId` to almost all of the existing case studies so there'll be a richer mix of them there)

controllers/funding/index.js
```diff
@@ -30,4 +31,23 @@ router.get('/', sMaxAge('30m'), injectHeroImage('active-plus-communities'), inje
...
+    '/past-grants',
+    injectCopy('funding.pastGrants'),
+    injectHeroImage('active-plus-communities'),
+    async (req, res, next) => {
+        let caseStudies = await contentApi.getCaseStudies({
+            locale: req.i18n.getLocale()
+        });
+
+        // Shuffle the valid case studies and grab the first few
+        caseStudies = take(shuffle(caseStudies.filter(c => c.grantId)), 3);
```
davidrapson: Mentioned on the other PR but I think lodash's `sampleSize` will do this for you.

PR #1390, opened 2018-10-22
Translate past grants [WIP]

mattandrews:
Still waiting for the actual Welsh translations and there's also a few things to do on the API end (watch this space!), but this is worth a review.

### DIFF
--- a/controllers/funding/index.js
+++ b/controllers/funding/index.js
@@ -44,7 +44,7 @@ router.get(
         caseStudies = take(shuffle(caseStudies.filter(c => c.grantId)), 3);
 
         res.render(path.resolve(__dirname, './views/past-grants'), {
-            title: 'Search awarded grants: Beta',
+            title: res.locals.copy.title,
             caseStudies: caseStudies
         });
     }
--- a/controllers/funding/views/past-grants.njk
+++ b/controllers/funding/views/past-grants.njk
@@ -28,7 +28,6 @@
                             promptLabel = copy.survey.prompt,
                             fieldLabel = copy.survey.label
                         )  }}
-
                         <hr />
                         <p>{{ copy.optionsIntro }}</p>
                     </section>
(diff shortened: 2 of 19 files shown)


controllers/past-grants/views/index.njk
```diff
@@ -13,7 +13,8 @@
             queryParams: {{ queryParams | dump | safe }},
             totalResults: {{ meta.totalResults }},
             totalAwarded: {{ meta.totalAwarded}},
-            sort: {{ meta.sort | dump | safe }}
+            sort: {{ meta.sort | dump | safe }},
+            lang: {{ copy | dump | safe }}
```
mattandrews: Open to suggestions on other ways to approach this, but generally it seems to work reasonably well here.

davidrapson: I think this is pretty sensible for now. Longer term I was starting to look at https://www.i18next.com/ which looks to be a more full featured i18n library that supports hybrid server/client stuff. Only mentioning in case we do more of this. For now this totally works.

assets/sass/components/_forms.scss
```diff
@@ -133,6 +133,15 @@
     display: block;
     margin-bottom: 5px;
     font-weight: bold;
+
+    &.ff-label--append:after {
+        margin-left: 5px;
+        content: attr(data-append);
```
davidrapson: Oh! Why didn't I think of doing this before. I added a `ff-label-note` class a while back that had the text inline in the html when I translated the forms, but this is nicer. Might circle back and change my original stuff to match this.

controllers/funding/index.js
```diff
@@ -30,4 +31,23 @@ router.get('/', sMaxAge('30m'), injectHeroImage('active-plus-communities'), inje
...
+    '/past-grants',
+    injectCopy('funding.pastGrants'),
+    injectHeroImage('active-plus-communities'),
+    async (req, res, next) => {
+        let caseStudies = await contentApi.getCaseStudies({
+            locale: req.i18n.getLocale()
+        });
+
+        // Shuffle the valid case studies and grab the first few
+        caseStudies = take(shuffle(caseStudies.filter(c => c.grantId)), 3);
```
davidrapson: I think lodash has a dedicated method called `sampleSize` which does this for you.

controllers/funding/views/past-grants.njk
```diff
@@ -0,0 +1,129 @@
...
+                            </div>
+                        </li>
+                    </ul>
+
+                {% else %}
+                    <section class="s-prose u-constrained-content-wide">
+                        {{ __('funding.pastGrants.intro', grantNavLink) | safe }}
+                    </section>
+
+                    <form class="grants-searchbar"
```
davidrapson: Guessing we should make this a component at some point?

mattandrews: I did think this, but then the other version of it uses the JS-powered one by default, which we didn't need here.

mattandrews:
@davidrapson could you review the last 3 commits? Have fixed conflicts.

PR #1392, opened 2018-10-23
Add basic vue component unit tests

davidrapson:
This PR puts tooling in place to be able to unit test our vue components. Definitely not comprehensive yet but adds an example test for:

- Basic interactions by toggling state
- Snapshot test for a functional component
- Snapshot test for a full component

I'll need to rework this to account for translations, but review still welcome now.

### DIFF
--- a/assets/.eslintrc.js
+++ b/assets/.eslintrc.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-    extends: ['eslint:recommended', 'plugin:vue/recommended', 'plugin:compat/recommended'],
+    extends: ['eslint:recommended', 'plugin:vue/strongly-recommended', 'plugin:compat/recommended'],
     settings: {
         polyfills: ['promises']
     },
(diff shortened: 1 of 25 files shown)


davidrapson:
Spotted as part of this that whilst we have https://github.com/vuejs/eslint-plugin-vue installed it wasn't actually running against `.vue` files 🙈 . Fixed that and auto-fixed a few warnings, as well as disabling any rules that are overkill or are handled by prettier. Leaves a few warnings about prop-types which sound sensible but I'll come back and fix.

davidrapson:
@mattandrews Mind giving this another look 👀

assets/.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-    extends: ['eslint:recommended', 'plugin:vue/recommended', 'plugin:compat/recommended'],
+    extends: ['eslint:recommended', 'plugin:vue/strongly-recommended', 'plugin:compat/recommended'],
```
mattandrews: 💪

davidrapson: Weirdly "strongly-recommended" is weaker than "recommended" 🤷‍♂️

PR #1395, opened 2018-10-23
A11y tests

davidrapson:
Realised that due to the way we run our Cypress tests using a standalone test server we can probably use that same server to run our a11y smoke tests in the main repo rather than on a cron job. This PR is an attempt to get them running here.

As a result it fixes a few errors flagged by the latest version of `axe-cli`


views/components/nav.njk
```diff
@@ -13,10 +13,8 @@
 {% macro navSearchForm(isOffset = false) %}
     {% set id = 'global-search-input-' | appendUuid %}
     <form class="global-search{% if isOffset %} global-search--offset{% endif %} js-global-search-form" role="search" action="/search">
-        <label class="u-visually-hidden" for="{{ id }}">
-            {{ __("global.misc.search") }}
-        </label>
         <input
+            aria-label="{{ __("global.misc.search") }}"
```
davidrapson: Axe now flags that explicit `<labels>` must be visible, so for a hidden label it's more appropriate to use `aria-label`. TIL

controllers/past-grants/views/index.njk
```diff
@@ -45,31 +45,54 @@
...
+                            value="{{ copy.submit }}"
+                            @click="updateQuery"
+                        />
                     </div>
                     <p class="ff-help u-no-margin">
                         {{ copy.helpText }}
                     </p>
                 </div>
 
+                <noscript>
```
davidrapson: Axe was complaining about duplicate IDs so to make this easier to manage I'm duplicating the whole block of markup for noscript rather than trying to interleave stuff. This does mean that we can probably move the js-version of the search bar into a vue component if we want too.

config/locales/en.yml
```diff
@@ -543,6 +543,9 @@ funding:
         reset: Reset filters
         clearSelection: Clear selection
         toggle: Toggle
+        organisationLegend: Organisation
```
davidrapson: This separates out the FieldGroup names from the individual fields in those groups.

assets/js/vue-apps/components/facet-choice.vue
```diff
@@ -27,26 +27,27 @@ export default {
...
-                    :checked="option.value === value"
-                    @input="$emit('input', $event.target.value)"
-                />
-                <label class="ff-choice__label" :for="fieldId(index)">
-                    {{ option.label }}
-                </label>
-            </li>
-        </ul>
-
+    <div>
```
davidrapson: A bit of awkward markup juggling, but axe warned about having the buttons inside the fieldset. So grouping them separately.

assets/js/vue-apps/components/__tests__/__snapshots__/facet-select.test.js.snap
```diff
@@ -6,7 +6,7 @@ exports[`FacetSelect should render with optgroup 1`] = `
 >
   <label
     class="ff-label"
-    for="field-dynamic-nodejs"
```
davidrapson: lol. This was just wrong before.

config/locales/en.yml
mattandrews: 👍

PR #1403, opened 2018-10-26
Migrate our people section

davidrapson:
Dependent on https://github.com/biglotteryfund/craft-dev/pull/95 and content being migrated over to the new CMS.

Adds a reworked Our People section using a new API endpoint which allows us to fetch (and migrate) all people pages not just the senior management team.

Uses a single API endpoint https://cms.blf.digital/api/v1/en/our-people rather than also having a dedicated slug endpoint. Craft is smart enough to still handle drafts correctly here.

This allows us to render a secondary navigation in the sidebar and use the same API call for the listing and detail pages.

Includes some grusome sounding methods like `injectOurPeople` just in time for halloween. I promise I mean no harm 💉 👻


davidrapson:
@mattandrews Reworked this a bit to make it possible to gradually migrate these pages rather than needing to do a big content migration before I can merge this. Mind giving this another look?

mattandrews:
Looks sensible 👍

PR #1405, opened 2018-10-30
Add basic event tracking for past grants filters and UI

mattandrews:
Adds some analytics event tracking for various bits of the past grants search.

### DIFF
--- a/assets/js/vue-apps/components/facet-choice.vue
+++ b/assets/js/vue-apps/components/facet-choice.vue
@@ -2,7 +2,7 @@
 import take from 'lodash/take';
 
 export default {
-    props: ['value', 'type', 'name', 'label', 'hideLabel', 'options', 'optionLimit', 'copy', 'handleActiveFilter'],
+    props: ['value', 'type', 'name', 'label', 'hideLabel', 'options', 'optionLimit', 'copy', 'handleActiveFilter', 'trackUi'],
     data() {
         return { isOpen: false };
     },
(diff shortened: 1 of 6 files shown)


assets/js/vue-apps/components/facet-choice.vue
```diff
@@ -28,6 +28,12 @@ export default {
         },
         fieldId(index) {
             return `field-dynamic-${this.name}-${index}`;
+        },
+        toggle() {
+            this.isOpen = !this.isOpen;
+            let action = 'Toggle additional options ';
+            action += this.isOpen ? 'on' : 'off';
+            this.trackUi(action, this.name);
```
davidrapson: You might be able to do this declaratively using `data-ga-*` events in the markup.

assets/js/vue-apps/components/facet-disclose.vue
```diff
@@ -13,6 +13,14 @@ export default {
         ariaId() {
             return `facet-disclose-${this.id}`;
         }
+    },
+    methods: {
+        toggle() {
+            this.isOpen = !this.isOpen;
+            let action = 'Toggle disclose ';
+            action += this.isOpen ? 'on' : 'off';
+            this.trackUi(action, this.filterName);
```
davidrapson: As above

assets/js/vue-apps/components/facet-choice.vue
mattandrews: I think I prefer leaving them as-is for now as the some of the attributes are a bit tricky to define in one-liners (eg. the above), and it'll mean hardcoding strings like `Past Grants Interface` everywhere – I like having the two tracking functions defined next to each other in one place.

Issue #1397, opened 2018-10-24
Separate out UK-wide funding programmes from country-specific ones


mattandrews:
(eg. include all programmes on a page, but split them apart so country-specific ones are more prominent)

PR #1408, opened 2018-10-30
Show UK-wide programmes in the funding finder again

mattandrews:
Fixes https://github.com/biglotteryfund/blf-alpha/issues/1397

Adds region groups (eg. a specific country, then UK-wide) to all funding programme lists where a country is specified, eg:

image

... then:

image

One mild quirk of this is that the Welsh version of the above page still says "Scotland" for the heading, not "Yr Alban" – this is because the Scottish programmes don't have Welsh translations so we're using the English, whose title is obviously in English too. Don't think this is the end of the world.

### DIFF
--- a/controllers/programmes/index.js
+++ b/controllers/programmes/index.js
@@ -1,6 +1,6 @@
 'use strict';
 const path = require('path');
-const { concat, map } = require('lodash');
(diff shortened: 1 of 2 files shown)


mattandrews:
Yeah, maybe down the line we'd do that, but right now think we're okay with this.

Issue #585, opened 2018-01-23
Add a field to CMS programme pages for post-tab content


mattandrews:
Some pages (eg. AFA Scotland) require a set of partner logos below tab content. Currently we have no way to display "footer" content below the tabs. 

- [x] Add a field to the `TEST` and `PRODUCTION` CMS to allow for "footers"
- [x] Add a field to the Content API to output these "footers"
- [x] Add the "footer" content to the programme detail template
- [ ] Modify the AFA Scotland page to put the logo here instead

Open for debate about the name for this field – in the (WIP) funding guidance section we used "intro" and "outro" to represent these fields. Thoughts @davidrapson?

PR #1409, opened 2018-10-30
Add programme footer

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/100

(Fixes https://github.com/biglotteryfund/blf-alpha/issues/585)

### DIFF
--- a/controllers/programmes/views/programme.njk
+++ b/controllers/programmes/views/programme.njk
@@ -69,6 +69,12 @@
                         </div>
                     </div>
                 {% endif %}
+
+                {% if entry.footer %}
+                    <div class="content-box content-box--frameless">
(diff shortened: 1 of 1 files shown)


controllers/programmes/views/programme.njk
```diff
@@ -44,10 +44,8 @@
                     {% set tabItems = [] %}
                     {% for section in entry.contentSections -%}
                         {% set tabContent -%}
-                            <div class="content-box content-box--frameless">
```
mattandrews: These two classes are already added in the `contentTabs` macro, so we were doubling up on spacing, eg:

broken

davidrapson: 💯

PR #1411, opened 2018-10-31
Rename grant URLs

mattandrews:
This looks heftier than it is(!).

Renames grant search URLs to the following structure:

- `/funding/grants` => search landing page (now with case studies and links to other search tools)
- `/funding/grants/$id` => grant detail page
- `/funding/grants/recipients/$id` => recipient detail page (with possibility to make a recipients landing page one day)

Also redirected some stuff:
- `/funding/past-grants` => redirects to `/funding/grants`
- `/funding/past-grants/search` => redirects to `/funding/grants`
- `/funding/search-past-grants-alpha` => redirects to `/funding/grants`

I was also able to generate the filter summary (in JS only, but could do on the server if we wanted) on pageload, which wasn't very hard in the end – really just involved mapping over the few parameters which don't share values/labels (eg. `10000%7C50000` versus `£10,000–£50,000`) which I'm pretty pleased with.

Finally, I added some links to return to searches when users click through from results pages:

image
image

This works by forwarding all the search query parameters along to the grant/recipient detail pages, along with a `?from=search` param which has to exist. This feels a bit ugly as we're dumping a bunch of otherwise-unrelated parameters onto all of those pages, but I can't think of another way to do it other than something like `window.history.back()` which feels messier. Thoughts welcome!


config/default.json
```diff
@@ -50,5 +50,11 @@
   },
   "googleAnalyticsCode": "UA-98908627-1",
   "legacyDomain": "https://wwwlegacy.biglotteryfund.org.uk",
-  "ebulletinApiEndpoint": "https://apiconnector.com/v2"
+  "ebulletinApiEndpoint": "https://apiconnector.com/v2",
+  "grantData": {
+    "dateRange": {
+      "start": "1 April 2004",
+      "end": "12 September 2018"
```
mattandrews: We'll need to update this whenever we add a new dataset, basically.

assets/js/vue-apps/past-grants.js
```diff
@@ -45,14 +45,47 @@ function init() {
             // @ts-ignore
             const initialData = window._PAST_GRANTS_SEARCH;
             const initialQueryParams = get(initialData, 'queryParams', {}) || {};
+            const facets = get(initialData, 'facets', {});
+
+            // Reconstruct the summary by grabbing labels from facets
+            // (eg. for the subset of filters which have different labels
+            // to their URL-provided values
+            const initialFilterSummary = () => {
```
davidrapson: This if in this function is always true right? If so you might be able use `lodash/map` over the initialQuery params stuff rather than pushing to a new array. Could possibly be something like this:

```suggestion
const initialFilterSummary = map(initialQueryParams, (value, key) => {
    let label;
    switch (key) {
        case 'amount':
            label = get(facets, 'amountAwarded[0].label');
            break;
        case 'awardDate':
            label = get(facets, 'awardDate[0].label');
            break;
        case 'localAuthority':
            label = get(facets, 'localAuthorities[0].label');
            break;
        case 'westminsterConstituency':
            label = get(facets, 'westminsterConstituencies[0].label');
            break;
    }

    return {
        name: key,
        label: label || value
    };
});
```

mattandrews: Good call!

PR #1412, opened 2018-11-01
Past grants i18n

mattandrews:
Adds a whole load of Welsh translations. Hopefully this is all of it – the rest are over in https://github.com/biglotteryfund/grants-service/pull/36. 

Possibly some of this is redundant (eg. things we changed after requesting translations) but it's fiddly to say until we switch over to the new search.

Once this (and the corresponding API one) are live on the `TEST` env I'll send it to Dave for some final sanity checks, given it's quite hard for us to tell if it's right.


mattandrews:
Managed to mess up the merge/base branch so go here instead: https://github.com/biglotteryfund/blf-alpha/pull/1413

PR #1423, opened 2018-11-02
Past grants improvements

mattandrews:
Adds a dictionary lookup with suggested spellings (both in English and Welsh!):

typo

You'll also see that triggering a new search from the `noResults` screen also shows the loading message, to make it clearer something has happened.

Also fixes this bug that crept back in: 
summary

(eg. clicking the summary item for the `q` filter didn't remove it).

This also involved me publishing my first `npm` package with the Welsh dictionary – hoping my open sourcery here is all above-board.

welsh


davidrapson:
Did you mean…

Looks good to me

PR #1427, opened 2018-11-05
Extract related grants into a separate endpoint

davidrapson:
Avoids needing to reference class names in the controller or pass conditionals around.

### DIFF
--- a/assets/js/vue-apps/grant-detail.js
+++ b/assets/js/vue-apps/grant-detail.js
@@ -1,5 +1,6 @@
-import $ from 'jquery';
 import Vue from 'vue';
+import $ from 'jquery';
+import find from 'lodash/find';
 
 function init() {
     const mountEl = document.getElementById('js-grant-related-projects');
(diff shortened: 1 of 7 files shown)


assets/js/vue-apps/grant-detail.js
```diff
@@ -8,27 +9,26 @@ function init() {
...
                 grant: PGS.project,
-                relatedGrants: false
+                relatedGrants: null
             };
         },
         mounted: function() {
             let geocode;
             if (this.grant.beneficiaryLocation) {
-                let loc = this.grant.beneficiaryLocation.find(l => l.geoCodeType === 'CMLAD');
+                let loc = find(this.grant.beneficiaryLocation, location => location.geoCodeType === 'CMLAD');
```
mattandrews: Out of interest, why load in a module to do this instead of using the native version?

controllers/grants/search.js
```diff
@@ -161,13 +161,11 @@ router.get(
...
-        const searchQueryString = querystring.stringify(
-            Object.assign({}, queryWithPage, {
-                from: 'search'
-            })
-        );
+        /**
+         * Add a parameter so we know the user came from search
+         * so we can link them back to their results.
+         */
+        const searchQueryString = querystring.stringify({ ...queryWithPage, ...{ from: 'search' } });
```
mattandrews: Nice. The spread syntax always confuses me, I can't parse it.

assets/js/vue-apps/grant-detail.js
davidrapson: Because IE 11 doesn't support `Array.find`

PR #1433, opened 2018-11-09
Fix disabled merch item

mattandrews:
Had a weird thing where if we disabled a single product within a merch item (eg. a bilingual banner), the frontend still tried to render it, resulting in a broken image:

image

This fix prevents these items from rendering, so we can disable things on a per-product basis (rather than removing the entire merch item).

### DIFF
--- a/controllers/materials/views/materials.njk
+++ b/controllers/materials/views/materials.njk
@@ -103,9 +103,12 @@
 {% macro listProducts(languageToShow) %}
     <ul class="flex-grid flex-grid--3up">
         {% for material in materials %}
-            <li class="flex-grid__item">
-                {{ materialCard(material, languageToShow, loop.index) }}
-            </li>
+            {% set productCount = material.products | filter("language", languageToShow) | length %}
+            {% if productCount > 0 %}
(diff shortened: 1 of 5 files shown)


controllers/materials/views/materials.njk
```diff
@@ -103,9 +103,12 @@
 {% macro listProducts(languageToShow) %}
     <ul class="flex-grid flex-grid--3up">
         {% for material in materials %}
-            <li class="flex-grid__item">
-                {{ materialCard(material, languageToShow, loop.index) }}
-            </li>
+            {% set productCount = material.products | filter("language", languageToShow) | length %}
```
davidrapson: Might it be better to do this filtering up front in the controller?

mattandrews: Not sure it's possible – the controller already knows there isn't a bilingual product (eg. it's not in the data), the problem comes down to the template trying to render both a mono/bilingual product because it's just assuming (not checking) there's one there.

davidrapson: Ah, OK, must've missed that. Crack on.

PR #1435, opened 2018-11-12
More Vue grant components

mattandrews:
This refactors things a little so the related grants widget is a fully-fledged Vue component, and converts the the "back to search results" button into a Vue component with state stored for an hour in `localStorage`. Was going to use `sessionStorage` but this isn't shared across tabs which may be a use case for this search.

### DIFF
--- a/assets/js/helpers/storage.js
+++ b/assets/js/helpers/storage.js
@@ -22,3 +22,33 @@ export function storageAvailable(type) {
         );
     }
 }
+
+// Based on https://gist.github.com/anhang/1096149
+export function setWithExpiry({ type, key, data, expiryInMinutes }) {
+    if (!storageAvailable) {
+        return;
+    }
+    const expirationMs = expiryInMinutes * 60 * 1000;
+    const record = {
+        value: JSON.stringify(data),
+        timestamp: new Date().getTime() + expirationMs
+    };
+    window[type].setItem(key, JSON.stringify(record));
+    return data;
+}
+
+export function getWithExpiry({ type, key }) {
+    if (!storageAvailable || !window[type].getItem(key)) {
+        return;
+    }
+    const record = JSON.parse(window[type].getItem(key));
+    if (!record) {
+        return false;
+    }
+    const hasExpired = new Date().getTime() > record.timestamp;
+    if (hasExpired) {
+        window[type].removeItem(key);
+        return;
+    }
+    return JSON.parse(record.value);
+}
--- a/assets/js/helpers/storage.js
+++ b/assets/js/helpers/storage.js
@@ -25,7 +25,7 @@ export function storageAvailable(type) {
 
 // Based on https://gist.github.com/anhang/1096149
 export function setWithExpiry({ type, key, data, expiryInMinutes }) {
-    if (!storageAvailable) {
+    if (!storageAvailable(type)) {
         return;
     }
     const expirationMs = expiryInMinutes * 60 * 1000;
@@ -38,7 +38,7 @@ export function setWithExpiry({ type, key, data, expiryInMinutes }) {
 }
 
 export function getWithExpiry({ type, key }) {
-    if (!storageAvailable || !window[type].getItem(key)) {
+    if (!storageAvailable(type) || !window[type].getItem(key)) {
         return;
     }
     const record = JSON.parse(window[type].getItem(key));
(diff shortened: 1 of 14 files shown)


assets/js/vue-apps/components/grants-back-to-search.vue
```diff
@@ -0,0 +1,46 @@
...
+export default {
+    components: { IconArrowLeft },
+    props: ['label', 'storageKey', 'prefix'],
+    data() {
+        return {
+            returnLink: undefined
+        };
+    },
+    computed: {
+        id() {
```
davidrapson: Don't worry about this for now, but I reckon we should push this unique id logic into the icon component (happy to pick that up)

assets/js/helpers/storage.js
```diff
@@ -22,3 +22,33 @@ export function storageAvailable(type) {
...
+    const record = {
+        value: JSON.stringify(data),
+        timestamp: new Date().getTime() + expirationMs
+    };
+    window[type].setItem(key, JSON.stringify(record));
+    return data;
+}
+
+export function getWithExpiry({ type, key }) {
+    if (!storageAvailable || !window[type].getItem(key)) {
```
davidrapson: Should `storageAvailable` be a function call?

assets/js/vue-apps/components/grants-back-to-search.vue
```diff
@@ -0,0 +1,46 @@
...
+            return Math.random()
+                .toString(36)
+                .substr(2, 9);
+        }
+    },
+    mounted: function() {
+        if (canStore) {
+            const searchData = getWithExpiry({ type: 'localStorage', key: this.storageKey });
```
davidrapson: As `getWithExpiry` references `storageAvailable` do we need the explicit `canStore` here?

assets/js/helpers/storage.js
mattandrews: Oops, yes!

assets/js/vue-apps/components/grants-back-to-search.vue
```diff
@@ -0,0 +1,46 @@
...
+            if (searchData) {
+                this.returnLink = this.prefix + '?' + searchData;
+            }
+        }
+    }
+};
+</script>
+
+<template>
+    <p class="u-no-margin"
```
davidrapson: Rather than a `p` with a no margin class this could be a plan ol' div right?

assets/js/vue-apps/past-grants.js
```diff
@@ -179,6 +183,21 @@ function init() {
                 this.filterResults();
             },
 
+            storeSearchPath(filters) {
+                if (Object.keys(filters).length === 0) {
+                    return canStore && window.localStorage.removeItem(STORAGE_KEY);
+                }
+                return (
+                    canStore &&
```
davidrapson: As above, do we need the `canStore` here?

controllers/grants/views/recipient-detail.njk
```diff
@@ -4,122 +4,126 @@
...
-                                            {% if grant.isActive %}
-                                                <span class="u-pull-right active-badge">{{ copy.grantDetail.statuses.activeProject }}</span>
-                                            {% endif %}
-                                        </dd>
-                                    {% endfor %}
-                                </dl>
-                            </div>
+            <div class="u-inner-wide-only">
+                <h2>
+                    {{ __('funding.pastGrants.search.recipientDetail.numFundedTotal', totalResults, totalAwarded) }}
```
davidrapson: Whilst we're tweaking this, any chance you could make this use `__n` to pluralise it, i.e. 10 projects, rather than 10 project(s)

controllers/grants/views/recipient-detail.njk
```diff
@@ -4,122 +4,126 @@
...
-                                        <dd>
-                                            <a href="{{ pgLink('programme', programme.title) }}">{{ programme.title }}</a>
-                                            {% if grant.isActive %}
-                                                <span class="u-pull-right active-badge">{{ copy.grantDetail.statuses.activeProject }}</span>
-                                            {% endif %}
-                                        </dd>
-                                    {% endfor %}
-                                </dl>
-                            </div>
+            <div class="u-inner-wide-only">
```
davidrapson: Think this should be `u-inner` otherwise there's no padding on small screens which it needs here.

mattandrews:
Thanks – can you take another look when you get a sec?

PR #1446, opened 2018-11-14
Add updates controller

mattandrews:
Adds a very basic frontend for the new "updates" content type:

image

image

image

Very much copied from the existing blog controllers/views. Few things to discuss based off this (see comments!).


controllers/updates/index.js
```diff
@@ -0,0 +1,125 @@
...
+            });
+
+            if (response.result) {
+                // Do we have a listing page?
+                if (isArray(response.result)) {
+                    // @TODO i18n for 'All news' and pageTitles below
+                    // Is this a specific listing page (eg. blog, press releases) or everything?
+                    const filterType = get(response, 'meta.pageType');
+                    const isNewsLandingPage = !req.params.type && filterType === 'single';
+                    // @TODO these titles are singular - should they be pluralised here manually or in the CMS?
```
mattandrews: Really not sure how best to handle this – we could map the types here (guess we'll need to translate them) or translate them in Craft?

controllers/updates/views/listing.njk
```diff
@@ -0,0 +1,82 @@
...
+                {% elseif entriesMeta.pageType === 'category' %}
+                    {% set activeCategory = entriesMeta.activeCategory %}
+                    <h1>Viewing all posts in {{ activeCategory.title }}</h1>
+                {% else %}
+                    <h1>{{ title }}</h1>
+                {% endif %}
+
+                {% if isNewsLandingPage %}
+                    {% for type, entrylist in entries | groupItemsBy('updateType.name') %}
+                        {# @TODO find a better way to handle this #}
```
mattandrews: Possibly we could do this replacement work at the Craft end so only the API needs to worry about the `_`/`-` replacement?

controllers/updates/views/listing.njk
```diff
@@ -0,0 +1,82 @@
...
+{% extends "layouts/main.njk" %}
+
+{% block content %}
+    {% set pageAccent = 'cyan' %}
+
+    <main role="main" id="content">
+        <section class="content-box u-inner-wide-only accent--{{ pageAccent }} a--border-top">
+            {{ breadcrumbTrail(breadcrumbs) }}
+
+            <div class="{% if not isNewsLandingPage %}u-constrained-wide{% endif %}">
```
davidrapson: Not a big fan of this `isNewsLandingPage` flag, would rather render two separate templates. I think they are distinct enough to warrant it.

controllers/updates/index.js
```diff
@@ -0,0 +1,125 @@
...
+        const isSinglePost = req.params.date && req.params.slug;
+
+        if (req.params.type) {
+            // Due to quirks of Craft's entry types,
+            // we have to convert snake_case into kebab-case
+            // both at the API end and then in reverse here.
+            urlPath += `/${req.params.type.replace('-', '_')}`;
+        }
+
+        if (isSinglePost) {
```
davidrapson: Is there a way of rewriting this so the URL path code is one block of logic and the query is another? Would be nice to collect related logic together.

controllers/updates/index.js
```diff
@@ -0,0 +1,125 @@
...
+                if (isArray(response.result)) {
+                    // @TODO i18n for 'All news' and pageTitles below
+                    // Is this a specific listing page (eg. blog, press releases) or everything?
+                    const filterType = get(response, 'meta.pageType');
+                    const isNewsLandingPage = !req.params.type && filterType === 'single';
+                    // @TODO these titles are singular - should they be pluralised here manually or in the CMS?
+                    let pageTitle = isNewsLandingPage ? 'All news' : response.result[0].updateType.name;
+                    let filterLabel;
+
+                    if (filterType === 'author') {
```
davidrapson: Wondering if we could isolate the blocks of logic a bit more here, something like:

```js
// @TODO these titles are singular - should they be pluralised here manually or in the CMS?
const defaultPageTitle = isNewsLandingPage ? 'All news' : response.result[0].updateType.name;

const filterLabel = {
    'author': get(response, 'meta.activeAuthor'),
    'tag': get(response, 'meta.activeTag'),
    'category': get(response, 'meta.activeCategory'),
}[filterType];

const pageTitle = {
    'author': `Author: ${filterLabel.title}`,
    'tag': `Tag: ${filterLabel.title}`,
    'category': `Category: ${filterLabel.title}`,
}[filterType] || defaultPageTitle;

const breadcrumbs = concat(res.locals.breadcrumbs, {
    label: filterLabel ? filterLabel.title : pageTitle
});
```

controllers/updates/index.js
mattandrews: Yeah, have simplified this now.

controllers/updates/index.js
mattandrews: Yeah agreed – have added singular/plural translations so we can do this sort of thing:

image

image

controllers/updates/index.js
mattandrews: Good shout, thanks for the code.

controllers/updates/views/listing.njk
mattandrews: Done (see the other PR!)

mattandrews:
@davidrapson just pushed changes, could you give this another quick look?

davidrapson:
Lovely work 🚢 -it

PR #1449, opened 2018-11-14
Tidy up news listings

davidrapson:
Does a bit of work to group by type at the controller level so we can decide what order to list things on the landing page. Also does some work to tidy up our use of macros, making the `article-teaser` the main one. Fixes up a few spacing issues at the same time.

image


controllers/updates/views/landing.njk
```diff
@@ -1,46 +1,66 @@
...
+            {{ newsSection(
+                title = copy.types.blog.plural,
+                entries = groupedEntries.blog,
+                allLink = localify('/news/blog')
+            )}}
+
+            {# Press releases #}
+            {{ newsSection(
+                title = copy.types['press-releases'].plural,
+                entries = groupedEntries.press_releases,
```
davidrapson: I think this will need to swap to `groupedEntries['press-releases'] once I've updated my version of the API.

PR #1450, opened 2018-11-15
Add debugging and more defensive code to grants search

davidrapson:
Makes a couple of changes to make the grants code a little more error proof and to add some more debugging in.

### DIFF
--- a/controllers/grants/index.js
+++ b/controllers/grants/index.js
@@ -87,13 +87,6 @@ function buildAllowedParams(queryParams) {
     return pick(queryParams, allowedParams);
 }
 
-/**
- * Append a page number parameter to a map of query parameters
- */
-function addPaginationParameters(existingParams, pageNumber = 1) {
-    return Object.assign({}, existingParams, { page: pageNumber });
-}
-
 // Query the API with a set of parameters and return a promise
 async function queryGrantsApi(parameters) {
     return request({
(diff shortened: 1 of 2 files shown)


controllers/grants/index.js
```diff
@@ -150,85 +143,82 @@ async function checkSpelling(searchTerm, locale = 'en') {
...
 router.get('/', injectHeroImage('tinylife'), injectCopy('funding.pastGrants.search'), async (req, res, next) => {
-    let data;
-    const facetParams = buildAllowedParams(req.query);
-    const paginationLabels = req.i18n.__('global.misc.pagination');
-    let queryWithPage = addPaginationParameters(facetParams, req.query.page);
-    queryWithPage.locale = res.locals.locale;
-
     try {
-        data = await queryGrantsApi(queryWithPage);
+        const locale = req.i18n.getLocale();
```
davidrapson: Inlined this code all into the try/catch block. Not 100% how the async behaviour acts if we have one path of code in the try/catch and the other outside it. Moves this all into the main block.

controllers/grants/index.js
```diff
@@ -150,85 +143,82 @@ async function checkSpelling(searchTerm, locale = 'en') {
...
+                });
+            }
+        });
     } catch (errorResponse) {
         return res.format({
             html: () => {
                 next(errorResponse.error);
             },
             'application/json': () => {
+                Raven.captureMessage(errorResponse);
```
davidrapson: This nested `error.error` logic is still a bit funny so logging out the raw error to see if we can make this more robust.

controllers/grants/index.js
```diff
@@ -269,20 +259,23 @@ router.get('/related', injectCopy('funding.pastGrants.search'), async (req, res,
...
-            qs: qs
+            qs: query
         });
 
         const paginationLabels = req.i18n.__('global.misc.pagination');
 
-        if (data && data.meta.totalResults > 0) {
-            const firstResult = head(data.results);
-            const organisation = head(firstResult.recipientOrganization);
+        const organisation = get(data, 'results[0].recipientOrganization[0]');
```
davidrapson: Using `get` for the whole thing (turns out you can pluck array items out using this) to avoid any red herrings from trying to get the organisation.

controllers/grants/views/recipient-detail.njk
```diff
@@ -109,12 +109,13 @@
...
                                             <dd>
                                                 <a href="{{ pgLink('programme', programme.title) }}">{{ programme.title }}</a>
-                                                {% if grant.isActive %}
-                                                    <span class="u-pull-right active-badge">{{ copy.grantDetail.statuses.activeProject }}</span>
-                                                {% endif %}
                                             </dd>
                                         {% endfor %}
                                     </dl>
+
+                                    {% if grant.isActive %}
```
davidrapson: Fixes a small layout bug where the active flag was overlaying text weirdly.

PR #1452, opened 2018-11-16
Grants cache

davidrapson:
Extends the cache window for past grants to 7 days. Also updates the cloudfront config to allow opting out of having the session cookie in the cache key for certain routes. Opts-out only for grants urls.


modules/cloudfront.js
```diff
@@ -13,9 +13,8 @@ const CLOUDFRONT_PATHS = [
     { path: '*~/link.aspx', isPostable: true, allowAllQueryStrings: true },
     { path: '/api/*', isPostable: true, allowAllQueryStrings: true },
     { path: '/funding/funding-finder', isPostable: true, allowAllQueryStrings: true, isBilingual: true },
-    { path: '/funding/grants', isPostable: true, allowAllQueryStrings: true, isBilingual: true },
-    { path: '/funding/grants/*', allowAllQueryStrings: true, isBilingual: true },
-    { path: '/funding/grants/recipients/*', allowAllQueryStrings: true, isBilingual: true },
+    { path: '/funding/grants', isPostable: true, allowAllQueryStrings: true, isBilingual: true, noSession: true },
+    { path: '/funding/grants/*', allowAllQueryStrings: true, isBilingual: true, noSession: true },
```
mattandrews: Can we double check this on TEST? I think I had to manually fix this (eg. just use `/funding/grants*`) because it didn't add both those routes or something, and the `/related` endpoint was always cached (for Welsh, which I forgot to change until this morning).

PR #1493, opened 2018-11-27
Make sure fontfaceobserver still works with new typefaces

davidrapson:
Make sure fontfaceobserver still works with new typefaces

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -4,5 +4,12 @@ const cssnano = require('cssnano');
 const foutWithAClass = require('postcss-fout-with-a-class').default;
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


postcss.config.js
```diff
@@ -4,5 +4,12 @@ const cssnano = require('cssnano');
 const foutWithAClass = require('postcss-fout-with-a-class').default;
 
 module.exports = {
-    plugins: [autoprefixer(), foutWithAClass({ families: ['Poppins', 'Roboto'], className: 'fonts-loaded' }), cssnano()]
+    plugins: [
+        autoprefixer(),
+        foutWithAClass({
+            families: ['Poppins', 'Roboto', 'caecilia', 'caecilia-sans-text'],
```
davidrapson: Checked and this list works for both stylesheets, i.e. it's OK to have a single list here.

PR #1496, opened 2018-11-29
Press releases

davidrapson:
WIP in progress press-releases templates


controllers/updates/views/post/press-release.njk
```diff
@@ -0,0 +1,72 @@
...
+{% extends "layouts/main.njk" %}
+
+{% macro articleMeta(entry) %}
+    <div class="article-meta">
+        {% if entry.authors.length > 0 %}
+            <p class="article-meta__author">
+                By
+                {% set separator = joiner(',') %}
+                {% for author in entry.authors -%}
+                {{ separator() }} <a href="{{ localify('/news?author=' + author.slug) }}">{{ author.title }}</a>
```
mattandrews: Probably an open question about whether we want authors for press releases, but I guess the field isn't required so if nobody uses it, this won't show up.

davidrapson: I've removed this from the template (sorry this PR was a little pre-emptive so was never meant to stay in here)

PR #1497, opened 2018-11-29
Add SRI integrity checks to third party (versioned) modules

mattandrews:
Just read [this](https://shkspr.mobi/blog/2018/11/major-sites-running-unauthenticated-javascript-on-their-payment-pages/) and realised we had a few things we could fix this for.

The remaining stuff where we can't add these parameters are:

- `https://www.google-analytics.com/cx/api.js` – presumably safe, since it's Google
- `https://www.google-analytics.com/analytics.js` – ditto
- `https://platform.twitter.com/widgets.js` – possibly going soon
- `https://cdn.polyfill.io<snip>` – impossible to SRI...

### DIFF
--- a/controllers/tools/index.js
+++ b/controllers/tools/index.js
@@ -18,13 +18,7 @@ router.use(
     noCache,
     noindex,
     buildSecurityMiddleware({
-        defaultSrc: [
-            "'self'",
-            'maxcdn.bootstrapcdn.com',
-            'ajax.googleapis.com',
-            'cdnjs.cloudflare.com',
-            'cdn.jsdelivr.net'
-        ]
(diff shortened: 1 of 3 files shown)


davidrapson:
👍 Excellent work.

For https://www.google-analytics.com/cx/api.js this is the experiments API right? I think we could comfortably remove this. At least until such time as we decide A/B test are worth it.

Analytics, yeah we have to make that trade-off

For `polyfill.io`, if we want to be particularly cautious it's possible to download it with [&flags=always,gated](https://cdn.polyfill.io/v2/polyfill.min.js?features=HTMLPictureElement,Promise,Array.prototype.find,Element.prototype.closest&flags=always,gated) and self-host the file. But that has the trade-off of loading polyfills where they are not needed

mattandrews:
Let's revisit the A/B experiments stuff post-rebrand, just in case. We can do a whole bunch of removal then!

PR #1503, opened 2018-11-30
Rework rebrand flags to allow switching copy in locals

davidrapson:
Rework rebrand flags to allow switching copy in locals. Allows us to switch out global rebrand copy at a global level rather than needing to differ in templates.


middleware/locals.js
```diff
@@ -4,120 +4,120 @@ const moment = require('moment');
...
+const features = config.get('features');
 
 /**
  * Set request locals
  * - Local properties that depend on the request
  * - Local methods for use in views that depend on the request
  */
-module.exports = {
-    // Export for tests
-    getMetaTitle,
```
davidrapson: Removing the test for this and inlining the function as we test the output of this more thoroughly as part of the end-to-end tests.

middleware/locals.js
```diff
@@ -4,120 +4,120 @@ const moment = require('moment');
...
+    /**
+     * Rebrand flag
+     */
+    const useNewBrand = appData.isNotProduction && req.cookies[cookies.rebrand] === REBRAND_SECRET;
+    res.locals.useNewBrand = useNewBrand;
+
+    /**
+     * Global copy
+     */
+    const globalCopy = {
```
davidrapson: Replaced all references to `{{ __('global.brand.*') }}` in the templates with this local, allows us to wholesale swap out the copy, will look at doing this for `global.contact` too once the update branch has been merged in (as it's been changed there).

PR #1473, opened 2018-11-26
Streamline domain config

davidrapson:
Begins work to streamline references to the live domain ahead of the domain move.

PR #1506, opened 2018-12-03
Domain prep

davidrapson:
Re-do of https://github.com/biglotteryfund/blf-alpha/pull/1473

Fewer config changes at once here, but the biggest change is to us JS syntax for the config files to allow us to include comments and TODOs noting which bits need to change as part of the domain move.


config/default.js
```diff
@@ -0,0 +1,87 @@
...
+     */
+    domains: {
+        // Which domain is the primary/canonical domain
+        // @TODO: Update when moving primary domain
+        primary: 'www.biglotteryfund.org.uk',
+        // Which domains should be indexable by search engines
+        // @TODO: Add new domain when moving primary domain
+        indexable: ['www.biglotteryfund.org.uk'],
+        // Explicit config for new domain
+        new: 'www.tnlcommunityfund.org.uk'
```
mattandrews: Could add a `@TODO` here to remove this post-rebrand, too.

config/default.js
```diff
@@ -0,0 +1,87 @@
...
+            cy: '/welsh'
+        }
+    },
+    features: {
+        enableDigitalFundApplications: true,
+        enablePrompt: false,
+        enableSurvey: true,
+        enableAbTests: true,
+        enableHotjar: false,
+        useRemoteAssets: true,
```
mattandrews: ... wonder if we should rename this to `enableRemoteAssets` for the sake of consistency/OCD.

PR #1509, opened 2018-12-03
Redirect old blogpost URLs to the new structure

mattandrews:
Can go live when we're ready.

### DIFF
--- a/controllers/blog/index.js
+++ b/controllers/blog/index.js
@@ -18,78 +18,91 @@ function renderListing({ res, title, entries = [], entriesMeta = null, breadcrum
     });
 }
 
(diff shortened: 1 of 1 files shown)


davidrapson:
Given how few posts and urls there are for the current blog I'd personally prefer adding a static list of everything in `aliases` rather than needing any custom express handlers.

mattandrews:
Fair enough – done!

davidrapson:
What's the simplest number of steps we could take to get this live? Thinking we could redirect `/news` to `/news/blog` for the time-being, but what do we need to do to switch the blog over? Is it a line on introductory copy?

PR #1510, opened 2018-12-03
Add S3 origin to Cloudfront

mattandrews:
This configures `/media/*` (for CMS files) and `/assets/*` to serve from S3 directly via Cloudfront, so we can move off `media.biglotteryfund.org.uk` (which we won't have a certificate for post-rebrand).

### DIFF
--- a/modules/filters.js
+++ b/modules/filters.js
@@ -17,11 +17,9 @@ try {
     assets = JSON.parse(fs.readFileSync(path.join(__dirname, '../config/assets.json'), 'utf8'));
 } catch (e) {} // eslint-disable-line no-empty
 
-function getCachebustedPath(urlPath, useRemoteAssets = config.get('features.useRemoteAssets')) {
-    const version = assets.version || 'latest';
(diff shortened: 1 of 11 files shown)


mattandrews:
Have tested this on the `TEST` Cloudfront env and it's working as expected.

modules/filters.js
```diff
@@ -19,7 +19,7 @@ try {
 
 function getCachebustedPath(urlPath, useRemoteAssets = config.get('features.useRemoteAssets')) {
     const version = assets.version || 'latest';
-    const baseUrl = useRemoteAssets ? 'https://media.biglotteryfund.org.uk/assets' : `/assets`;
```
davidrapson: If we're going to change this here we don't need the domain name anymore, or the `useRemoteAssets` flag, this can return `/assets` all the time.

mattandrews: Ah I thought we still needed/used this for the email templates. Just removed it!

PR #1514, opened 2018-12-04
New global navigation

davidrapson:
Updates new header based on latest designs


middleware/locals.js
```diff
@@ -57,6 +58,44 @@ module.exports = function(req, res, next) {
         };
     });
 
+    /**
+     * Global navigation model
+     */
+    res.locals.globalNavigation = {
```
davidrapson: Given how one-off the global navigation is I've gone with the approach of explicitly modelling the structure here rather than computing it from the routes file. There's probably a way to simplify some of the boilerplate here, but I think the concept makes sense.

mattandrews: You could possibly simplify this a bit with something like this:

```
const { isString } = require('lodash');

// Nav elements here are either strings
// (eg. sharing the same key for translations as the URL path
// or objects (eg. specifying a custom label/path if they differ)
const navElements = {
    primary: [
        'funding',
        {
            label: 'updates',
            path: '/news'
        },
        'research',
        'contact'
    ],
    secondary: [
      'about',
      'jobs'
    ]
};

const createNavItem = navItem => {
    if (isString(navItem)) {
        return {
            label: req.i18n.__(`global.nav.${navItem}`),
            url: localify(locale)(`/${navItem}`)
        }
    } else {
        return {
            label: req.i18n.__(`global.nav.${navItem.label}`),
            url: localify(locale)(${navItem.path})
        }
    }
};

res.locals.globalNavigation = {
    home: {
        label: req.i18n.__('global.nav.home'),
        url: localify(locale)('/')
    },
    primaryLinks: navElements.primary.map(createNavItem),
    secondaryLinks: navElements.secondary.map(createNavItem)
};
```

(saves on the repeated calls to `localify()` and `req.i18n.__()` boilerplate – though might be a bit too magic?)

davidrapson: This feels too magic for me, I think it's good here to be as explicit as possible even if that means a little boilerplate. I've just reduced the `req.i18n.__` noise with a variable though.

Issue #1512, opened 2018-12-04
Migrate BBO pages


mattandrews:
The BBO pages are complex and will need attention.

- [ ] Ensure links on the [landing page](https://www.biglotteryfund.org.uk/funding/programmes/building-better-opportunities) point to migrated content
- [ ] Create location [subpages](https://www.biglotteryfund.org.uk/funding/programmes/building-better-opportunities#section-2) for each of the regions (eg. like [this one](https://www.biglotteryfund.org.uk/global-content/programmes/england/building-better-opportunities/cheshire-and-warrington))
- [ ] Migrate/redirect the [Case Studies](https://www.biglotteryfund.org.uk/global-content/programmes/england/building-better-opportunities/building-better-opportunities-case-studies)
- [ ] Migrate all of the files listed on the [Resources page](https://www.biglotteryfund.org.uk/funding/programmes/building-better-opportunities/building-better-opportunities-resources)
- [ ] Migrate all of the files listed on the [Guide](https://www.biglotteryfund.org.uk/funding/programmes/building-better-opportunities/guide-to-delivering-european-funding) page

PR #1515, opened 2018-12-05
Add BBO subpages

mattandrews:
Makes BBO pages a wildcard route meaning we can do some of the things in https://github.com/biglotteryfund/blf-alpha/issues/1512.

### DIFF
--- a/controllers/programmes/index.js
+++ b/controllers/programmes/index.js
@@ -3,6 +3,8 @@ const path = require('path');
 const { concat, map, groupBy } = require('lodash');
 const express = require('express');
 
+const { basicContent } = require('../common');
+
 const {
     injectBreadcrumbs,
     injectCopy,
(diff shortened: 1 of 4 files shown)


controllers/programmes/index.js
```diff
@@ -134,4 +136,16 @@ router.get('/:slug', injectFundingProgramme, (req, res, next) => {
     }
 });
 
+router.use(
```
davidrapson: Thinking about it it might be better to do something like:

```suggestion
router.use(
    '/building-better-opportunities/*',
    (req, res, next) => {
            if (customAncestors) {
                res.locals.customAncestors = customAncestors;
            }
            next();
        },
    basicContent()
);
```

And then we could remove the middleware from `basicContent`

mattandrews: Good call, have a look now

PR #1520, opened 2018-12-06
Wildcard aliases

mattandrews:
This is to help with [migrating BBO](https://github.com/biglotteryfund/blf-alpha/issues/1512) – eg. to not have to specify 38 URLs which in turn ends up outputting ~300 aliases due to all the country/language permutations.

This allows a `from`/`to` pair to include a wildcard which gets replaced at request time with the requested path after the wildcard, eg:

```
{
    from: '/foo/bar/*',
    to: '/some/endpoint/*'
}
```

... would mean the following URLs would redirect like so:

```
/foo/bar/lol => /some/endpoint/lol
/foo/bar/purple/monkey/dishwasher => /some/endpoint/purple/monkey/dishwasher
// etc
```

Tested locally and works as expected – I think it's a bit neater than specifying all of them, and will make the migration work a bit simpler post-rebrand.

Main risk is some unknown page we don't know about which will get caught up in this – we can double check analytics to see if there's anything else going on for this section (my research suggests we've done all of these pages now).


server.js
```diff
@@ -185,8 +185,25 @@ app.use('/patterns', require('./controllers/pattern-library'));
...
  */
+
+const removeWildcard = urlPath => urlPath.replace('/*', '');
+
 aliases.forEach(redirect => {
-    app.get(redirect.from, (req, res) => res.redirect(301, redirect.to));
+    // Allow wildcard redirects to avoid specifying a whole directory of links
+    // Currently we only support paths that end with a wildcard (eg. /foo/bar/*)
+    if (redirect.from.match(/\/\*$/)) {
+        const pathsMinusWildcards = {
```
davidrapson: Could this be inlined `const wildcardPath = removeWildcard(redirect.to) + req.path.replace(removeWildcard(redirect.from), '');`

server.js
```diff
@@ -185,8 +185,25 @@ app.use('/patterns', require('./controllers/pattern-library'));
 /**
  * Handle Aliases
  */
+
+const removeWildcard = urlPath => urlPath.replace('/*', '');
```
davidrapson: Even though this is tiny mind popping this in `modules/urls`

mattandrews:
@davidrapson could you give this a fresh look? I've basically removed the wildcard option after looking into the other URLs this would catch – there are a bunch of other pages in there (FAQs etc) which would get caught by this and unfortunately the URL aliases are registered before the archived URLs so I can't put them there. It's easier right now to just put the entire list in here and be explicit (and remove them eventually when traffic drops off).

This also now includes tweaks to the listing page to allow for a simplified list view, eg:

image

This view will kick in if any of the children in the listing view don't have an associated photo (and pairs with https://github.com/biglotteryfund/craft-dev/pull/120).

mattandrews:
Argh, commits didn't push properly

PR #1525, opened 2018-12-11
Remove lint staged

davidrapson:
Seen cases where pre-commit hooks via lint-staged and nodemon (presumably any file-watcher? Like `webpack --watch`) join forces to cause 💻 🔥

This is not an exact diagnosis of the cause but the recent stash/apply behaviour in lint-staged is an obvious candidate for something that could interfere with file watching behaviour. We don't get enough value from lint-staged relative to the potential loss of work-in-progress so this PR removes it.

The one thing we loose is auto-magic prettier formatting but I've added a `format` command similar to our other repos which formats on a one-off basis. It's possible to configure editors to prettier stuff for you on save too. But I'd rather this be a manual thing then have as much magic going on when committing.


davidrapson:
"I like what it does, just not to this extreme...". Yeah, definitely!

PR #1528, opened 2018-12-11
Fix news landing

mattandrews:
Spotted an issue where a lot of new posts in one category over another means there's a chance one category will be empty, eg:

image

This is also partly due to a bug in https://github.com/biglotteryfund/blf-alpha/pull/1527 which changes the format (eg. we now need to use `.result` for the entries, though I can't work out why we didn't need this before...).

### DIFF
--- a/controllers/updates/index.js
+++ b/controllers/updates/index.js
@@ -22,13 +22,22 @@ if (appData.isNotProduction) {
         async (req, res, next) => {
             try {
                 const { copy, breadcrumbs } = res.locals;
-                const response = await contentApi.getUpdates({
-                    locale: req.i18n.getLocale()
+
+                const pressReleases = await contentApi.getUpdates({
+                    locale: req.i18n.getLocale(),
+                    type: 'press-releases'
                 });
+
+                const blogposts = await contentApi.getUpdates({
+                    locale: req.i18n.getLocale(),
+                    type: 'blog'
+                });
+
+                const allEntries = concat(pressReleases.result, blogposts.result);
 
                 res.render(path.resolve(__dirname, `./views/landing`), {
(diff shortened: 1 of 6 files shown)


mattandrews:
Oh, and this also adds related programmes to press releases, which almost always have one (it turns out):

image

controllers/updates/index.js
```diff
@@ -22,13 +22,22 @@ if (appData.isNotProduction) {
...
+                    locale: req.i18n.getLocale(),
+                    type: 'blog'
+                });
+
+                const allEntries = concat(pressReleases.result, blogposts.result);
+
                 res.render(path.resolve(__dirname, `./views/landing`), {
                     title: copy.title,
-                    groupedEntries: groupBy(response.result, 'entryType'),
+                    groupedEntries: groupBy(allEntries, 'entryType'),
```
davidrapson: Do you need this groupBy anymore. Given you are concatenating the already grouped entries then ungrouping/grouping them again? Could you pass through the two separate results as they are?

controllers/updates/views/view-helpers.njk
```diff
@@ -76,3 +76,26 @@
 
     {% endcall %}
 {% endmacro %}
+
+
+{% macro relatedProgrammes(entry) %}
```
davidrapson: Maybe make this into a first-class component in `components/`?

controllers/updates/index.js
mattandrews: Haha, great point. Will fix!

controllers/updates/views/view-helpers.njk
mattandrews: Done.

PR #1530, opened 2018-12-12
Insights section

davidrapson:
This PR prepares the new Insights section to go live.  Moves all new pages to `/insights` leaving `/research` for legacy content which we can eventually blanket archive.

Remaining steps:

- [x] Publish third research article (pending minor design update and small bit of translation)

image

This is dependant on us launching one more research article—which will happen within the next few days—and to get the locale text in this branch translated.


services/content-api.js
```diff
@@ -172,15 +173,15 @@ function getFundingProgramme({ locale, slug, previewMode = false }) {
...
-function getResearch({ locale, slug = null, searchQuery = null, previewMode = null }) {
+function getResearch({ locale, slug = null, previewMode = null }) {
     if (slug) {
         return fetch(`/v1/${locale}/research/${slug}`, {
             qs: addPreviewParams(previewMode)
-        }).then(response => get('data.attributes')(response));
+        }).then(getAttrs);
     } else {
-        const baseUrl = `/v1/${locale}/research`;
-        const url = searchQuery ? `${baseUrl}?q=${searchQuery}` : baseUrl;
```
davidrapson: The search query parameter wasn't being used at all, we _may_ re-add this when we have enough insights articles to warrant it but at the current rate of publishing that's not going to be a problem for a while.

davidrapson:
Noticed in the screenshot that the entries appear out of order. This can be disregarded. It's sorted by post date by the "youth employment" entry doesn't have a post date yet so is being sorted last.

davidrapson:
Yep 😄 

> This is dependant on us launching one more research article—which will happen within the next few days—and to get the locale text in this branch translated.

davidrapson:
image

Now with translations

PR #1547, opened 2018-12-13
Add grant case studies

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/127

Allows us to add content to a case study which will then appear on its associated grant detail page:

image

### DIFF
--- a/assets/sass/scaffolding/_content.scss
+++ b/assets/sass/scaffolding/_content.scss
@@ -90,6 +90,11 @@
         @include mq('medium-major') {
             flex: 0 0 320px;
         }
+
+        // Prevent sidebar from stretching to match primary content height
+        &.content-sidebar__secondary--minimal {
+           align-self: flex-start;
+        }
     }
 }
 
--- a/config/default.json
+++ b/config/default.json
@@ -40,7 +40,7 @@
     "enableHotjar": false,
     "enableTimingMetrics": false,
     "enableMailSendMetrics": false,
-    "enableRelatedGrants": false
+    "enableRelatedGrants": true
   },
   "imgix": {
     "mediaDomain": "biglotteryfund-assets.imgix.net"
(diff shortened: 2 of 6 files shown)


assets/sass/scaffolding/_content.scss
```diff
@@ -90,6 +90,11 @@
         @include mq('medium-major') {
             flex: 0 0 320px;
         }
+
+        // Prevent sidebar from stretching to match primary content height
+        &.content-sidebar__secondary--minimal {
```
mattandrews: Without this it looks... wrong:

image

config/default.json
```diff
@@ -40,7 +40,7 @@
     "enableHotjar": false,
     "enableTimingMetrics": false,
     "enableMailSendMetrics": false,
-    "enableRelatedGrants": false
+    "enableRelatedGrants": true
```
mattandrews: Think we can safely re-enable this now?

assets/sass/scaffolding/_content.scss
davidrapson: Rather than doing this could we take the manual tone classes off `content-sidebar__secondary` and make that grant summary thing use the card classes (or basically move the classes down a level). Should do the same thing without one-off styles.

config/default.json
davidrapson: Worth a try

controllers/grants/index.js
```diff
@@ -265,6 +265,15 @@ router.get('/:id', injectCopy('funding.pastGrants.search'), async (req, res, nex
     try {
         const data = await grantsService.getById({ id: req.params.id, locale: req.i18n.getLocale() });
 
+        let caseStudy;
```
davidrapson: Not sure how I feel about this `setHeroLocals` function. Could this whole thing be inside an `injectCaseStudy` function?

assets/sass/scaffolding/_content.scss
mattandrews: Done!

PR #1564, opened 2018-12-14
New programme card layout

davidrapson:
One of the proposed changes as part of the rebrand is to make programme cards look like this.

image

I'd rather not maintain multiple templates for this for current and rebranded styles so have updated these across the board. 💭 Thoughts welcome.


davidrapson:
@mattandrews Mind giving this another look? I realised the new programme card style is the same as promo cards so I've removed the concept of programme cards in favour of using the promo card across the board.

PR #1567, opened 2018-12-14
Initial version of name change message

davidrapson:
image

We need to agree the wording and then get it translated but getting something in place for now.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -33,14 +33,15 @@
     }
   },
   "features": {
-    "enableDigitalFundApplications": false,
-    "enablePrompt": false,
-    "enableSurvey": true,
     "enableAbTests": true,
+    "enableDigitalFundApplications": false,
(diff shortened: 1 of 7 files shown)


config/default.json
```diff
@@ -33,14 +33,15 @@
...
-    "enableDigitalFundApplications": false,
-    "enablePrompt": false,
-    "enableSurvey": true,
     "enableAbTests": true,
+    "enableDigitalFundApplications": false,
     "enableHotjar": false,
-    "enableTimingMetrics": false,
     "enableMailSendMetrics": false,
-    "enableRelatedGrants": true
+    "enableNameChangeMessage": true,
```
davidrapson: `enableNameChangeMessage` is the new flag, the rest is me sorting the list

davidrapson:
I've disabled this in production for now so this is safe to merge.

PR #1569, opened 2018-12-14
grants search e2e test

davidrapson:
On the back of https://github.com/biglotteryfund/grants-service/pull/50 I've added a basic e2e test for the grants search.

Performs a search query for a term which returns <100 results so we can also run a smoke-test for pagination using relative total counts.

image

This test will need to be updated if any future grants match this term but as it's narrow enough and we only import monthly that shouldn't mean too much churn

🍰


davidrapson:
@mattandrews Just tweaked this to use the filters for this query too if you don't mind giving it another check.

mattandrews:
Nice – :shipit:

PR #1570, opened 2018-12-17
Archive legacy pages and files

### LINKED ISSUE #1552: Fix issue with archived pages showing archived pages (#INCEPTION)

mattandrews:
This extracts the National Archives code out into a separate file as I suspect we'll be doing more of this (with more complex logic) in the near future. Also fixes #1552 by tying NA links to a known-working archive.

It also adds a handler for static files (along with a corresponding Cloudfront path which I've verified on `TEST`) which shows this message:

image

This feedback form shows up in the tools like so:

image

(eg. with an appended file path so we can see what was requested)

Finally, these requests are counted in Cloudwatch along with a slightly shonky version of the file path, so we can see if particular files are being requested more than others. We can also see this in Analytics but thought it was worth tracking this separately here: 

image


assets/js/vue-apps/components/feedback.vue
```diff
@@ -11,12 +11,17 @@ export default {
     },
     methods: {
         handleSubmit() {
+            let message = this.feedback;
```
davidrapson: Might be more vue-ish to make this a `computed` property?

modules/cloudfront.js
```diff
@@ -22,11 +22,21 @@ const CLOUDFRONT_PATHS = [
     { path: '/user/*', isPostable: true, queryStrings: ['redirectUrl', 's', 'token'] }
 ];
 
+if (config.get('features.enableLegacyFileArchiving')) {
```
davidrapson: Not 100% sure this flag makes sense here. This will always be `true` as this is enabled in development and the cloudfront configs are only ever generated in development. Not sure what to suggest as an alternative approach though.

package.json
```diff
@@ -28,7 +28,7 @@
     "build": "npm run clean && npm run build-css && npm run build-js",
     "build-production": "export NODE_ENV=production && npm run build && ./bin/hash-rev",
     "build-autotrack": "npx autotrack -o public/autotrack/autotrack.js -p outboundLinkTracker,eventTracker",
-    "watch-css": "watch 'npm run build-css' assets/sass",
+    "watch-css": "watch 'npm run build-css' assets/sass --interval=0.1",
```
davidrapson: ?

controllers/archived.js
```diff
@@ -0,0 +1,59 @@
...
+ * and b) to allow us to record the redirect as a pageview using standard analytics behaviour.
+ */
+// prettier-ignore
+flatMap([
+    '/about-big/10-big-lottery-fund-facts',
+    '/funding/funding-guidance/applying-for-funding/*',
+    '/global-content/programmes/england/building-better-opportunities/building-better-opportunities-qa*'
+], urlPath => [urlPath, makeWelsh(urlPath)]).forEach(urlPath => {
+    router.get(urlPath, cached.noCache, function(req, res) {
+        const fullUrl = `https://${config.get('siteDomain')}${req.originalUrl}`;
```
davidrapson: This `siteDomain` variable will probably change to be the new one soon, but for archive purposes we always want this to be biglotteryfund.org.uk so it would be good to hard-coded this here.

davidrapson: (I know you're just moving this code)

package.json
mattandrews: I was seeing a really slow (like, 2-3 seconds) delay on seeing file changes here. [This](https://github.com/mikeal/watch/pull/84#issuecomment-193960026) helped a bit, so adding it here while I remembered.

controllers/archived.js
mattandrews: Yeah good call, will tweak.

package.json
davidrapson: Oh wow didn't realise the default was 5 seconds! (according to that link), go for it

mattandrews:
@davidrapson happy to give this another look?

controllers/archived.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+});
+
+/**
+ * Archived files
+ * Requests for legacy files (eg. stored on Sitecore) will be shown an error message
+ * along with a feedback form to explain what they were looking for.
+ * We also log all requests for these files to ensure we can update anything missing.
+ */
+/*
```
davidrapson: Got some bonus comments here

PR #1574, opened 2018-12-17
Rebrand refinement

davidrapson:
Doing one more pass through our components list to bring them in line with our proposed designs.

I'm keen to avoid too many conditionals in our templates to support the rebrand so these changes may result in some minor cosmetic differences to the current live designs. I'll call out where those are so we can make a call on if the changes are acceptable.


davidrapson:
Removing border from basic card macro

image

davidrapson:
Removing border from media figures

image

davidrapson:
Made segment images flush

image

davidrapson:
Removed border from miniature heroes

image

davidrapson:
Removed colour themes from CMS powered pages (after conversation with @LynseyJReynolds )

image

views/includes/metadata.njk
```diff
@@ -37,7 +37,7 @@
 <meta name="apple-mobile-web-app-title" content="{{ globalCopy.brand.title }}">
 <meta name="application-name" content="{{ globalCopy.brand.title }}">
 
-{% set metaThemeColour = "#ec008c"  %}
+{% set metaThemeColour = "#e5007d" if useNewBrand else "#e20087"  %}
```
mattandrews: Good spot 👍

PR #1581, opened 2018-12-20
All funding programmes

davidrapson:
This PR reinstates the all programmes list after pairing with @LynseyJReynolds 

Dependencies:

- [x] https://github.com/biglotteryfund/craft-dev/pull/134 needs to be deployed first
- [x] New UI text needs translating

image

The main things to note are:

- This shows *all* programmes in alpha-numeric order on a single page (the API call is little intensive but this page is low-access and can have a long cache window)
- We made the call that pending further research into needs the simplest approach we can take is to show title and description to allow us to show everything on a browsable page.
- Includes the ability to filter by region like you can with the main list
- I've not reinstated the code for the detail pages yet but this is behind a flag so we can merge this as a  work in progress


controllers/programmes/views/all-programmes.njk
```diff
@@ -0,0 +1,55 @@
...
+            image = pageHero.image,
+            accent = pageAccent
+        ) }}
+
+        <section class="content-box u-inner-wide-only accent--{{ pageAccent }} a--border-top nudge-up">
+            {{ breadcrumbTrail(breadcrumbs) }}
+
+            <nav class="alpha-links">
+                <ol class="alpha-links__list">
+                    {% for letter in "#ABCDEFGHIJKLMNOPQRSTUVWXYZ" | list %}
```
mattandrews: Just going to leave this here: https://wales.com/welsh-alphabet :trollface:

mattandrews: (ignore this, just thought it was funny)

davidrapson:
> Looks great. Worth flagging expired programmes, or should we do that as part of the followup (eg. when we decide how the expired programmes themselves should look?). 👍

Discussed this with @LynseyJReynolds and we're going to leave this until we do the wider bit of work around expired programmes. Ideally we'd want to flag truly "open" programmes rather than archived as 90% of what is on this page is archived/expired.

controllers/programmes/views/archived-programme.njk
```diff
@@ -0,0 +1,39 @@
+{% extends "layouts/main.njk" %}
+{% from "components/content-box-intro/macro.njk" import contentBoxIntro with context %}
+{% from "components/content-box/macro.njk" import contentBox %}
+{% from "components/page-title/macro.njk" import pageTitle %}
+{% from "components/programmes.njk" import programmeStats with context %}
+{% from "components/badges/macros.njk" import badgeArchived with context %}
+
+{#  Add archived prefix to help with analytics reporting #}
+{% block title %}Archived | {{ title }} | {% endblock %}
```
mattandrews: Should "Archived" be in the translations file?

PR #1585, opened 2018-12-21
Simplified hero

davidrapson:
Last PR of the year…

This is a little prep I realised we'll need to support the new hero images in the new year.

- The current hero image is hard-coded to fixed aspect ratios. But with new images we are likely to have a couple different sizes (shallow for listing pages and deeper for content pages) so it's easier to drop the fixed aspect ratios and let the img element do it's thing.
- A fair amount of extra logic is needed to support pages with no hero images

By removing the page title from the hero image and treating it as an `<aside>` (these heroes are content but not essential content) we can simplify the markup necessary to support pages with and without hero images.

This means a little churn whilst we migrate over to the new macro, but I've attempted to do this in such a way that we can gradually migrate (although post-Christmas clear-brained David might disagree with this) we'll need to review all the pages as we update hero images anyway.

### DIFF
--- a/assets/sass/_vars-rebrand.scss
+++ b/assets/sass/_vars-rebrand.scss
@@ -26,7 +26,7 @@ $constrained: 700px;
 $constrainedWide: 840px;
 
 // Nudge offset, used to nudge content areas above hero images
-$nudgeOffset: 60px;
+$nudgeOffset: 125px;
 
 /* =========================================================================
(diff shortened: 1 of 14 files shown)


davidrapson:
> although post-Christmas clear-brained David might disagree with this

Post-Christmas David disagrees with making this branch of the all funding programmes one…so going to make a clean version of this PR.

PR #1588, opened 2019-01-03
Update e2e tests for new brand

davidrapson:
As we get closer to launching the new brand styles I think it makes sense to run our integration tests against the new styles. Not sure if it warrants running the tests twice against each version, currently only running in new-brand mode. Thoughts welcome 💭


davidrapson:
This should fail until https://github.com/biglotteryfund/blf-alpha/pull/1587 is merged

davidrapson:
Will fail until https://github.com/biglotteryfund/blf-alpha/pull/1587 is merged in

davidrapson:
FWIW 90% of the tests still apply regardless of the brand, the main change is the new header. So we'd no longer be e2e testing that but I agree, we can live with that.

PR #1594, opened 2019-01-07
Update name change message and enable in production

davidrapson:
- [ ] Agree on message wording, update date
- [ ] Agree on where this links to, currently linking to the original press-release
- [ ] Arrange welsh translation


davidrapson:
Waiting on welsh translation (only the message, the linked press-release is already translated) and then this can go live.

Issue #1598, opened 2019-01-08
Redirect old research section to National Archives


davidrapson:
Redirect old research section to National Archives in favour of the new Insights section.

Determine if it's possible to do this with a wildcard redirect (e.g. all pages under /research/*) or if we need a more granular approach.

PR #1617, opened 2019-01-09
Archive research section

davidrapson:
Archives old research pages with the exception of a handful of redirects.

The top-level `/research` url redirects to `/insights` and the three new page redirect to their new home under insights. Plus I've made the following link go explicitly to the national archives.

image

I've also taken this as an opportunity to clean up references to research in the code.

### DIFF
--- a/views/components/promo-card/macro.njk
+++ b/views/components/promo-card/macro.njk
@@ -35,7 +35,8 @@ featured: {Boolean} - Use featured style for promo card
                     {% endcall %}
                 </p>
             {% endif %}
-            <h3 class="promo-card__title t4">
+            {% set level = props.headingLevel | default(3, true) %}
+            <h{{ level }} class="promo-card__title t4">
                 {% if props.link.url %}
                     <a
                         class="u-link-minimal"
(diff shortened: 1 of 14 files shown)


davidrapson:
@mattandrews Thoughts on `630ef81` in particular. Ended up making a change to the promo-card to support setting a custom heading level, otherwise the Insights page fails the a11y check.

views/components/promo-card/macro.njk
```diff
@@ -35,7 +35,8 @@ featured: {Boolean} - Use featured style for promo card
                     {% endcall %}
                 </p>
             {% endif %}
-            <h3 class="promo-card__title t4">
+            {% set level = props.headingLevel | default(3, true) %}
```
mattandrews: Worth putting a sanity check in here to avoid it being higher than 6? Not that I don't trust us...

davidrapson: Good call.
```js
if (commitAuthor = 'davidrapson' && headingLevel > 6) {
  throw new Error('Nope')
}
```

davidrapson:
Closes #1598

PR #1622, opened 2019-01-09
Redirect remaining pages in our approach section

davidrapson:
This is pending a decision on translation for international and non-lottery funding guidance pages. Once these have been published we can redirect remainder of the our approach section.


davidrapson:
After discussion have launched the two untranslated pages as is and will send to be translated as soon as possible, but we can launch as is.

PR #1624, opened 2019-01-10
Enable file archive

mattandrews:
This makes the archived file behaviour the default. We'll definitely see requests come in for these files but the ones at the top of the list have now been migrated (monitoring forms). 

We'll likely see referrals from Google rather than the website, which might mean we should add aliases for highly-requested files – we can monitor the logs to see what people are looking for.


modules/cloudfront.js
```diff
@@ -162,6 +162,7 @@ function generateBehaviours(origins, originName) {
      * or custom cookies to be whitelisted you must define those rules here.
      */
     let customPaths = [
+        { path: '/-/media/files/*', isPostable: false, allowAllQueryStrings: true },
```
davidrapson: Not 100% sure why we need this route in here? Also it's likely going to clash with the `/-/*` path rule above which sends anything under that to the legacy servers.

https://github.com/biglotteryfund/blf-alpha/blob/b412017bc281c0136f672f3d9f4f7e9855b3de2f/modules/cloudfront.js#L123-L124

mattandrews: This is where the original code was adding this path before my change...? I just took out the `if` statement.

mattandrews: This is how it looks on TEST currently, based on the same config:

image

mattandrews: ... so yeah, you're right, the first rule takes precedence. Will have a fiddle.

PR #1658, opened 2019-01-21
Update mail service to support new domain

davidrapson:
Update mail service to support new domain so that internal emails sent to the new domain get sent from the internal address.

### DIFF
--- a/services/mail.js
+++ b/services/mail.js
@@ -72,8 +72,10 @@ function generateHtmlEmail({ template, templateData }) {
  * @return {string}
  */
 function getSendAddress(recipients) {
-    const addressess = recipients.map(recipient => recipient.address);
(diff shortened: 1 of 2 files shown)


services/mail.js
```diff
@@ -73,7 +73,9 @@ function generateHtmlEmail({ template, templateData }) {
  */
 function getSendAddress(recipients) {
     const addressess = recipients.map(recipient => recipient.address);
-    if (addressess.some(address => /@biglotteryfund.org.uk$/.test(address))) {
+    if (
+        addressess.some(address => /@biglotteryfund.org.uk$/.test(address) || /@tnlcommunityfund.org.uk$/.test(address))
```
mattandrews: I know it was already wrong but should we rename `addressess` to `addresses`?

PR #1661, opened 2019-01-22
 Alternate approach to digital assistance sign-up

davidrapson:
Follow up to https://github.com/biglotteryfund/blf-alpha/pull/1659

Updates the digital fund assistance sign-up so that it sends an internal email rather than signing up to a soon to be non-existent list.


davidrapson:
I've checked with the digital fund team they are happy with this approach, once they've confirmed they've got a test email this can be merged.

PR #1665, opened 2019-01-22
Tweak the import date in readiness for new grant dataset

mattandrews:
Minor change but means when I run the import script later, the dates will be correct.


mattandrews:
> 👍 Maybe add a ticket to move this to the API response for next time. We have to deploy the API anyway to release a new dataset so this should really come from there as a `meta` field. Could still be hard-coded.

Done!

PR #1671, opened 2019-01-23
Fixup page title spacing

davidrapson:
Fixes this issue by making sure the page-title styles match the content box / constrained styles. 

image


davidrapson:
@mattandrews Realised we can use the `u-padded u-inner-wide-only` directly so have removed the custom styles.

PR #1672, opened 2019-01-23
Allow sending email from the new domain when switched on

mattandrews:
Once we switch over to the new domain, we can start sending emails from it too.

### DIFF
--- a/services/mail.js
+++ b/services/mail.js
@@ -48,10 +48,10 @@ function generateHtmlEmail({ template, templateData }) {
                 juice.juiceResources(
                     html,
                     { removeStyleTags: false, webResources: { relativeTo: publicRoot } },
-                    function(juceErr, newHtml) {
-                        /* istanbul ignore if  */
-                        if (juceErr) {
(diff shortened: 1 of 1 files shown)


services/mail.js
```diff
@@ -48,10 +48,10 @@ function generateHtmlEmail({ template, templateData }) {
                 juice.juiceResources(
                     html,
                     { removeStyleTags: false, webResources: { relativeTo: publicRoot } },
-                    function(juceErr, newHtml) {
-                        /* istanbul ignore if  */
-                        if (juceErr) {
-                            reject(juceErr);
+                    function(juiceErr, newHtml) {
```
mattandrews: (the typo was annoying me)

mattandrews:
Have tested this locally and can confirm it works:

image

image

services/mail.js
davidrapson: I leave thme jsut fro you

davidrapson: (thanks for fixing this though)

PR #1673, opened 2019-01-23
Archive press releases

mattandrews:
Redirects `/news-and-events` to `/news` and sends old press releases to the archives.

One thing we could consider doing is redirecting old press releases themselves to the archives, eg. [this](https://www.biglotteryfund.org.uk/global-content/press-releases/wales/060818_wal_a-dream-come-true-for-margaret-and-sarah) to [here](https://webarchive.nationalarchives.gov.uk/20181001135409tf_/https://www.biglotteryfund.org.uk/global-content/press-releases/wales/060818_wal_a-dream-come-true-for-margaret-and-sarah) – worth considering?

### DIFF
--- a/modules/archived.js
+++ b/modules/archived.js
@@ -2,9 +2,9 @@
 
 const LAST_GOOD_CRAWL_DATE = '20171011152352';
 
-function buildArchiveUrl(urlPath) {
+function buildArchiveUrl(urlPath, crawlDate = LAST_GOOD_CRAWL_DATE) {
     const fullUrl = `https://www.biglotteryfund.org.uk${urlPath}`;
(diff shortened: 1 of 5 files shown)


modules/archived.js
```diff
@@ -2,9 +2,9 @@
 
 const LAST_GOOD_CRAWL_DATE = '20171011152352';
 
-function buildArchiveUrl(urlPath) {
+function buildArchiveUrl(urlPath, crawlDate = LAST_GOOD_CRAWL_DATE) {
```
davidrapson: I think it's worth inlining this as `crawlDate = '20171011152352'`, one less bit of indirection.

PR #1674, opened 2019-01-23
Fix user controller regressions

mattandrews:
These issues came up when I was testing the email sending changes: it looks like when we introduced staff login to the site, I forgot to update a few of the references to `req.user.*`, which should have become `req.user.userData.*` – eg. so we could use the top-level object to distinguish between user types.

This issue meant that user activation didn't work properly as the code always regarded the user as inactive. Luckily this is disabled in production, but still worth fixing.

Finally there was a weird edge case bug where a deleted user who was logged in would only see a 500 error as Passport tried to deserialise an account that no longer existed. Admittedly this is unlikely to happen in reality, but the fix was simply returning `null` in a callback.

### DIFF
--- a/middleware/passport.js
+++ b/middleware/passport.js
@@ -80,6 +80,9 @@ module.exports = function() {
     }
 
     const makeUserObject = user => {
+        if (!user) {
+            return null;
+        }
         return {
             userType: user.constructor.name,
             userData: user
--- a/controllers/user/activate.js
+++ b/controllers/user/activate.js
@@ -43,10 +43,11 @@ function activate(token, user) {
 
 router.route('/').get(requireUserAuth, async (req, res) => {
     const token = req.query.token;
+    const user = req.user.userData;
 
     if (token) {
         try {
-            await activate(token, req.user);
+            await activate(token, user);
             res.redirect('/user');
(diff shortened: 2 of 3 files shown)


middleware/passport.js
```diff
@@ -80,6 +80,9 @@ module.exports = function() {
     }
 
     const makeUserObject = user => {
+        if (!user) {
+            return null;
```
mattandrews: We need this because the `user` object we're working with comes from `findById()`, which returns `null` if the user doesn't exist (eg. doesn't trigger any of our `catch()` blocks).

mattandrews: ... otherwise, this threw an actual (uncaught) error and the user sees an unstyled 500 page for every URL on the site.

davidrapson: Mind adding this rationale as a comment to the code?

davidrapson:
This is a great catch by the way. Is there anything we could do to have allowed tests to catch this? Or does that involve bypassing / mocking staff-auth to do that?

middleware/passport.js
mattandrews: Done!

PR #1676, opened 2019-01-23
Add translations for home page

davidrapson:
Welsh translation for new intro text and links.

### DIFF
--- a/controllers/home/views/home-rebrand.njk
+++ b/controllers/home/views/home-rebrand.njk
@@ -52,7 +52,6 @@
             <ul class="flex-grid flex-grid--3up">
                 {% for item in copy.featuredLinks %}
                     <li class="flex-grid__item">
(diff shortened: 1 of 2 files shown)


controllers/home/views/home-rebrand.njk
```diff
@@ -52,7 +52,6 @@
             <ul class="flex-grid flex-grid--3up">
                 {% for item in copy.featuredLinks %}
                     <li class="flex-grid__item">
-                        {# @TODO: Fetch real image path #}
```
davidrapson: Removing this TODO as these images are "real" but they are _hardcoded_. In that if we change the images for the pages they link to we need to sync them up manually. This is fine for now but I've raised a ticket to look at this in future (#1677). Possibly updating the hero-images API endpoint to allow fetching multiple images in one request.

PR #1682, opened 2019-01-24
Add a new favicon for new brand

mattandrews:
Yes, this is really the new favicon.

The only one I didn't replace here was the root `/favicon.ico` because we need a `req` object to determine whether we show the rebrand, and I didn't think it was worth adding a fully-fledged middleware to handle that (eg. on deployment day we can just swap that file out). The others will be shown if the rebrand is switched on.


mattandrews:
image

🤯

PR #1685, opened 2019-01-25
Design tweaks

davidrapson:
Paired with @LynseyJReynolds on a few refinements and corrections to the new design.

- Tweaks logo size to be the same at all breakpoints; within thresholds and allows for the needed minimum spacing around the logo
- Tweak colours on the data page to use new colour scheme
- Update toggle link colours on small screens
- Updates the hero caption to align to the top of the page to avoid clashing with page titles (we'd like to revisit image captions more widely soon, but this avoids the immediate issue)
- Update contact page image on the homepage
- Update some hero image slugs which have been changed in the CMS


views/components/hero.njk
```diff
@@ -26,5 +26,15 @@
                 </span>
             {% endif %}
         </div>
+
+        {% if image.caption and  useNewBrand %}
```
davidrapson: This code is duplicated from above as the markup order needs to change which is not ideal but we can clean this up fairly swiftly post rebrand.

PR #1688, opened 2019-01-25
Update brand copy

mattandrews:
Replaces a bunch of brand names / links / email addresses across the site using the same script as the [CMS replacer](https://gist.github.com/mattandrews/80bfa770d2c5d43fdb1a70f731394491).

If you want to run this yourself, put this bash script in the root folder and run it.

```
#!/bin/bash
shopt -s nullglob

for f in $(find .  -type f \( -name '*.js' -o -name '*.njk' -o -name '*.yml' -o -name '*.json' \) -not -path "*node_modules/*" -not -path "*idea/*" -not -path "*/config/cloudfront/*")
do
    echo "Replacing brand data in file - $f"
    OUTPUT="$(cat $f | ./renamer)";
    echo "$OUTPUT" > $f;
done

```


davidrapson:
(Branch is failing on an `aXe` test, may be the branch isn't up to date with `master`?)

PR #1714, opened 2019-01-30
Theme cleanup

davidrapson:
Now that the rebrand is out we no longer have the concept of "page accents".

This PR removes any references to `pageAccent` variables and `a--` accent classes. I'll call out any notable changes in comments in the PR as there's a lot going on here.


assets/sass/components/_news.scss
```diff
@@ -116,29 +116,6 @@
     margin-bottom: 4px;
 }
 
-/* =========================================================================
```
davidrapson: This wasn't used anywhere in the code 🗑

assets/sass/scaffolding/_lists.scss
```diff
@@ -47,28 +47,3 @@
         @include wrap-words();
     }
 }
-
-/* =========================================================================
```
davidrapson: This wasn't used anywhere in the code 🗑

assets/sass/utilities/_utilities-accents.scss
```diff
@@ -1,14 +1,9 @@
...
    Utilities: Tones & Accents
    ========================================================================= */
-// Utility classes to help with colour accent theming
-
-/* =========================================================================
-   Tones
-   ========================================================================= */
 
 /* Foreground tones */
-.u-tone-background-brand-primary {
```
davidrapson: The class name included **background** but it changed the **foreground** colour. 🤦‍♂️

assets/sass/utilities/_utilities-accents.scss
```diff
@@ -27,108 +22,3 @@
 .u-border-top-brand-primary {
     border-top: 4px solid get-color('brand', 'primary');
 }
-
-.u-border-minor {
```
davidrapson: This is gone, I replaced the one use case with a `<hr/>` which is a semantically correct thing to use here so we can avoid the extra styles.

controllers/common/views/information-page.njk
```diff
@@ -8,46 +8,23 @@
...
-                image = pageHero.image,
-                accent = pageAccent
-            ) }}
+            {{ hero(title, pageHero.image) }}
         {% endif %}
 
         {% if pageHero.image %}<div class="nudge-up">{% endif %}
 
         {% block contentPrimary %}
-            {% if content.relatedContent %}
```
davidrapson: By looking at the Field Inventory plugin in Craft I noticed that a) this `relatedContent` field is only included on BBO pages an b) it's never filled in. I _think_ this is a hold over from an old version of the people pages. Basically this can go and the field can be removed from the API.

controllers/errors/views/error.njk
```diff
@@ -1,39 +1,33 @@
...
-            image = fallbackHeroImage,
-            accent = pageAccent
-        ) }}
+        {{ hero(title, fallbackHeroImage) }}
 
         <div class="nudge-up">
-            <div class="content-box u-inner-wide-only accent--{{ pageAccent }} a--border-top">
-                <div class="s-prose">
-                    <h3 class="t2">Error {{ status }}</h3>
+            {% call contentBox() %}
```
davidrapson: This file looks churn-y but it's only wrapping the content in a `contentBox` macro.

views/static-pages/research.njk
```diff
@@ -1,29 +0,0 @@
-{% from "components/hero.njk" import hero with context %}
```
davidrapson: This is the old research landing page—which is now archived—so this page can go 🗑

controllers/digital-fund/strand-2.js
```diff
@@ -10,11 +10,6 @@ const features = config.get('features');
 
 const router = express.Router();
 
-router.use((req, res, next) => {
-    res.locals.pageAccent = 'cyan';
```
davidrapson: These are going because `pageAccent` doesn't make sense any more in the new brand. But we will want to come back to this as and when we relaunch the digital fund. I know @LynseyJReynolds is keen to keep _some_ kind of distinction between the two strands.

controllers/common/views/information-page.njk
mattandrews: Excellent!

PR #1722, opened 2019-01-31
Template and trail image cleanup

davidrapson:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/142 which must go out before this can be deployed.

Updates references to old field names in templates, notably trail and hero images.

### DIFF
--- a/controllers/programmes/views/programmes-list.njk
+++ b/controllers/programmes/views/programmes-list.njk
@@ -36,8 +36,7 @@
                     <a href="{{ localify('/funding/programmes/all') }}" class="btn btn--medium btn--outline">{{ copy.allProgrammes }}</a>
                 </div>
             {% else %}
-                {# @TODO i18n #}
-                <p>No results</p>
+                <p>{{ __('global.misc.noResults') }}</p>
             {% endif %}
         </section>
     </main>
--- a/controllers/routes.js
+++ b/controllers/routes.js
@@ -15,7 +15,6 @@ const { basicContent, flexibleContent, staticPage } = require('./common');
  * @typedef {object} Section
  * @property {string} path
  * @property {boolean} showInNavigation
- * @property {string} [langTitlePath]
  * @property {Array<Route>} pages
  */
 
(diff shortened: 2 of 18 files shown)


controllers/routes.js
```diff
@@ -26,7 +25,6 @@ const { basicContent, flexibleContent, staticPage } = require('./common');
 const toplevel = {
     path: '',
     showInNavigation: true,
-    langTitlePath: 'global.nav.home',
```
davidrapson: We changed how the navigation is constructed with the new header and this isn't needed anymore.

controllers/routes.js
```diff
@@ -205,19 +199,13 @@ const updates = {
     ]
 };
 
-/**
- * Sections
- * The order here defines the order of the navigation
```
davidrapson: No longer true.

middleware/locals.js
```diff
@@ -80,14 +80,6 @@ module.exports = function(req, res, next) {
         caption: 'The Outdoor Partnership'
     };
 
-    res.locals.getSocialImageUrl = function(socialImage) {
```
davidrapson: The social images are now always an external link, so removing this logic

server.js
```diff
@@ -226,7 +226,6 @@ forEach(routes.sections, function(section, sectionId) {
      */
     router.use(function(req, res, next) {
         const locale = req.i18n.getLocale();
-        res.locals.sectionId = sectionId;
```
davidrapson: No longer used in templates.

davidrapson:
This will probably fail until the API update goes out.

controllers/programmes/views/programmes-list.njk
```diff
@@ -36,8 +36,7 @@
                     <a href="{{ localify('/funding/programmes/all') }}" class="btn btn--medium btn--outline">{{ copy.allProgrammes }}</a>
                 </div>
             {% else %}
-                {# @TODO i18n #}
-                <p>No results</p>
+                <p>{{ __('global.misc.noResults') }}</p>
```
mattandrews: 💯

server.js
mattandrews: From a simpler era.

Issue #1690, opened 2019-01-25
Block National Archives from archiving archives


mattandrews:

mattandrews:
Emailed them today to see how we can actually do this – their docs suggest `robots.txt` is the best way.

mattandrews:
Done in https://github.com/biglotteryfund/blf-alpha/pull/1727

PR #1727, opened 2019-02-04
Block archived paths

mattandrews:
Fixes https://github.com/biglotteryfund/blf-alpha/issues/1690

This prevents any page we no longer serve from appearing in Google (or indeed the National Archives), which should gradually reduce the occurrence of random old document links still showing up in results. This should also prevent circular redirects happening where we send people to NA pages which in turn redirect them... back to the NA.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -49,5 +49,31 @@
     "session": "blf-alpha-session"
   },
   "hotjarId": 1036292,
-  "googleAnalyticsCode": "UA-98908627-1"
+  "googleAnalyticsCode": "UA-98908627-1",
+  "archivedPaths": {
+    "legacyFilesPath": "/-/media/files/*",
+    "legacyPagePaths": [
+      "/about-big/10-big-lottery-fund-facts",
+      "/about-big/big-lottery-fund-in-your-constituency",
+      "/about-big/community-closed",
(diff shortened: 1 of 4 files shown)


config/default.json
```diff
@@ -49,5 +49,31 @@
     "session": "blf-alpha-session"
   },
   "hotjarId": 1036292,
-  "googleAnalyticsCode": "UA-98908627-1"
+  "googleAnalyticsCode": "UA-98908627-1",
+  "archivedPaths": {
```
davidrapson: Rather than living in the config could this all go in `modules/archived`?

controllers/robots/index.js
```diff
@@ -3,10 +3,12 @@ const express = require('express');
 const moment = require('moment');
 const sitemap = require('sitemap');
 const domains = require('config').get('domains');
-const { includes } = require('lodash');
+const legacyFilesPath = require('config').get('archivedPaths.legacyFilesPath');
```
davidrapson: As above, could this come from `modules/archived` like `legacyPagePaths` does?

config/default.json
mattandrews: Yeah good call – take a look now.

PR #1728, opened 2019-02-04
Print styles

davidrapson:
Fixes a handful of issues with print styles. This is largely down to the announcement banner, new header, and new cookie banner not having appropriate print styles. I've also added some print styles to `.content-box` to reduce the amount of UI clutter in the printed version.

screenshot 2019-02-04 at 10 21 20
screenshot 2019-02-04 at 10 20 48

I also cleaned up the brand copy so that we only have one version and removed some code supporting the old homepage (repeat visitor code)

### DIFF
--- a/assets/js/head.js
+++ b/assets/js/head.js
@@ -14,10 +14,6 @@ function getCommonBodyClasses() {
         classes.push('is-ie');
     }
 
-    if (AppConfig.hasSeenPageBefore) {
-        classes.push('is-repeat-visitor');
-    }
-
     return classes;
 }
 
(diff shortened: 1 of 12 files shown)


views/includes/global-header.njk
```diff
@@ -38,10 +38,10 @@
             <div class="global-header__content-primary">
                 <a href="/" class="global-header__logo brand-logo">
                     <span class="u-visually-hidden">
-                        {{ __("global.brand.title") }}
```
davidrapson: This was causing the descriptive text for the logo to still read "Big Lottery Fund"

assets/js/head.js
```diff
@@ -14,10 +14,6 @@ function getCommonBodyClasses() {
         classes.push('is-ie');
     }
 
-    if (AppConfig.hasSeenPageBefore) {
-        classes.push('is-repeat-visitor');
```
mattandrews: image

PR #1740, opened 2019-02-11
Remove timings middleware

davidrapson:
Now that we no longer proxy the legacy website we can remove our custom timings code. We can instead rely directly on the latency metrics on the ELB which monitor the same thing.


package.json
```diff
@@ -84,7 +84,6 @@
     "dictionary-en-gb": "2.0.0",
     "dompurify": "^1.0.3",
     "express": "^4.16.3",
-    "express-ab": "^0.7.1",
```
davidrapson: Removing this as we no longer load it.

davidrapson:
Percy test failed because the news stories on the homepage changed, which is not unexpected. If it becomes onerous it might be worth mocking API data for these tests.

PR #1744, opened 2019-02-12
Allow querystrings on document links for tracking downloads

mattandrews:
Fixes https://github.com/biglotteryfund/blf-alpha/issues/1741 (eg. clicks on Insights pages weren't being tracked because of a `?mtime=` query at the end of the link)

### DIFF
--- a/assets/js/bootstraps/analytics.js
+++ b/assets/js/bootstraps/analytics.js
@@ -4,7 +4,7 @@ import forEach from 'lodash/forEach';
 import { trackEvent } from '../helpers/metrics';
 
 function trackDocumentDownloads() {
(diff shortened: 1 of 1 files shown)


assets/js/bootstraps/analytics.js
```diff
@@ -6,7 +6,7 @@ import { trackEvent } from '../helpers/metrics';
 function trackDocumentDownloads() {
     const links = document.querySelectorAll('.content-box a[href]');
```
davidrapson: I'd be tempted to remove `.content-box` from here too, and look at all links…

mattandrews: Works for me!

PR #1758, opened 2019-02-15
Use a single API call for homepage

davidrapson:
We currently make two content API calls for the homepage. This reduces it to 1. See https://github.com/biglotteryfund/craft-dev/pull/151


davidrapson:
@mattandrews API endpoint is now live so this can go out now

PR #1759, opened 2019-02-15
npm updates

davidrapson:
A few npm updates.

- Fixes `npm audit` warnings.
- Bump express version (minor patch version)
- Update express-validator to latest version
- Update to latest nunjucks version
- Move client-side only dependencies to `devDependencies`, should reduce deploy artifact size and reduce the number of dependencies on instances

A few larger changes in here so needs a bit of extra testing.


davidrapson:
In retrospect there's a bit too much changing here, going to extract the `npm audit fix` change as a first step as that's the key thing to go out.

PR #1760, opened 2019-02-15
Prune client-side code

davidrapson:
- Move "legacy" form JS into materials module as it only relates to that form.
- Remove a feature flag which was always `true`
- Don't scope fitvids to `.video-container` class, we don't always know this will be applied as it comes from the CMS.


davidrapson:
(That would be nice, for now at least we can keep it where it belongs)

PR #1762, opened 2019-02-15
Press-release tweaks

davidrapson:
**Before**

screenshot 2019-02-15 at 14 33 59

**After**

screenshot 2019-02-15 at 14 38 28

### DIFF
--- a/assets/sass/utilities/_utilities-text.scss
+++ b/assets/sass/utilities/_utilities-text.scss
@@ -81,6 +81,10 @@
     @include wrap-words();
 }
 
+.u-wrap-links a {
+    @include wrap-words();
+}
+
 .unstyled,
 .unstyled li,
 .u-list-unstyled,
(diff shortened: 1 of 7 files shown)


assets/sass/utilities/_utilities-text.scss
```diff
@@ -81,6 +81,10 @@
     @include wrap-words();
 }
 
+.u-wrap-links a {
```
davidrapson: In the spirit of being intentional where we use this I've added a new utility class which opts in to wrapping links, we can use it on a case-by-case basis like the more general `u-wrap-words`. Trying to add these styles in more global places has proven too wide reaching.

davidrapson:
@mattandrews Spotted a couple of extra things if you don't mind checking this again.

Noticed the style of the message for "filtered with results" and "filtered with no results" was different, and the former didn't give you a way to reset the list, so I made these consistent.

screenshot 2019-02-15 at 15 00 32

screenshot 2019-02-15 at 15 03 17

In testing this I realised the pagination code doesn't account for other query parameters so the selected region was being lost if you tried to paginate. We currently have a slightly different function for this on the past grants search which _does_ account for this so I copied the approach from there and added some tests. There's probably a case to be made to combine this logic but as one method is for CMS API responses and the other is for grants API responses I _think_ it makes sense to keep them separate.

PR #1765, opened 2019-02-15
Deploy config update

davidrapson:
---

🚨🚨🚨 

This PR makes some changes to the environment name the staging site runs as. Currently we run the passenger app environment as `test`. As per https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_app_env this also sets the `NODE_ENV` variable.

`NODE_ENV=test` is a bit of a special case. For example `jest` sets this when running tests and this is the default environment for travis.

This ends up conflating the concept of `NODE_ENV=test` and a staging environment. The big change in this PR is to change the `NODE_ENV` value for the staging site to be `staging`.

🚨🚨🚨

---

The other changes in this PR are renaming the web root and debug config to use the new name.


appspec.yml
```diff
@@ -2,7 +2,7 @@ version: 0.0
 os: linux
 files:
   - source: /
-    destination: /var/www/biglotteryfund
+    destination: /var/www/tnlcommunityfund
```
mattandrews: 💯

davidrapson:
Going to re-raise this in a couple of parts next week (renaming the directory and _then_ the environment change)

PR #1768, opened 2019-02-20
Enhance funding prototype summary

mattandrews:
This does a few things:

- Refactors the translation aspect so we do it once at the very beginning of a request, merging in the session data where applicable

- Adds some breadcrumbs for navigation between pages and adds a "Skip to summary" shortcut button (probably not final but good for debugging)

- Adds styling and edit links to the summary screen

- Adds a basic start page

Probably can't go much further with this before we work on validation, eg. so we can validate an entire section/step and mark it as such.


controllers/apply-next/simple/form-model.js
```diff
@@ -9,6 +9,12 @@ const sectionProject = {
         en: 'Your Project',
         cy: '(WELSH) Your Project'
     },
+    introduction: {
```
mattandrews: This field is for the summary screen, eg:

image

PR #1769, opened 2019-02-20
Add a simple processor

mattandrews:
This is mostly a copy/paste job from previous processors, but it flattens the data out and constructs a table of data for an email. This will likely change a fair bit down the line but it means we can have an end-to-end application now with a success screen and email.


controllers/apply-next/form-router.js
```diff
@@ -217,20 +223,84 @@ function initFormRouter(formModel) {
...
+        const errorCopy = req.i18n.__('apply.error');
+        res.render(path.resolve(__dirname, './views/error'), {
+            error: error,
+            title: errorCopy.title,
+            errorCopy: errorCopy,
+            returnUrl: `${req.baseUrl}/review`
+        });
+    }
+
+    function flattenFormData(formData) {
```
mattandrews: In the original code this was a one-liner:

           function flattenFormData(formData) {
                return Object.assign({}, ...flatMap(formData));
            }

... but this doesn't work with the new structure. Maybe there's a better method?!

assets/sass/emails.scss
```diff
@@ -88,3 +88,79 @@ small {
...
+
+        > p {
+            margin: 0;
+        }
+    }
+
+    td {
+        vertical-align: top;
+
+        // Fix word-break in IE when used in tables
```
davidrapson: We can remove IE fixes from email styles.

PR #1770, opened 2019-02-21
Simple funding prototype flow

davidrapson:
Further modelling of form flow


controllers/apply-next/views/step.njk
```diff
@@ -46,8 +46,7 @@
                 {% endif %}
 
                 <div class="form-actions">
-                    <a href="{{ formBaseUrl }}/summary" class="btn btn--outline">Skip to summary</a>
```
mattandrews: Can you leave this in? Been helping me debug things and I suspect we'll want a link to it to be present on all pages at some point.

PR #1771, opened 2019-02-21
Add form validation for the entire application

mattandrews:
This still needs a bit of thought around structure/timings, but it adds validation of the entire form.

It does this when you POST to `/summary` (this should probably change) by running each field's validator function and setting a `state` property for every field, step and section (and then the form as  a whole), which we then display on the summary screen like this:

image

The main thing missing here is working out how we integrate this so any visit to the summary screen shows this data. We can only run the validators on a POST request, and they can only get their data from a limited subset of `req.*` fields (`req.session` isn't one of them and it doesn't appear we can write a custom `check()`) function. Possibly the solution is to redirect all GET requests to` /summary` to a POST? Bit ugly though... thoughts welcome!


controllers/apply-next/form-router.js
```diff
@@ -81,9 +81,112 @@ function translateForm(formModel, locale, formData) {
...
+            });
+
+            // See if this step's fieldsets are all empty (as opposed to invalid)
+            const stepIsEmpty = isEmpty(
+                pickBy(
+                    step.fieldsets.reduce((acc, fieldset) => {
+                        return concat(fieldset.fields.map(f => f.value), acc);
+                    }, []),
+                    identity
+                )
```
mattandrews: This looks a bit magic – it's similar to the code we use in the grants service to strip out missing/invalid fields (eg. `_.isEmpty(_.pickBy([undefined, null], _.identity))`). It basically reduces each step down to an array of the input values after removing invalid/missing ones, then checks to see if the resulting array is empty or not.

controllers/apply-next/form-router.js
```diff
@@ -236,37 +340,50 @@ function initFormRouter(formModel) {
     /**
      * Route: Summary
      */
+    const validateAllFields = getAllFormFields(formModel).map(getFieldValidator);
     router
         .route('/summary')
         .get(function(req, res) {
-            if (isEmpty(res.locals.formData)) {
```
mattandrews: We probably don't want to block access to the summary page now, at any time.

controllers/apply-next/form-router.js
```diff
@@ -236,37 +340,50 @@ function initFormRouter(formModel) {
...
-                try {
-                    await formModel.processor({
-                        form: res.locals.form,
-                        data: res.locals.formData
+        .post(
+            (req, res, next) => {
+                // Fake a post body so the validators can run as if the entire form
+                // was submitted in one go
+                req.body = res.locals.formData;
+                next();
```
mattandrews: This saves having to output all the form fields as HTML on the summary page.

PR #1775, opened 2019-02-25
Add various styles to improve summary screen

mattandrews:
image


assets/js/modules/forms.js
```diff
@@ -44,6 +44,18 @@ function toggleReviewAnswers() {
...
 }
 
+function handleExpandingDetails() {
+    let isClosed = true;
+    const $toggleBtn = $('.js-toggle-all-details');
+    $toggleBtn.text($toggleBtn.data('label-closed')).show();
+
+    $toggleBtn.on('click', function() {
+        $toggleBtn.text(isClosed ? $toggleBtn.data('label-open') : $toggleBtn.data('label-closed'));
+        $('details').attr('open', isClosed);
```
davidrapson: I'd be tempted to scope this to a specific class, or parent element. _All_ details on the page might get a little weird. Might be overthinking it but we don't know where else we might want to use details on this page.

assets/sass/components/_applications.scss
```diff
@@ -0,0 +1,93 @@
+$statuses: (
+    'complete': #55ac3b,
+    'incomplete': #fd992b,
+    'empty': #7b7b7b,
+    'invalid': red
+);
+
+$statusColumnWidth: 160px;
+
+.app-summary__status {
```
davidrapson: Mind scoping this within `app-summary`? Also I'd be inclined to call this `application-summary` as app reads a little meta software-y.

assets/sass/components/_applications.scss
```diff
@@ -0,0 +1,93 @@
...
+    display: flex;
+    justify-content: space-between;
+    border-left: 3px solid get-color('background', 'light-neutral');
+    padding-left: $spacingUnit / 2;
+
+    @each $state, $color in $statuses {
```
davidrapson: Not a huge fan of generated class names, I find it reduces search-ability. I'd be inclined to write the classes out longhand.

assets/sass/components/_applications.scss
```diff
@@ -0,0 +1,93 @@
...
+}
+
+.app-summary__data {
+    font-size: 18px;
+    display: flex;
+    justify-content: start;
+    border-top: 1px solid get-color('border', 'base');
+
+    > * {
+        padding: $spacingUnit / 2;
```
davidrapson: Really padding on everything as a wild-card?

assets/sass/components/_buttons.scss
```diff
@@ -327,6 +327,27 @@ $buttonColor: get-color('brand', 'primary');
     }
 }
 
+/**
+ * Button group separated left
+ * Allows a set of buttons to be split from other buttons
+ * with a divider on their left
+ */
+.o-button-group--separated {
```
davidrapson: Can this be rewritten so that it's not a modifier? Asking based on the others (`o-button-group-flex` etc.) being standalone classes. It might not make sense here.

PR #1779, opened 2019-02-26
Add data storage

mattandrews:
Work in progress

### DIFF
--- a/controllers/apply-next/form-router.js
+++ b/controllers/apply-next/form-router.js
@@ -187,6 +187,14 @@ function initFormRouter(formModel) {
         editingId: 'currentEditingId'
     };
 
+    // Require login and redirect users back here once authorised
+    router.use((req, res, next) => {
+        req.session.redirectUrl = req.baseUrl;
+        req.session.save(() => {
+            next();
+        });
+    }, requireUserAuth);
+
     router.use(cached.csrfProtection, async (req, res, next) => {
         res.locals.setSessionData = (dataPath, value) => set(req.session, `${sessionKeys.form}.${dataPath}`, value);
         res.locals.getSessionData = dataPath => get(`${sessionKeys.form}.${dataPath}`)(req.session);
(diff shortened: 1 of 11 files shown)


models/index.js
```diff
@@ -38,14 +38,15 @@ sequelize
...
-db.Staff = sequelize.import('../models/staff.js');
+db.Users = sequelize.import('./user.js');
+db.Staff = sequelize.import('./staff.js');
 db.Feedback = sequelize.import('./feedback');
+db.Application = sequelize.import('./application');
 
 db.SurveyAnswer = sequelize.import('./survey');
 
-db.Order = sequelize.import('../models/materials/order.js');
-db.OrderItem = sequelize.import('../models/materials/orderItem.js');
```
davidrapson: The `.js` extension can also go fyi.

controllers/apply-next/views/startpage.njk
```diff
@@ -12,10 +13,35 @@
                 suffix = copy.description | safe
             ) }}
 
-            <h2>Introduction</h2>
-            <p>Form intro goes here</p>
+            {% if applications %}
```
davidrapson: We'll probably need to distinguish between the "dashboard" page, where your applications live, and the "startpage"  which summaries the form and includes basic guidance. This is fine for now though.

mattandrews: Yeah – I figured at this point we only want to show you applications for *this* form, but yeah, we'll tweak this later.

controllers/apply-next/form-router.js
```diff
@@ -196,29 +226,92 @@ function initFormRouter(formModel) {
         next();
     });
 
+    // Require login before using everything else after this
+    router.use((req, res, next) => {
```
davidrapson: This should probably be moved up above (or in to?) the previous `req.use`. Currently it looks like we're doing a bunch of work in the previous middleware that might not apply if the user isn't logged in yet.

services/applications.js
```diff
@@ -1,17 +1,75 @@
 'use strict';
-const request = require('request-promise-native');
-const { APPLICATIONS_SERVICE_ENDPOINT } = require('../modules/secrets');
```
davidrapson: Can probably remove `APPLICATIONS_SERVICE_ENDPOINT` from `modules/secrets` too

PR #1780, opened 2019-02-27
Refine form fields and conditional steps

davidrapson:
- Move validators into a standalone file, add tests
- Determine the next step to render rather than redirecting through in order, avoids slow redirect chains for conditional steps.


bin/www
```diff
@@ -92,7 +92,7 @@ function onListening() {
     const bind = typeof addr === 'string' ? 'pipe ' + addr : 'port ' + addr.port;
     debug(`App now running on ${bind}`);
 
-    if (!!process.env.ENABLE_BROWSER_SYNC === true) {
+    if (process.env.ENABLE_BROWSER_SYNC === 'true') {
```
davidrapson: Was trying to turn browser sync off locally and realised that setting to false was still enabling the feature. Explicitly checking for a 'true' value rather than coercing. Not sure if this is me missing something here.

davidrapson:
@mattandrews As discussed I've removed the summary step validation, so for now you can proceed to the terms screen as long as we _think_ you've completed all the steps. This unblocks us to demo the flow whilst we review options for validation.

controllers/apply-next/views/summary.njk
```diff
@@ -107,41 +107,6 @@
                                                             </div>
                                                         </div>
                                                     {% endfor %}
-
-                                                    <table class="table table-striped mb-5" style="display: none">
```
mattandrews: Ha sorry, forgot I left it in this state when developing 😊

controllers/apply-next/form-router.js
```diff
@@ -313,24 +144,8 @@ function initFormRouter(formModel) {
...
-                res.render(path.resolve(__dirname, './views/section-summary'), {
-                    title: `${sectionLocalised.title} | ${res.locals.form.title}`,
-                    section: sectionLocalised,
-                    backUrl: null, // @TODO: Determine backUrl
-                    nextUrl: `${req.baseUrl}/${sectionModel.slug}/1`
-                });
-            } else {
-                res.redirect(`${req.baseUrl}/${sectionModel.slug}/1`);
-            }
+            res.redirect(`${req.baseUrl}/${sectionModel.slug}/1`);
```
mattandrews: Does this mean we're dropping the concept of a summary "step" for a given section?

controllers/apply-next/helpers.js
```diff
@@ -0,0 +1,186 @@
...
+
+        return step;
+    };
+
+    const clonedForm = cloneDeep(formModel);
+    clonedForm.title = localise(formModel.title);
+
+    clonedForm.sections = clonedForm.sections.map(section => {
+        section = translateSection(section);
+        section.steps = section.steps.map(step => translateStep(step));
```
mattandrews: This can just be:

`section.steps = section.steps.map(translateStep);`

views/components/form-fields/macros.njk
```diff
@@ -52,13 +50,11 @@
...
-        {% if field.autocompleteName %}autocomplete="{{ field.autocompleteName }}"{% endif %}
-        size="{{ field.size | default(40, true) }}"
         {% if field.value %}value="{{ field.value }}"{% endif %}
-        {% if field.customRegex %}pattern="{{ field.customRegex }}"{% endif %}
         {% if field.isRequired %}required aria-required="true"{% endif %}
-        {% if field.min %}min="{{ field.min }}"{% endif %}
-        {% if field.max %}max="{{ field.max }}"{% endif %}
+        {% if field.autocompleteName %}autocomplete="{{ field.autocompleteName }}"{% endif %}
+        {% if not field.attributes.size %}size="{{ field.size | default(40, true) }}"{% endif %}
+        {% if field.attributes %}{% for attr, value in field.attributes %} {{ attr }}="{{ value }}"{% endfor %}{% endif %}
```
mattandrews: Really nice 💯

controllers/apply-next/form-router.js
davidrapson: I think so for now, going to come back to it when we revisit the beneficiaries screen. It felt like it could go on the first screen in the step.

controllers/apply-next/helpers.js
davidrapson: Personally I prefer the explicitness. `section.steps.map(translateStep)` implicitly passes `translateStep(currentValue, index, array)` where the long-hand only passes the explicit arguments.

PR #1782, opened 2019-02-28
Add basic validation with joi

davidrapson:
This is an experimental branch which adds basic validation with `joi`.

The general approach is to:
- define a schema for the whole form
- when submitting a step, validate against the schema and filter the errors based on the current step

image

Still to explore:

- [x] Custom error messages with translation
- [x] Validating the whole form on the review screen and updating status markers to be based on this validation rather than session
- [x] Expand validation schema to cover conditional fields
- [x] Reconcile differences in data format from display format (e.g. dates)


controllers/apply-next/form-router.js
```diff
@@ -33,92 +37,70 @@ function initFormRouter(formModel) {
     router.use(cached.csrfProtection, async (req, res, next) => {
         res.locals.setSessionData = (dataPath, value) => set(req.session, `${sessionKeys.form}.${dataPath}`, value);
         res.locals.getSessionData = dataPath => get(`${sessionKeys.form}.${dataPath}`)(req.session);
-        res.locals.unsetSessionData = dataPath => unset(req.session, `${sessionKeys.form}.${dataPath}`);
```
davidrapson: This wasn't being used.

controllers/apply-next/helpers.js
```diff
@@ -1,186 +1,268 @@
...
             fieldset.legend = localise(fieldset.legend);
             fieldset.introduction = localise(fieldset.introduction);
-            fieldset.fields = fieldset.fields.map(translateField);
+            fieldset.fields = fieldset.fields.map(enhanceField);
             return fieldset;
         });
 
+        // Handle steps that don't need to be completed based on current form data
+        if (step.matchesCondition && step.matchesCondition(data) === false) {
+            step.notRequired = true;
```
davidrapson: Moved the `notRequired` step up into this main function as it lets us simplify the "completion status" function we use on the summary page and means we can perform this check earlier.

controllers/apply-next/helpers.js
```diff
@@ -1,186 +1,268 @@
...
-        });
+/**
+ * Calculate form progress
+ * Validates the form and returns a status
+ * for each section and the form as a whole.
+ *
+ * @param {Object} form
+ * @param {Object} data
+ */
+function calculateFormProgress(form, data) {
```
davidrapson: Managed to simplify this down to a simpler status object:

```
{
  all: 'incomplete':
  sections: {
    'section-slug': 'complete',
    …
  }
}

File: controllers/apply-next/helpers.js [hunk 4]
```diff
@@ -1,186 +1,268 @@
...
+    }
 }
 
-function getFieldsForSection(section) {
-    return flatMap(section.steps, getFieldsForStep);
+/**
+ * Find next and previous matching URLs
+ * @param {MatchOptions} options
+ */
+function nextAndPrevious(options) {
```
davidrapson: Refactored the "determine next step" logic to be usable to work out both the next and previous url. Added tests for this logic.

controllers/apply-next/form-router.js
```diff
@@ -33,92 +37,70 @@ function initFormRouter(formModel) {
...
-        res.locals.breadcrumbs = [{ label: res.locals.form.title, url: req.baseUrl }];
 
         next();
     });
 
     /**
      * Route: Start page
      */
+    async function renderStartPage(req, res, errors = []) {
+        const { startPage } = res.locals.form;
```
mattandrews: Any reason not to keep throwing the errors for missing startpages etc?

controllers/apply-next/form-router.js
```diff
@@ -33,92 +37,70 @@ function initFormRouter(formModel) {
...
-            const { startPage } = res.locals.form;
-            if (!startPage) {
-                throw new Error('No startpage found');
-            }
-            if (!startPage.urlPath && !startPage.template) {
-                throw new Error('No valid startpage types found');
-            }
-            next();
-        })
-        .get(cached.noCache, async function(req, res) {
```
mattandrews: I think we might still need the `cached.noCache` stuff here (possibly on every route?) as when I was using this on the TEST environment, I kept getting redirected to the login page even after logging in (eg. the 302 redirect from `/apply-next/simple` had been cached)?

controllers/apply-next/form-router.js
```diff
@@ -33,92 +37,70 @@ function initFormRouter(formModel) {
...
-                        return renderError(error, req, res);
-                    });
-            } else {
-                const { startPage } = res.locals.form;
-                res.render(startPage.template, {
-                    title: res.locals.form.title,
-                    csrfToken: req.csrfToken(),
-                    errors: errors.array()
                 });
+                res.redirect(`${req.baseUrl}/edit/${application.id}`);
```
mattandrews: Does this need to be in `then()` callback (or should it be `await applicationsService`...)? Otherwise isn't there a race condition where it might redirect before the application has been saved?

controllers/apply-next/simple/fields.js
```diff
@@ -0,0 +1,682 @@
...
+        messages: {
+            base: { en: 'Enter a companies house number' }
+        }
+    },
+    accountingYearDate: {
+        name: 'accounting-year-date',
+        label: { en: 'What is your accounting year end date?', cy: '' },
+        type: 'date',
+        isRequired: true,
+        schema: Joi.date().min('now'),
```
mattandrews: Is this right? I think this logic is saying the accounting end date has to be today (or later), surely this should be the `max`? Also we should specify that value as an attribute for the client-side validation too, right?

controllers/apply-next/simple/fields.js
```diff
@@ -0,0 +1,682 @@
...
+        isRequired: true,
+        schema: Joi.string().required(),
+        messages: {
+            base: { en: 'Enter a sort-code', cy: '' }
+        }
+    },
+    bankAccountNumber: {
+        name: 'bank-account-number',
+        label: { en: 'Account number', cy: '' },
+        type: 'string',
```
mattandrews: Back and forth on this! If we're requiring a number in the schema then it should be `type: number`, right? Happy that sort code can be a string (for now...) but this is always just a number.

controllers/apply-next/validators.js
```diff
@@ -1,98 +1,24 @@
...
-            .isBefore(moment().toISOString())
-            .withMessage(
-                localiseMessage({
-                    en: 'Date must be in the past',
-                    cy: 'WELSH ERROR'
-                })
-            );
+    postcode: Joi.string()
+        // via https://github.com/chriso/validator.js/blob/master/lib/isPostalCode.js#L54
+        .regex(/^(gir\s?0aa|[a-z]{1,2}\d[\da-z]?\s?(\d[a-z]{2})?)$/i)
```
mattandrews: Worth making this field shared somewhere so we don't define it twice?

views/components/form-fields/macros.njk
```diff
@@ -58,6 +58,22 @@
...
+        {{ fieldLabelText(field, fieldCopy) }}
+    </label>
+    {{ fieldHelpText(field, fieldCopy) }}
+    <input
+        class="ff-text"
+        type="date"
+        id="field-{{ field.name }}"
+        name="{{ field.name }}"
+        {% if field.value %}value="{{ formatDate(field.value, 'YYYY-MM-DD') }}"{% endif %}
+        {# {% if field.isRequired %}required aria-required="true"{% endif %} #}
```
mattandrews: Should this be uncommented?

controllers/apply-next/simple/fields.js
```diff
@@ -0,0 +1,682 @@
...
+        isRequired: true,
+        schema: Joi.string().required(),
+        messages: {
+            base: { en: 'Enter the name on the account', cy: '' }
+        }
+    },
+    bankSortCode: {
+        name: 'bank-sort-code',
+        label: { en: 'Sort code', cy: '' },
+        type: 'string',
```
mattandrews: This (and the one below) should be `type: text`, right?

controllers/apply-next/form-router.js
davidrapson: It is on all the routes:

```
router.use((req, res, next) => {
        req.session.redirectUrl = req.baseUrl;
        req.session.save(() => {
            next();
        });
    }, requireUserAuth);

    router.use(cached.csrfProtection, async (req, res, next) => {
        res.locals.setSessionData
```

`cached.csrfProtection` also applies `noCache`. **But** I think the redirect thing you are talking about it because of the order of the middleware above, which needs to come after the csrf/nocache rules. Will take a look at merging the middleware to fix that.

controllers/apply-next/form-router.js
davidrapson: Forgot the `await` keyword, thanks!

controllers/apply-next/simple/fields.js
davidrapson: I think this is right, the accounting end date needs to be in the future, so needs to be _at least_ now?

mattandrews: It does? I guess I'd assumed that you're submitting a summary of your accounts up to a certain (past) date? But fine, let's leave it as-is as I'm sure the phrasing/validation will evolve over time.

controllers/apply-next/simple/fields.js
davidrapson: Data-wise it might be numeric-ish, but I disagree that number is the right input type. It's not a discreet value (like a price or quantity). The increment/decrement controls on an `input[type=number]` don't make sense for an account number. In terms of UI a normal text input feels the best fit.

controllers/apply-next/validators.js
davidrapson: The regex or the whole field?

controllers/apply-next/simple/fields.js
mattandrews: Fair enough! https://www.brucelawson.co.uk/2012/credit-card-bank-account-numbers-in-html5-forms/ bears you out.

controllers/apply-next/validators.js
mattandrews: The regex, sorry.

davidrapson: I'll come back to this in a separate PR. The version used for the front-end is double escaped so they are not exactly the same.

controllers/apply-next/simple/fields.js
davidrapson: Oh, I'd interpreted it completely differently, but you're right, we should leave it and confirm what the actual rule is.

controllers/apply-next/form-router.js
davidrapson: Yep, I think we can better guard against this by validating the form schema with `joi` or via a unit test. We can catch this earlier than at render.

PR #1791, opened 2019-03-05
Update user validation

davidrapson:
Updates the user authentication routes to use `joi` for validation. There's a fair bit going on here, so will add inline comments to expand on the changes here.

### DIFF
--- a/controllers/user/forgotten-password.js
+++ b/controllers/user/forgotten-password.js
@@ -1,25 +1,19 @@
 'use strict';
 const path = require('path');
 const jwt = require('jsonwebtoken');
-const { validationResult } = require('express-validator/check');
 const express = require('express');
 const Raven = require('raven');
+const Joi = require('joi');
 
 const { getAbsoluteUrl } = require('../../modules/urls');
 const { JWT_SIGNING_TOKEN } = require('../../modules/secrets');
-const { purifyUserInput } = require('../../modules/validators');
+const { normaliseErrors } = require('../../modules/errors');
 const { requireUnauthed } = require('../../middleware/authed');
 const { sendEmail } = require('../../services/mail');
-const { validators } = require('./helpers');
 const userService = require('../../services/user');
+const schema = require('./schema');
 
 const router = express.Router();
-
-function renderRequestReset(req, res) {
-    res.render(path.resolve(__dirname, './views/forgotten-password'), {
-        errors: res.locals.errors || []
-    });
-}
 
 async function processResetRequest(req, res, user) {
     const payload = { data: { userId: user.id, reason: 'resetpassword' } };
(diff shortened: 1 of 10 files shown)


controllers/user/forgotten-password.js
```diff
@@ -44,37 +38,51 @@ async function processResetRequest(req, res, user) {
...
 }
 
 router
     .route('/')
     .all(requireUnauthed)
-    .get(renderRequestReset)
-    .post(validators.emailAddress, handleRequestReset);
+    .get(renderForm)
+    .post(async function(req, res) {
+        const validationResult = Joi.object({
```
davidrapson: Validation here is only to check that the username is a valid format (e.g. an email)

controllers/user/login.js
```diff
@@ -35,13 +37,11 @@ router
...
         // @TODO consider rate limiting?
-        passport.authenticate('local', (err, user, info) => {
-
+        passport.authenticate('local', (err, user) => {
             if (!user) {
                 // User not valid, send them to login again
-                res.locals.errors = [{ msg: info.message }];
-                res.locals.formValues = req.body;
-                return renderForm(req, res);
+                const errors = [{ msg: 'Username or password is invalid' }];
```
davidrapson: Noticed whilst testing this that the errors returned to the front-end was the raw passport error. For login we only want to render a generic error that the log in credentials weren't valid. There was no express-validator code on this route before, we may want to add some basic validation like the forgotten password screen, if only to check that the username is an email address.

controllers/user/reset-password.js
```diff
@@ -56,50 +49,72 @@ router
...
 
-        if (!changePasswordToken) {
-            res.redirect('/user/login');
+        if (!token) {
+            redirect(req, res);
         } else {
-            // store the token to pass on to other requests (eg. outside the URL)
-            res.locals.token = changePasswordToken;
-            const errors = validationResult(req);
+            try {
```
davidrapson: This controller has the biggest structural changes:

- Separates out verification of the token from checking the user
- Decodes the token first and then validates the new password, passes in the username from the token to allow us to check against conditional rules e.g. password must not match username
- Only once we know the format of the data is good do we lookup the user and try to update the password

controllers/user/views/forgotten-password.njk
```diff
@@ -21,18 +21,22 @@
...
                     name: 'username',
                     label: 'Email address',
+                    explanation: 'Enter the email address associated with your account.',
                     isRequired: true,
                     value: formValues['username']
                 }, errors = errors) }}
 
                 <div class="form-actions">
-                    <input type="submit" class="btn" value="Reset password request" />
+                    <input type="submit" class="btn" value="Request password reset" />
```
mattandrews: 🙃

davidrapson: Honestly not sure If this more or less grammatically correct now. We might be better off with simply "Reset password:

PR #1793, opened 2019-03-06
Word count component

davidrapson:
Adds an initial version of a word-count as a vue component using an `aria-live` region. Using word-count as the count rather than characters until we've done more testing on variations. Includes unit tests for word boundaries and the component itself. Done a small pass of testing with VoicerOver too.

Our main file for vue component was getting a little unwieldy so I've moved "global" component initialisation (i.e. cookie consent, site feedback) to a separate file.

The counter looks like this:

word-count


controllers/apply-next/simple/fields.js
```diff
@@ -160,6 +160,10 @@ const allFields = {
             cy: 'TODO'
         },
         type: 'textarea',
+        settings: {
+            showWordCount: true,
+            maxWords: 100
```
mattandrews: Should we use this same value in the Joi validation schema too?

views/components/form-fields/macros.njk
```diff
@@ -101,18 +101,30 @@
...
-        class="ff-textarea"
-        id="field-{{ field.name }}"
-        name="{{ field.name }}"
-        {% if field.isRequired %}required aria-required="true"{% endif %}
-        {% if field.rows %}rows="{{ field.rows }}"{% endif %}
-        {% if field.attributes %}{% for attr, value in field.attributes %} {{ attr }}="{{ value }}"{% endfor %}{% endif %}
-    >{{ field.value }}</textarea>
+    {% set settings = field.settings %}
+    {% if settings.showWordCount %}<div class="js-character-count">{% endif %}
+        <label class="ff-label" for="field-{{ field.name }}">
```
mattandrews: Minor thing but this whole block needs to be outdented as it looks a bit like it's inside the `{% if %}` block above.

controllers/apply-next/simple/fields.js
davidrapson: Quite possibly, but I'm OK leaving it separate for now. Something to explore, my take is that we might want the server limit to be a tad higher. i.e. the word count encourages 500 words, but if you submit 501 we might not stop you.

PR #1794, opened 2019-03-07
New multi-field date input

davidrapson:
date-input

### DIFF
--- a/controllers/apply-next/simple/fields.js
+++ b/controllers/apply-next/simple/fields.js
@@ -105,7 +105,7 @@ const allFields = {
         },
         type: 'date',
         isRequired: true,
-        schema: commonValidators.futureDate('12', 'weeks'),
+        schema: commonValidators.futureDate({ amount: '12', unit: 'weeks' }),
         messages: {
             base: { en: 'Enter a date', cy: '' },
             'date.isoDate': { en: 'Enter a real date', cy: '' },
@@ -337,11 +337,15 @@ const allFields = {
         name: 'accounting-year-date',
         label: { en: 'What is your accounting year end date?', cy: '' },
         type: 'date',
+        settings: {
+            minYear: moment().format('YYYY')
+        },
         isRequired: true,
-        schema: Joi.date().min('now'),
+        schema: commonValidators.futureDate(),
         messages: {
-            base: { en: 'Enter an accounting year end date', cy: '' },
-            'date.min': { en: 'Accounting year end date must be in the future', cy: '' }
+            base: { en: 'Enter a date', cy: '' },
+            'date.isoDate': { en: 'Enter a real date', cy: '' },
+            'date.min': { en: 'Date must be in the future', cy: '' }
         }
     },
     totalIncomeYear: {
--- a/controllers/apply-next/simple/fields.js
+++ b/controllers/apply-next/simple/fields.js
@@ -1,9 +1,8 @@
 'use strict';
 const { forEach, reduce, values } = require('lodash');
-const Joi = require('joi');
 const moment = require('moment');
 
-const commonValidators = require('../validators');
+const { Joi, ...commonValidators } = require('../validators');
 
 const organisationTypes = {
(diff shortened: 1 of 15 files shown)


controllers/apply-next/validators.js
```diff
@@ -1,8 +1,33 @@
 'use strict';
-const Joi = require('joi');
 const moment = require('moment');
+const baseJoi = require('joi');
+const { toInteger, isObject } = require('lodash');
+
+const Joi = baseJoi.extend(joi => ({
+    base: joi.date(),
+    name: 'dateObject',
+    coerce: function(value, state, options) {
```
davidrapson: This is the thing that lets us submit the date in parts but still validate as a full date object. Uses a joi extension to coerce the nested value into a date string. It's possible to raise errors here through `this.createError`.

controllers/apply-next/simple/fields.js
```diff
@@ -96,22 +90,19 @@ const allFields = {
             cy: '(WELSH) When is the planned (or estimated) start date of your project?'
         },
         explanation: {
-            en:
-                'This date needs to be at least 12 weeks from when you plan to submit your application. If your project is a one-off event, please tell us the date of the event.',
-            cy:
-                '(WELSH) This date needs to be at least 12 weeks from when you plan to submit your application. If your project is a one-off event, please tell us the date of the event.'
+            en: `
+                <p><strong>For example: 12 11 2020</strong></p>
+                <p>This date needs to be at least 12 weeks from when you plan to submit your application. If your project is a one-off event, please tell us the date of the event.</p>
```
mattandrews: I wonder if we should add in the actual (dynamic) date this should be here?

views/components/form-fields/macros.njk
```diff
@@ -62,20 +62,47 @@
...
+                    value="{{ field.value.month or formatDate(field.value, 'M') }}"
+                    {% if field.isRequired %}required aria-required="true"{% endif %}
+                />
+            </label>
+            <label class="ff-inline">
+                <span class="ff-label">{{ __('global.misc.year') }}</span>
+                <input
+                    class="ff-text ff-width-4"
+                    type="number"
+                    pattern="[0-9]*" min="1"
```
mattandrews: Would it be worth requiring a 4 digit year here to avoid ambiguity, or are we happy letting `moment` handle that? Likewise a `min` of 1 is a bit... unlikely, right?

mattandrews: (I get that it prevents negative numbers, but it also means that the year 1 is now valid input...?

davidrapson: Yeah, this is a bare minimum check. The server logic will obviously catch invalid dates, but we might want to pass an appropriate minimum date based on the field (e.g. "current year" for future dates, "~1900" for date of birth)

controllers/apply-next/simple/fields.js
davidrapson: Good shout. This relates to the point below, if we configure this on the field we could a) interpolate this into the message and b) workout an appropriate minimum year for the input.

davidrapson:
As discussed added some more nuanced validation based on minimum date

image

views/components/form-fields/macros.njk
```diff
@@ -62,20 +62,47 @@
...
+                    value="{{ field.value.month or formatDate(field.value, 'M') }}"
+                    {% if field.isRequired %}required aria-required="true"{% endif %}
+                />
+            </label>
+            <label class="ff-inline">
+                <span class="ff-label">{{ __('global.misc.year') }}</span>
+                <input
+                    class="ff-text ff-width-4"
+                    type="number"
+                    pattern="[0-9]*" min="{{ field.settings.minYear | default(1900) }}"
```
davidrapson: Setting a minimum here, based on a date of birth answer. Might need revisiting, but is better than a default of "1". This falls down if we start asking people "On which date did the Battle of Hastings begin?"

PR #1796, opened 2019-03-08
Add budget tool input field

mattandrews:
This still needs a bit of `a11y` love so will work on that now, but in the meantime it's ready for review generally (eg. the approach and structure etc).

### JS-enhanced version
budget

### Non-JS fallback
no-js

### Summary view
image


assets/js/vue-apps/form-components.js
```diff
@@ -17,8 +19,108 @@ function initCharacterCount() {
...
+                ready: false,
+                error: null
+            };
+        },
+        watch: {
+            items: {
+                handler() {
+                    const lastItem = this.items[this.items.length - 1];
+                    if (lastItem) {
+                        if (this.items.length < this.maxItems && (!isEmpty(lastItem.item) || !isEmpty(lastItem.cost))) {
```
mattandrews: The thinking here by @LynseyJReynolds was that it should automatically add a new row as soon as you populate either field in the row above.

controllers/apply-next/validators.js
```diff
@@ -1,31 +1,91 @@
...
+                        { v: value, number: this._flags.maxBudget },
+                        state,
+                        options
+                    );
+                }
+            }
+            return value;
+        },
+        language: {
+            maxBudget: 'needs to be a number'
```
mattandrews: This error gets shown in your console (eg. when developing), not in the browser, FYI.

assets/js/vue-apps/form-components.js
davidrapson: Makes sense. I'd suggest extracting this out into a method within component as this logic could get a little complex.

assets/js/vue-apps/form-components.js
```diff
@@ -17,8 +19,117 @@ function initCharacterCount() {
...
+                        this.clearError();
+                    }
+                },
+                deep: true
+            }
+        },
+        mounted: function() {
+            // Clear out <noscript> fallbacks otherwise they submit as well as the
+            // Vue-enhanced fields (eg. double inputs)
+            $(this.$el)
```
davidrapson: Interesting. It might be worth comparing this to how the past grants works. There's some Vue behaviour where content inside the component or vue scope will get replaced with any new content.

davidrapson: That wasn't explained well, will see if I can find an example of what I mean.

controllers/apply-next/validators.js
```diff
@@ -1,31 +1,91 @@
...
+
+const budgetValidator = joi => {
+    return {
+        base: joi.array(),
+        name: 'budgetItems',
+        coerce(value, state, options) {
+            if (isArray(value)) {
+                // Strip out anything that doesn't have an item name and a cost
+                // (eg. validate things that are half-supplied, but not empty)
+                return filter(value, line => !isEmpty(line.item) || !isEmpty(line.cost));
```
davidrapson: I'd maybe write this as `reject(value, line => isEmpty(line.item) || isEmpty(line.cost))` to avoid the inverted logic.

views/components/form-fields/macros.njk
```diff
@@ -214,6 +217,122 @@
...
+{% endmacro %}
+
+{% macro inputBudget(field, fieldCopy) %}
+    <label class="ff-label" for="field-{{ field.name }}">
+        {{ fieldLabelText(field, fieldCopy) }}
+    </label>
+
+    {{ fieldHelpText(field, fieldCopy) }}
+
+    <div class="ff-budget"
```
davidrapson: This should really be a vue component rather than inline, we can then use props to pass data around rather that data attrs. Same approach as past grants.

views/components/icons.njk
```diff
@@ -136,7 +136,7 @@
          viewBox="0 0 459 459" width="14" height="14"
          aria-labelledby="aria-i-bin-title"
     >
-        <title id="aria-i-bin-title">A waste bin</title>
+        <title id="aria-i-bin-title">A waste bin icon</title>
```
davidrapson: Mind converting this icon to use the `icon` macro (see `iconSearch`) it will fix a11y issues complaining about the icon needing a unique id.

controllers/apply-next/simple/fields.js
```diff
@@ -184,14 +185,37 @@ const allFields = {
...
+            'budgetItems.overBudget': {
+                en: `Ensure your budget total is £${MAX_BUDGET_TOTAL.toLocaleString()} or less.`,
+                cy: `(WELSH) Ensure your budget total is £${MAX_BUDGET_TOTAL.toLocaleString()} or less.`
+            }
+        },
+        displayFormat(value) {
+            if (!isArray(value)) {
+                return value;
+            } else {
+                const total = value.reduce((acc, cur) => {
```
davidrapson: Not sure how I feel about this whole thing. Html here feels unpleasant, but I'm not sure I have a better idea at the moment.

views/components/form-fields/macros.njk
mattandrews: I started it like that but I wanted to approach this like a hybrid component as much as possible (eg. like the free merch material form) and keep the code together rather than having two lots of similar/related markup in both a Vue file and in this macro. Past grants uses a bunch of JSON embedded in `<script>` tags to populate state – I think anytime we're doing this kind of PE work with Vue, there's going to be tradeoffs. In this instance I wanted to keep all the related code together where possible.

views/components/form-fields/macros.njk
```diff
@@ -214,6 +217,122 @@
     </fieldset>
 {% endmacro %}
 
+{% macro budgetLineItemRow(field, counter, lineItem = false) %}
+    <tr>
+        <td>
+            <input type="text"
```
davidrapson: These need labels.

controllers/apply-next/simple/fields.js
mattandrews: Only option I can think of is to pass through the path of a template fragment to use when rendering it but that feels like overkill. Otherwise this code could move to the summary screen directly but that sort of defeats the object of this function.

davidrapson: Yeah I think it's fine. Perhaps try a template string to avoid the string concatenating, otherwise probably makes sense.

views/components/form-fields/macros.njk
davidrapson: Given the number of things this component has to cover it would be nice if we could get it into a self contained vue component as I think that gives us some strong benefits when it comes to testing as we can run much faster unit-level tests on the component using the `vue-test-utils` stuff in addition to some coverage from Cypress. As it stands we'd have to cover everything just using Cypress which would be much slower. I might be missing something where we could get the best of both worlds though, so open to exploring a few things.

controllers/apply-next/simple/fields.js
mattandrews: Just tried that and the ` | nl2br` filter on the summary screen clobbers it (adds in extra markup etc):

image

Will leave as-is for now.

davidrapson: 👍

views/components/form-fields/macros.njk
mattandrews: Just converted this over to a self-contained component, give it a look!

assets/js/vue-apps/form-components.js
mattandrews: Got this to work – the trick is to use a self-closing element, which seems to then consume anything following the component (but not before). Hope this isn't a bug...

PR #1798, opened 2019-03-08
Form refinement

davidrapson:
WIP brand which refines the new application workflow. Separates the concept of dashboard from new application screen (adding country choices) and restructures router to more explicitly guard against reaching steps without a valid application.

### DIFF
--- a/views/components/form-fields/macros.njk
+++ b/views/components/form-fields/macros.njk
@@ -55,7 +55,7 @@
         id="field-{{ field.name }}"
         name="{{ field.name }}"
         {% if field.value %}value="{{ field.value }}"{% endif %}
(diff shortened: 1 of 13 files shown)


views/components/form-fields/macros.njk
```diff
@@ -55,7 +55,7 @@
         id="field-{{ field.name }}"
         name="{{ field.name }}"
         {% if field.value %}value="{{ field.value }}"{% endif %}
-        {% if field.isRequired %}required aria-required="true"{% endif %}
+        {# {% if field.isRequired %}required aria-required="true"{% endif %} #}
```
mattandrews: Should this be here?

PR #1799, opened 2019-03-11
AFA prototype enhancements

mattandrews:
A few things in here:

- We now always pass users a `?redirectUrl` parameter when logging in so we don't need to manually specify where to return to after authing.

- Adds the ability to delete an application (after confirmation)

- Adds an eligibility section. This comes _after_ the initial application creation, so that we can track a user's progress through it and prevent them skipping ahead. If they fail the eligibility check then we mark the application as such and no longer let them edit it. 

I suspect the above flow will change over time but this gives us something demoable.

## Eligible application flow
elig

## Ineligible application flow
inelig


controllers/apply-next/form-router.js
```diff
@@ -147,6 +142,34 @@ function initFormRouter(formModel) {
...
+                res.render(path.resolve(__dirname, './views/delete'), {
+                    title: res.locals.formTitle,
+                    csrfToken: req.csrfToken()
+                });
+            } else {
+                res.redirect(req.baseUrl);
+            }
+        })
+        .post(async (req, res) => {
+            const deleteApplication = applicationsService.deleteApplication(
```
davidrapson: How come you're not using `async/await` here?

controllers/apply-next/form-router.js
```diff
@@ -147,6 +142,34 @@ function initFormRouter(formModel) {
         }
     });
 
+    // Allow an application owned by this user to be deleted
+    router
+        .route('/delete/:applicationId')
+        .get((req, res) => {
+            if (req.params.applicationId && req.user.userData.id) {
```
davidrapson: Should we look up the application here first? Only render this screen if the user owns the application?

controllers/apply-next/form-router.js
```diff
@@ -186,6 +210,79 @@ function initFormRouter(formModel) {
...
+                    await applicationsService.changeApplicationState(res.locals.currentlyEditingId, 'eligible');
+                    res.render(path.resolve(__dirname, './views/eligibility'), {
+                        eligibility: 'eligible'
+                    });
+                }
+            }
+        });
+
+    // Any pages after this point must be for eligible applications
+    router.use(requireEligible);
```
davidrapson: Maybe inline `requireEligible` as we only use it once?

controllers/apply-next/form-router.js
```diff
@@ -186,6 +210,79 @@ function initFormRouter(formModel) {
...
+                    set(
+                        req.session,
+                        `${SESSION_PREFIX}.eligibilityProgress.${res.locals.currentlyEditingId}`,
+                        nextStepNumber
+                    );
+                    res.redirect(`${req.baseUrl}/eligibility/${nextStepNumber}`);
+                } else {
+                    await applicationsService.changeApplicationState(res.locals.currentlyEditingId, 'eligible');
+                    res.render(path.resolve(__dirname, './views/eligibility'), {
+                        eligibility: 'eligible'
```
davidrapson: Might be worth extracting these different statuses out into a constant?

controllers/apply-next/views/eligibility.njk
```diff
@@ -0,0 +1,58 @@
+{% extends "layouts/main.njk" %}
+{% from "components/form-fields/macros.njk" import inputChoice with context %}
+
+{% block title %}Eligibility | {{ formTitle }} | {% endblock %}
+
+{% block content %}
+    <main role="main" id="content">
+        <div class="content-box u-inner u-inner-wide-only">
+            <div class="s-prose u-constrained-wide">
```
davidrapson: Wrapping this with `s-prose` will make form fields look a little wonky, might want to scope this a bit more, or remove it.

controllers/apply-next/form-router.js
mattandrews: Good catch!

controllers/apply-next/form-router.js
mattandrews: I thought about this but the constant would just use the same keys as the strings here so it feels a bit redundant? eg.

```
// before: eligibility: 'eligible'
//  after: eligibility: ELIGIBILITY_STATES.eligible

controllers/apply-next/views/eligibility.njk
mattandrews: 👍 (was a copy/paste job...)

controllers/apply-next/form-router.js
davidrapson: Can be useful if you can get the code to throw if we reference a nonexistent state, otherwise yeah it might be redundant.

davidrapson:
👍 New changes look good, ship it.

PR #1801, opened 2019-03-12
Add support for conditional options, for legal contact role

davidrapson:
Add support for conditional options, for legal contact role. Filters options based on a `matchesCondition` function (same idea as conditional steps). Mapping of values is based

- Always show chair, vice-chair, treasurer, trustee
- If a company number has been provided show director and company-secretary
- If a statutory body show clerk
- If a school show head teacher

This mapping is based on the current form as a starting point. Still needs thought about how to flag that options may be limited if you've not answered the organisation type yet (i.e. done steps out of order)

### DIFF
--- a/controllers/apply-next/simple/fields.test.js
+++ b/controllers/apply-next/simple/fields.test.js
@@ -1,0 +1,36 @@
+/* eslint-env jest */
+'use strict';
+
+const { filterOptionsBy } = require('../helpers');
+const { legalContactRoles } = require('./fields');
+
+describe('legalContactRoles', () => {
+    test('return default roles', () => {
+        expect(legalContactRoles.filter(filterOptionsBy()).map(option => option.value)).toEqual([
+            'chair',
+            'vice-chair',
+            'treasurer',
+            'trustee'
+        ]);
+    });
+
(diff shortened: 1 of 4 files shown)


controllers/apply-next/simple/form-model.js
```diff
@@ -182,16 +182,7 @@ const sectionLegalContact = {
             fieldsets: [
                 {
                     legend: { en: 'Who is your legally responsible contact?', cy: '' },
-                    introduction: {
```
davidrapson: Moved this help text inline as each paragraph really related to a discreet field.

controllers/apply-next/simple/fields.test.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-env jest */
+'use strict';
+
+const { filterOptionsBy } = require('../helpers');
+const { legalContactRoles } = require('./fields');
+
+describe('legalContactRoles', () => {
```
davidrapson: This test is a bit speculative. It might be better off to have a high level test of enhance form which asserts all rules based on mock submissions, rather than lower level tests like this.

PR #1802, opened 2019-03-12
Specify a different type of Twitter card if we have an image

mattandrews:
If we specify a trail image on a press release, we produce a 2:1 crop for Twitter cards which results in this output:

image

Turns out we should be using a different type of card in this instance.


mattandrews:
(the image it's loading is https://biglotteryfund-assets.imgix.net/media/trail-photos/ESOL-class.png?auto=compress%2Cformat&crop=entropy&fit=crop&h=405&w=1373&s=252594f6170466e8c65db92e9897a6ed)

PR #1804, opened 2019-03-12
Validator cleanup

davidrapson:
- Make all common validators functions for consistency
- Simplify budget validator

### DIFF
--- a/controllers/apply-next/validators.js
+++ b/controllers/apply-next/validators.js
@@ -1,7 +1,7 @@
 'use strict';
 const moment = require('moment');
 const baseJoi = require('joi');
-const { isEmpty, isArray, reject, toInteger } = require('lodash');
+const { isEmpty, isArray, reject, toInteger, sumBy } = require('lodash');
 
 const { POSTCODE_REGEX } = require('../../modules/postcodes');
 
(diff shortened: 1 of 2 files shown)


controllers/apply-next/validators.js
```diff
@@ -80,8 +80,25 @@ const dateParts = joi => {
 
 const budgetValidator = joi => {
     return {
-        base: joi.array(),
+        base: joi
```
davidrapson: The base can be any joi schema, so the full schema can live in here, rather than needing to be defined when called.

controllers/apply-next/validators.js
```diff
@@ -92,42 +109,24 @@ const budgetValidator = joi => {
                 return value;
             }
         },
-        pre(value, state, options) {
```
davidrapson: The ` if (!params.number) {` check isn't needed and the main total budget check can be done directly in the `validate` function. Because the max budget parameter is an explicit value (must be a number) we don't need to do this check ourselves and can get it straight from params.

I think the `_flags` example in the docs was to cover the example of `round` where that is a flag to _mutate_ a value (i.e. round it) rather than validate against it like we're doing here.

davidrapson:
Yeah, I mostly went by the list of properties. Those descriptions do explain what each stage is best used for, but the example is a little misdirecting.

PR #1809, opened 2019-03-13
Add component test for budget-input

davidrapson:
Adds component tests for the budget input.

### DIFF
--- a/assets/js/vue-apps/components/budget-input.vue
+++ b/assets/js/vue-apps/components/budget-input.vue
@@ -10,16 +10,9 @@ export default {
         maxItems: { type: Number, required: true },
         budgetData: { type: Array, required: false }
     },
-    mounted() {
-        if (this.budgetData) {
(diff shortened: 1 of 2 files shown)


assets/js/vue-apps/components/budget-input.vue
```diff
@@ -10,16 +10,9 @@ export default {
...
-        if (this.budgetData) {
-            this.budgetRows = this.budgetData;
-        } else {
-            this.addNewItem();
-        }
-    },
     data() {
         return {
-            budgetRows: [],
+            budgetRows: this.budgetData || [{ item: '', cost: '' }],
```
davidrapson: I know this is repeating the object schema, but in testing using mounted and updating state caused issues, setting explicitly seems to be the way to go, even at the expense of a bit of duplication.

mattandrews: 👍

PR #1811, opened 2019-03-14
Document archive 404

davidrapson:
Pointed out by @LynseyJReynolds that old documents are showing too prominently in search results. Spotted that we are not returning an explicit 404 response for archived documents, so whilst the file has been removed we're probably signalling to google that there's still content at the URL. Making this a 404.

Once this is out we can submit some removal requests for the most accessed old documents by submitting a removal request through webmaster tools (there's about a dozen that make up the majority of traffic to archived documents). This tool requires that page is already 404ing for a request to be successful so this needs to go out first.

### DIFF
--- a/controllers/aliases.js
+++ b/controllers/aliases.js
@@ -109,6 +109,7 @@ const aliases = {
 
     //===== Funding programmes =====//
     '/global-content/programmes/england/awards-for-all-england': '/funding/programmes/national-lottery-awards-for-all-england',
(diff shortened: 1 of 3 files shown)


controllers/aliases.js
```diff
@@ -109,6 +109,7 @@ const aliases = {
 
     //===== Funding programmes =====//
     '/global-content/programmes/england/awards-for-all-england': '/funding/programmes/national-lottery-awards-for-all-england',
+    '/global-content/programmes/england/awards-for-all-england/a4alightpilotintro': '/funding/programmes/national-lottery-awards-for-all-england',
```
davidrapson: Adding in this alias as it keeps cropping up.

PR #1812, opened 2019-03-14
Day month

davidrapson:
Adds a new field type for day and month only. Uses current year to check it's a valid date but only stores and renders the day and month.

image

### DIFF
--- a/assets/js/vue-apps/form-components.js
+++ b/assets/js/vue-apps/form-components.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 import Vue from 'vue';
 import $ from 'jquery';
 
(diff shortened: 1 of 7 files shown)


assets/js/vue-apps/form-components.js
```diff
@@ -9,12 +10,20 @@ function initCharacterCount() {
...
+                'character-count': CharacterCount
+            },
             data: {
                 // Assumes a v-model on the input within the component template
                 text: null
             },
-            components: {
-                'character-count': CharacterCount
+            mounted() {
+                this.text = this.$refs.textarea.value;
```
davidrapson: This fixes a bug where using `v-model` on the text area causes the input to be classed as a "managed component" so text only appears if vue knows about it. This meant that previously saved answers wouldn't show. Watching for input events and assigning the value manually avoids this.

PR #1817, opened 2019-03-15
Add AFA dashboard

mattandrews:
image

image

### DIFF
--- a/controllers/apply-next/simple/form-model.js
+++ b/controllers/apply-next/simple/form-model.js
@@ -376,6 +376,13 @@ module.exports = {
             }
         ],
         schema: schema,
-        processor: processor
+        processor: processor,
+        // @TODO these links will change when these docs are updated - should we just link to the programme page?
+        documentation: {
+            guidance:
+                'https://www.tnlcommunityfund.org.uk/media/documents/awards-for-all/England/A4AOnlineGuidance-03-2019.pdf?mtime=20190312112420',
+            terms:
+                'https://www.tnlcommunityfund.org.uk/media/documents/National-Lottery-Awards-for-All-Terms-and-Conditions.pdf?mtime=20181116141630'
+        }
     }
 };
--- a/controllers/apply-next/simple/form-model.js
+++ b/controllers/apply-next/simple/form-model.js
@@ -377,12 +377,6 @@ module.exports = {
         ],
         schema: schema,
         processor: processor,
-        // @TODO these links will change when these docs are updated - should we just link to the programme page?
-        documentation: {
-            guidance:
-                'https://www.tnlcommunityfund.org.uk/media/documents/awards-for-all/England/A4AOnlineGuidance-03-2019.pdf?mtime=20190312112420',
-            terms:
-                'https://www.tnlcommunityfund.org.uk/media/documents/National-Lottery-Awards-for-All-Terms-and-Conditions.pdf?mtime=20181116141630'
-        }
+        programmePage: '/funding/programmes/national-lottery-awards-for-all-england'
     }
 };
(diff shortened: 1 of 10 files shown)


controllers/apply-next/simple/form-model.js
```diff
@@ -376,6 +376,13 @@ module.exports = {
             }
         ],
         schema: schema,
-        processor: processor
+        processor: processor,
+        // @TODO these links will change when these docs are updated - should we just link to the programme page?
```
davidrapson: Yes please, we should encourage web pages over PDFs where we can, so let's link to pages for now.

controllers/apply-next/views/dashboard.njk
```diff
@@ -2,38 +2,159 @@
...
 {% from "components/form-header/macro.njk" import formHeader with context %}
 {% from "components/form-fields/macros.njk" import formErrors, formField with context %}
+{% from "components/icons.njk" import iconBin %}
+{% from "./view-helpers.njk" import showSectionProgress with context %}
 
 {% extends "layouts/main.njk" %}
 
+{% set fieldMissing %}<span class="application-data-missing">not yet completed</span>{% endset %}
+
+{% macro getStartDate(dateParts) %}
```
davidrapson: We possibly want to make this a view function like `formatDate` for this, but OK here for now.

controllers/apply-next/views/dashboard.njk
```diff
@@ -2,38 +2,159 @@
...
+                                        <a href="{{ form.documentation.terms }}">Terms and Conditions for {{ formTitle }}</a>
+                                    </p>
+                                </aside>
+                            </div>
+
+                            {# Section progress #}
+                            <div class="application-dashboard__progress">
+
+                                <div class="application-dashboard__progress__inner">
+                                    {% set expiryDate = application.createdAt | addTime(3, 'months') %}
```
davidrapson: We probably want to save the expiry date in the model.

controllers/apply-next/views/view-helpers.njk
```diff
@@ -65,3 +67,20 @@
         </div>
     {% endfor %}
 {% endmacro %}
+
+
+{% macro showSectionProgress(sectionProgress) %}
+    {% if sectionProgress %}
+        <div class="application-summary__status__label u-padded-vertical-s">
```
davidrapson: Makes sense for this to be an external macro, but I'd try and pull the `application-summary__status__label` out of the macro to avoid having a BEM fragment in a macro.

models/application.js
```diff
@@ -36,7 +37,15 @@ module.exports = function(sequelize, DataTypes) {
             }
         },
         {
-            freezeTableName: true
+            freezeTableName: true,
+            getterMethods: {
+                grantAmount() {
+                    if (this.application_data && this.application_data['project-budget']) {
```
davidrapson: This makes a lot of sense _but_ I'm conscious that this is specific to the simple funding form, might not apply to _all_ applications. Just noting, not sure we know the answer to this just yet.

modules/filters.js
```diff
@@ -80,6 +80,10 @@ function timeago(date) {
     return moment(date).fromNow();
 }
 
+function addTime(date, amount, unit) {
```
davidrapson: Will this make me live longer?

`addTime(now()), '25', 'years')`

models/application.js
mattandrews: Yeah... will leave it there for now but definitely will come up in future.

modules/filters.js
mattandrews: Sadly all you've done there is made your age `3025` 👴

PR #1821, opened 2019-03-18
Add open graph tags from CMS

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/159

This allows passing `?social=<slug>` URLs to the CMS API in order to request custom Open Graph tags, and also falls back to default/generic OG markup (if specified) via the new CMS fields, replacing the generic sitewide ones we currently have.

This means we can create richer social media sharing cards from the CMS.


views/includes/metadata.njk
```diff
@@ -1,58 +1,67 @@
...
+            <meta name="twitter:card" content="summary" />
+        {% endif -%}
+
+        {% if openGraph.images.twitter -%}
+            <meta property="twitter:image" content="{{ openGraph.images.twitter | safe }}">
+        {% elseif socialImage -%}
+            <meta property="twitter:image" content="{{ getSocialImageUrl(socialImage) | safe }}">
+        {% endif %}
+
+        <meta name="google-site-verification" content="YJm5YK8XXPjyF0oyks3mvJ2P2qc1mwm9GyF7OlyNl9A" />
```
mattandrews: Everything below this is just a whitespace change – I changed the indentation so all of these fields look right, eg. not like this:

image

davidrapson: Why does indentation of the output matter? I don't feel that strongly about this, just curious.

mattandrews: It doesn't really, it just made it easier to scan it by eye when checking for updates etc.

davidrapson:
👍 To latest changes

Issue #1329, opened 2018-10-02
Update site metadata for contact number


mattandrews:
Eg. remove Welsh number from Google

mattandrews:
For me this actually returns the England number now:

![image](https://user-images.githubusercontent.com/394376/47712837-b8b1c880-dc30-11e8-9337-ff34242cc9d6.png)

We have an ancient (closed) ticket for page metadata: https://github.com/biglotteryfund/blf-alpha/issues/40 (eg. to try to override the descriptions Google generates). Could possibly be used here? Otherwise we might need to explore using microformats on that page (which may make it challenging for non-technical CMS users to update).

mattandrews:
... maybe we could add a microdata field for custom JSON?

PR #1824, opened 2019-03-21
Add JSON schema data for organisation details

mattandrews:
Fixes https://github.com/biglotteryfund/blf-alpha/issues/1329 

Hopefully this will encourage Google to display the generic switchboard number rather than any others.

### DIFF
--- a/views/includes/metadata.njk
+++ b/views/includes/metadata.njk
@@ -24,6 +24,21 @@
 <meta name="twitter:card" content="summary" />
 {% endif %}
 
+<script type="application/ld+json">
+    {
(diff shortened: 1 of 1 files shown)


views/includes/metadata.njk
```diff
@@ -24,6 +24,21 @@
 <meta name="twitter:card" content="summary" />
 {% endif %}
 
+<script type="application/ld+json">
+    {
```
davidrapson: Should this be https?

mattandrews: Oops! Yes.

PR #1826, opened 2019-03-22
Refine dashboard components

davidrapson:
Cleans up the dashboard. Extracts application card component and updates the template to match the latest designs (only shows status breakdown for non-submitted applications).

image

### DIFF
--- a/controllers/apply-next/views/dashboard.njk
+++ b/controllers/apply-next/views/dashboard.njk
@@ -1,157 +1,32 @@
-{% from "components/buttons.njk" import startButton, startButtonSubmit %}
+{% from "components/application-card/macro.njk" import applicationCard with context %}
 {% from "components/content-box/macro.njk" import contentBox %}
-{% from "components/form-header/macro.njk" import formHeader with context %}
-{% from "components/form-fields/macros.njk" import formErrors, formField with context %}
-{% from "components/icons.njk" import iconBin %}
-{% from "./view-helpers.njk" import showSectionProgress with context %}
 
 {% extends "layouts/main.njk" %}
 
-{% set fieldMissing %}<span class="application-data-missing">not yet completed</span>{% endset %}
-
-{% macro getStartDate(dateParts) %}
-    {% if dateParts.day and dateParts.month and dateParts.year %}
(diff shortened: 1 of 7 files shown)


views/components/application-card/macro.njk
```diff
@@ -0,0 +1,114 @@
...
+    {% else %}
+        Not submitted
+    {% endif %}
+{% endmacro %}
+
+{% macro applicationOverview(application) %}
+    <dl class="o-definition-list o-definition-list--compact u-margin-top">
+        <dt>Proposed start date</dt>
+        <dd>
+            {% set dateParts = application.application_data['project-start-date'] %}
```
davidrapson: This made me realise that we need a way to use the `displayFormat` logic for the form model for these application summaries too, as this is duplicating that logic in templates which is much harder to test.

mattandrews: Yeah... wonder how we can do that.

controllers/apply-next/views/dashboard.njk
```diff
@@ -1,157 +1,32 @@
...
+{% from "components/application-card/macro.njk" import applicationCard with context %}
 {% from "components/content-box/macro.njk" import contentBox %}
-{% from "components/form-header/macro.njk" import formHeader with context %}
-{% from "components/form-fields/macros.njk" import formErrors, formField with context %}
-{% from "components/icons.njk" import iconBin %}
-{% from "./view-helpers.njk" import showSectionProgress with context %}
 
 {% extends "layouts/main.njk" %}
 
-{% set fieldMissing %}<span class="application-data-missing">not yet completed</span>{% endset %}
```
mattandrews: Any reason for removing this? If it's no longer in the design then the CSS can go, too.

views/components/application-card/macro.njk
davidrapson: My thinking is that each "application" in this list should be a copy of the full form model rather than the raw application data, so the same process we go through to "enhance" the data in steps but per application.

controllers/apply-next/views/dashboard.njk
davidrapson: I think this is me being overeager, this should stay.

PR #1828, opened 2019-03-22
Form router cleanup

davidrapson:
The `helpers` file was getting pretty big so I've broken each part into `lib` and shifted some apply specific code that was in `modules` to live with the rest of the helper code.


controllers/apply-next/simple/fields.test.js
```diff
@@ -1,36 +0,0 @@
-/* eslint-env jest */
```
davidrapson: Removing this test as it tests at an awkward level, requiring another unit in order to be able to test the current one suggests that this test should be at a different level.

PR #1830, opened 2019-03-27
Add Insights Document section

mattandrews:
This needs some design/translation, but pairs with https://github.com/biglotteryfund/craft-dev/pull/164 to allow us to list insights documents and filter them.

### DIFF
--- a/assets/sass/components/_icons.scss
+++ b/assets/sass/components/_icons.scss
@@ -55,3 +55,8 @@ $spinnerColour: get-color('brand', 'primary');
         transform: rotate(360deg);
     }
 }
+
+.icon--download-btn {
+    position: relative;
+    top: 3px;
+}
--- a/assets/sass/components/_icons.scss
+++ b/assets/sass/components/_icons.scss
@@ -59,4 +59,12 @@ $spinnerColour: get-color('brand', 'primary');
 .icon--download-btn {
     position: relative;
     top: 3px;
+    height: 13px;
+}
+
+.icon--tick-plain {
+    width: 11px;
+    height: 11px;
+    margin-right: 3px;
+    fill: darken(get-color('border', 'base'), 30%);
 }
--- a/controllers/insights/index.js
+++ b/controllers/insights/index.js
@@ -11,14 +11,28 @@ const {
 } = require('../../middleware/inject-content');
 const { buildArchiveUrl } = require('../../modules/archived');
 const { localify } = require('../../modules/urls');
+const { buildPagination } = require('../../modules/pagination');
 
 const router = express.Router();
 
-router.get('/', injectHeroImage('insights-letterbox-new'), injectCopy('insights'), injectResearch, (req, res) => {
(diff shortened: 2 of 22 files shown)


controllers/insights/index.js
```diff
@@ -11,15 +11,29 @@ const {
...
     res.render(path.resolve(__dirname, './views/insights-landing'), {
         researchArchiveUrl: buildArchiveUrl(localify(req.i18n.getLocale())('/research'))
     });
 });
 
+router.get(
+    '/documents',
+    injectHeroImage('insights-letterbox-new'),
+    injectCopy('insights.documents'),
+    injectResearch('documents', 10),
```
mattandrews: Not sure if this is the best way to do this – could just make another middleware and control this there, perhaps?

controllers/insights/views/insights-documents.njk
```diff
@@ -0,0 +1,164 @@
+{% extends "layouts/main.njk" %}
+{% from "components/hero.njk" import hero with context %}
+{% from "components/content-box/macro.njk" import contentBox %}
+{% from "components/breadcrumb-trail/macro.njk" import breadcrumbTrail %}
+{% from "components/article-teaser/macro.njk" import articleTeaser %}
+{% from "components/split-nav/macro.njk" import splitNav %}
+{% from "components/card/macro.njk" import card %}
+
+{% from "../../updates/views/view-helpers.njk" import entryRegions with context %}
```
mattandrews: This feels a bit ugly – do you think we need a more general shared helpers file/folder?

assets/sass/components/_icons.scss
```diff
@@ -55,3 +55,16 @@ $spinnerColour: get-color('brand', 'primary');
         transform: rotate(360deg);
     }
 }
+
+.icon--download-btn {
```
davidrapson: Should this be scoped more? Should _all_ download buttons be positioned this way or just for this use case? Same for the icon-tick styles.

controllers/insights/index.js
davidrapson: I'm starting to find the `inject*` pattern a little awkward in cases where we only use it in a single controller. One two many levels of abstraction. It might be easier to inline the content-api calls into this method. Otherwise yeah a separate middleware would work.

controllers/insights/views/insights-documents.njk
davidrapson: I'd say that `views/components` is exactly for this 😄 . If we need to use this across routers can this become a component?

mattandrews: Funnily enough I ended up not using this component here, but I did move `facetGroup` to its own component as it's now shared between here and past grants search.

controllers/insights/index.js
mattandrews: Done!

assets/sass/components/_icons.scss
mattandrews: I think so, until they get used anywhere else? These styles are to basically make the icons usable, eg. align them optically and size them correctly.

PR #1835, opened 2019-03-29
Add flexible content support for information/listing pages

mattandrews:
This allows us to use flexible content for listing pages which behave in the same way as the segments currently do. For now the template will display both types of content, but we could migrate away from one in favour of the other. Probably worth demoing this in person!

image

Also introduces a new block type for Flexible Content (see [corresponding CMS PR here](https://github.com/biglotteryfund/craft-dev/pull/167)):

image

### DIFF
--- a/controllers/common/views/information-page.njk
+++ b/controllers/common/views/information-page.njk
@@ -3,9 +3,12 @@
 {% from "components/content-box/macro.njk" import contentBox %}
 {% from "components/segment-links/macro.njk" import segmentLinks %}
 {% from "components/segment/macro.njk" import segment %}
+{% from "components/flexible-content/macro.njk" import flexibleContentPart %}
 
 {% extends "layouts/main.njk" %}
 {% if not pageHero.image %}{% set bodyClass = 'has-static-header' %}{% endif %}
+
+{% set flexAnchor = 'item' %}
 
 {% block content %}
     <main id="content">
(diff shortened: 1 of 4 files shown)


controllers/common/views/information-page.njk
```diff
@@ -41,6 +47,16 @@
                     {% endfor %}
                 </div>
             {% endif %}
+
+            {# Flexible content #}
+            {% if content.flexibleContent %}
+                {% for part in content.flexibleContent %}
```
davidrapson: This feels a little awkward here that each flexible content part is a _new_ content area, the way we use this elsewhere is that flexible content is all within a _single_ content block.

views/components/flexible-content/macro.njk
```diff
@@ -1,39 +1,56 @@
 {% from "components/callout/macro.njk" import callout %}
 {% from "components/media-aside/macro.njk" import mediaAside %}
 
+{% macro flexibleContentPart(part) %}
```
davidrapson: Not the `mediaAsideCounter` logic probably doesn't work here now this has been extracted out.

mattandrews: Have fixed this now using the nifty `cycler` tool in Nunjucks, which only calls its next parameter when `next()` is triggered.

image

davidrapson: Nice

PR #1855, opened 2019-04-09
Refactor conditional fields

davidrapson:
This PR does some work to refactor how conditional steps and fields are handled.

The biggest change is to move the conditional logic from steps down to individual fields. This allows us to handle conditional fields within a step (e.g. not asking for date of birth and address for school contacts) as well as whole conditional steps (e.g. registration numbers).

Still a little work in progress, needs more unit tests to validate these changes.

### DIFF
--- a/controllers/apply-next/lib/pagination.js
+++ b/controllers/apply-next/lib/pagination.js
@@ -7,22 +7,17 @@ const { findIndex, findLastIndex } = require('lodash');
  * @property {Array} sections
  * @property {Number} currentSectionIndex
  * @property {Number} currentStepIndex
- * @property {Object} formData
  */
 
 /**
  * Find next matching URL
  * @param {MatchOptions} options
  */
-function findNextMatchingUrl({ baseUrl, sections, currentSectionIndex, currentStepIndex, formData }) {
+function findNextMatchingUrl({ baseUrl, sections, currentSectionIndex, currentStepIndex }) {
     const currentSection = sections[currentSectionIndex];
     const nextSection = sections[currentSectionIndex + 1];
 
-    const targetStepIndex = findIndex(
-        currentSection.steps,
-        step => (step.matchesCondition ? step.matchesCondition(formData) === true : true),
-        currentStepIndex + 1
-    );
+    const targetStepIndex = findIndex(currentSection.steps, step => step.isRequired === true, currentStepIndex + 1);
 
     if (targetStepIndex !== -1 && targetStepIndex <= currentSection.steps.length) {
         return `${baseUrl}/${currentSection.slug}/${targetStepIndex + 1}`;
@@ -37,14 +32,14 @@ function findNextMatchingUrl({ baseUrl, sections, currentSectionIndex, currentSt
  * Find previous matching URL
  * @param {MatchOptions} options
  */
-function findPreviousMatchingUrl({ baseUrl, sections, currentSectionIndex, currentStepIndex, formData }) {
+function findPreviousMatchingUrl({ baseUrl, sections, currentSectionIndex, currentStepIndex }) {
     const currentSection = sections[currentSectionIndex];
     const previousSection = sections[currentSectionIndex - 1];
 
     if (currentStepIndex !== 0) {
         const targetStepIndex = findLastIndex(
             currentSection.steps,
-            step => (step.matchesCondition ? step.matchesCondition(formData) === true : true),
+            step => step.isRequired === true,
             currentStepIndex - 1
         );
         return `${baseUrl}/${currentSection.slug}/${targetStepIndex + 1}`;
--- a/controllers/apply-next/simple/fields.js
+++ b/controllers/apply-next/simple/fields.js
@@ -1,5 +1,5 @@
 'use strict';
-const { get, has, reduce, values } = require('lodash');
+const { get, includes, reduce, values } = require('lodash');
 const moment = require('moment');
 
 const { Joi, ...commonValidators } = require('../lib/validators');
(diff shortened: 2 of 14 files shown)


controllers/apply-next/lib/pagination.js
```diff
@@ -37,14 +32,14 @@ function findNextMatchingUrl({ baseUrl, sections, currentSectionIndex, currentSt
...
  */
-function findPreviousMatchingUrl({ baseUrl, sections, currentSectionIndex, currentStepIndex, formData }) {
+function findPreviousMatchingUrl({ baseUrl, sections, currentSectionIndex, currentStepIndex }) {
     const currentSection = sections[currentSectionIndex];
     const previousSection = sections[currentSectionIndex - 1];
 
     if (currentStepIndex !== 0) {
         const targetStepIndex = findLastIndex(
             currentSection.steps,
-            step => (step.matchesCondition ? step.matchesCondition(formData) === true : true),
```
davidrapson: Avoids duplicating matching logic here and directly checks the new `step.isRequired` property.

controllers/apply-next/views/step.njk
```diff
@@ -35,26 +35,27 @@
                 {{ formErrors(errors = errors) }}
 
                 {% for fieldset in step.fieldsets %}
-                    <fieldset class="form-fieldset u-constrained-wide">
-                        <legend class="form-fieldset__legend t2 t--underline">
-                            {{ fieldset.legend }}
-                        </legend>
+                    {% if fieldset.fields.length > 0 %}
```
davidrapson: This is whitespace heavy, but I'm wrapping this in a `fieldset.fields.length > 0` check to avoid rendering fieldsets with no fields in them. Maybe I should be filtering this out in `enhanceForm` rather than in the view?

views/components/form-fields/macros.njk
```diff
@@ -204,7 +204,7 @@
                                 name="{{ field.name }}"
                                 value="{{ option.value }}"
                                 {% if field.isRequired %}required aria-required="true"{% endif %}
-                                {% if field.value[0] === option.value %}checked{% endif %}
+                                {% if field.value[0] === option.value or field.value === option.value %}checked{% endif %}
```
davidrapson: This is to account for cases where the option is stored as a single value, not an array.

controllers/apply-next/simple/fields.js
```diff
@@ -538,39 +573,58 @@ const allFields = {
...
         },
         type: 'radio',
-        options: legalContactRoles,
-        isRequired: true,
-        get schema() {
-            return Joi.array()
-                .items(Joi.string().valid(this.options.map(option => option.value)))
-                .single()
-                .optional();
+        options(formData) {
```
davidrapson: I've allowed options to (optionally) be a function to allow us to differ the options based on the form data. Previously I had this as a function on each option. This approach allows us to call a single function rather than one for each option.

controllers/apply-next/simple/fields.js
```diff
@@ -405,20 +438,44 @@ const allFields = {
...
         type: 'text',
-        attributes: {
-            size: 20
-        },
-        isRequired: true,
-        schema: Joi.any().when('organisation-type', {
-            is: Joi.valid(organisationTypes.unincorporatedRegisteredCharity.value),
-            then: Joi.number().required()
+        attributes: { size: 20 },
+        isRequired(formData = {}) {
```
davidrapson: I've allowed `isRequired` to be a function so that we can conditionally determine whether a field should be marked as required on the front-end. Duplicating the field didn't work as the two schema's clashed with one another.

controllers/apply-next/simple/fields.js
```diff
@@ -538,39 +573,58 @@ const allFields = {
...
-                .single()
-                .optional();
+        options(formData) {
+            let options = [];
+            switch (get(formData, 'organisation-type')) {
+                case organisationTypes.unregisterdVco.value:
+                    options = [
+                        { value: 'chair', label: { en: 'Chair', cy: '' } },
+                        { value: 'vice-chair', label: { en: 'Vice-chair', cy: '' } },
+                        { value: 'cecretary', label: { en: 'Secretary', cy: '' } },
```
mattandrews: Typo in "cecretary"!

controllers/apply-next/simple/fields.js
```diff
@@ -15,76 +15,49 @@ const countries = {
 };
 
 const organisationTypes = {
-    constitutedVoluntaryCommunity: {
-        value: 'constituted-voluntary-community',
-        label: { en: 'Constituted voluntary or community organisation', cy: '' },
-        explanation: { en: 'i.e. not registered as a company or charity', cy: '' }
+    unregisterdVco: {
```
mattandrews: Typo here ("unregisterdVco" should be "unregisteredVco")

controllers/apply-next/views/step.njk
mattandrews: Hmm... possibly? There are a few places where we loop over the entire form (summary, email, new questions endpoint)... would be neater not to have to repeat this check multiple times.

controllers/apply-next/views/summary.njk
```diff
@@ -71,7 +71,7 @@
                                         Step {{ stepNumber }} of {{ section.steps.length }} &ndash; {{ step.title }}
                                     </h4>
 
-                                    {% if not step.notRequired %}
+                                    {% if step.isRequired %}
```
mattandrews: 💯 for readability!

controllers/apply-next/views/step.njk
davidrapson: 👍 I've moved this up into `enhanceForm` now

PR #1857, opened 2019-04-10
Add questions summary for AFA, plus PDF download

mattandrews:
image

Also downloads a basically-formatted PDF which looks like this:

image

I'm curious about how performant the PDF rendering will be on a real server (takes a few seconds locally) and whether it puts much strain on the server. If it does, we could explore pre-rendering this PDF when deploying, as it's not dynamic (eg. contains no user data). Thoughts welcome!


controllers/apply-next/views/questions-list.njk
```diff
@@ -0,0 +1,110 @@
...
+        </div>
+        <div class="u-padded">
+            {% for step in section.steps %}
+                <h4 class="t--underline{% if not loop.first %}  u-margin-top-l{% endif %}">
+                    Step {{ loop.index }}: {{ step.title }}
+                </h4>
+                {% for fieldset in step.fieldsets %}
+                    {% for field in fieldset.fields %}
+                        <div class="{% if not loop.first %}u-padded-top u-border-top--thin u-margin-top{% endif %}">
+                            <h5>{{ field.label }}</h5>
```
davidrapson: Not sure headings are the appropriate markup here, is this a definition list of sorts?

controllers/apply-next/views/questions-list.njk
```diff
@@ -0,0 +1,110 @@
...
+    </div>
+{% endmacro %}
+
+{% macro question(first = false) %}
+    <div class="{% if not first %}u-padded-top u-border-top--thin u-margin-top{% endif %}">
+        {{ caller() }}
+    </div>
+{% endmacro %}
+
+{% macro questionList(form) %}
```
davidrapson: I'd be tempted to style this up using BEM class names rather than lots of utility classes. Not sure if this is just to get something on the page for now?

.editorconfig
```diff
@@ -9,5 +9,4 @@ indent_size = 4
 max_line_length = 120
 
 [{*.yml,*.json}]
-indent_style = space
```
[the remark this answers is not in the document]
mattandrews: Yep, IntelliJ gave me a warning that it's redundant due to the `[*]` selector above it. Not directly related to this PR I guess but did some tidying up as I was going through.

views/layouts/emails.njk
```diff
@@ -5,10 +5,16 @@
         <style>
             @import url('https://use.typekit.net/yyc1dgh.css');
         </style>
-        <link rel="stylesheet" type="text/css" href="/build/latest/stylesheets/emails.css">
+        {% if appData.isDev %}
+            <link rel="stylesheet" type="text/css" href="{{ 'stylesheets/emails.css' | getCachebustedPath }}">
```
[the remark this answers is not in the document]
mattandrews: Yes (we pass through all the `res.locals` and globals in the places where we render nunjucks templates directly). The point of this isn't the cachebust, it's to get a path that works locally (eg. begins with `/assets/build` rather than `/build`) so that when testing locally, the Nunjucks/PDF renderer can find the stylesheets.

controllers/apply-next/views/questions-list.njk
mattandrews: If you like – I thought each question would be a subheading in the document outline, but I'll turn it into DLs/DDs.

controllers/apply-next/views/questions-list.njk
mattandrews: Do you think it's better for this stuff to have its own classnames that are just duplicating those utility classes underneath? Happy to rewrite them all if that's the case, but otherwise why have the `u-` classes?

controllers/apply-next/views/questions-list.njk
davidrapson: Ah, I think I see. In which case it might just be the help text that feels semantically wrong, it should probably be a plain paragraph rather than a `<h6>`. The help text shouldn't create a new heading-level should it?

controllers/apply-next/views/questions-list.njk
davidrapson: My general school of thought on this is that if something can be named as a concept I'd err towards dedicated BEM classes. And to use utility classes mostly for exceptions. I could go either way in this case as it's not using _mountains_ of utility classes to achieve this layout.

mattandrews: Have tried to address this – see latest commits.

assets/sass/components/_applications.scss
```diff
@@ -198,3 +198,7 @@ $statusColumnWidth: 160px;
     }
     color: lighten(get-color('text', 'base'), 50%);
 }
+
+.application-text-minor {
```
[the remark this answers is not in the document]
mattandrews: 👍 forgot about those classes! Added a new one, though, for this smaller size.

controllers/apply-next/views/questions-list.njk
mattandrews: Went with DLs in the end!

davidrapson:
👍 Latest changes look good to me

mattandrews:
image

Here's how it looks now.

controllers/apply-next/form-router.js
```diff
@@ -36,6 +40,82 @@ function initFormRouter(formModel) {
...
+                            }
+
+                            // Write the file locally so we can look it up next time instead of rendering
+                            fs.writeFile(filePath, buffer, writeError => {
+                                if (writeError) {
+                                    Raven.captureException(writeError);
+                                    return res.status(400).json({ error: 'ERR-PDF-WRITE-ERROR' });
+                                }
+                                // Give the user the file directly
+                                res.setHeader('Content-Disposition', `attachment; filename=${formModel.id}.pdf`);
```
davidrapson: I think `res.download` does this for you https://expressjs.com/en/4x/api.html#res.download

PR #1859, opened 2019-04-11
Apply common body classes separately to fonts-loaded

davidrapson:
Noticed yesterday that if fonts fail to load we don't replace the `no-js` class with `js-on` class. We don't use this class in many places (we largely use `v-cloak` instead) but it effectively couples font-loading with the display of some javascript functionality.

I've reworked the `head.js` file to handle this class separately to the `fonts-loading` class.

### DIFF
--- a/views/includes/metaHead.njk
+++ b/views/includes/metaHead.njk
@@ -9,11 +9,6 @@
         isNotProduction: {{ appData.isNotProduction }},
         locale: '{{ locale }}',
         localePrefix: '{{ localePrefix }}',
-        isModernBrowser:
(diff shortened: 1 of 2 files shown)


views/includes/metaHead.njk
```diff
@@ -9,11 +9,6 @@
         isNotProduction: {{ appData.isNotProduction }},
         locale: '{{ locale }}',
         localePrefix: '{{ localePrefix }}',
-        isModernBrowser:
```
davidrapson: We don't use this check or the corresponding CSS class anywhere so I've removed the check.

PR #1860, opened 2019-04-11
Document organisation type rules

davidrapson:
This adds documentation for the organisation type, registration numbers, and role questions in their current state.

### DIFF
--- a/controllers/apply-next/docs/validation-rules.md
+++ b/controllers/apply-next/docs/validation-rules.md
@@ -125,11 +125,23 @@ Address field group, consists of:
 
 ### Company number
 
-@TODO
+**Type**: string
+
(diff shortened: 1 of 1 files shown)


controllers/apply-next/docs/validation-rules.md
```diff
@@ -225,12 +241,31 @@ Address field group, consists of:
...
-@TODO
+**Type**: single choice (radio)
+
+| Rule           | Message       |
+| -------------- | ------------- |
+| Required field | Choose a role |
+
+#### Mappings
+
+The legal contact role shows the following choices depending on the organisation type:
```
davidrapson: I've added a trello card based to this list so we can refine it. I'm not 100% sure I've interpreted the rules for this correctly as it feels off that there is only one choice of role.

PR #1861, opened 2019-04-11
Update API endpoints to use joi schema

davidrapson:
This PR reworks the API endpoints for surveys and feedback forms to use `joi` for validation instead of `express-validator`. Gives a low-impact opportunity to use this validation in production.

### DIFF
--- a/controllers/api/index.js
+++ b/controllers/api/index.js
@@ -1,106 +1,98 @@
 'use strict';
 const express = require('express');
+const Joi = require('joi');
 
 const router = express.Router();
 
-const { pick } = require('lodash');
-const { body, validationResult } = require('express-validator/check');
-const { matchedData } = require('express-validator/filter');
-
-const { purifyUserInput } = require('../../modules/validators');
-const surveyService = require('../../services/surveys');
 const feedbackService = require('../../services/feedback');
+const surveyService = require('../../services/surveys');
 
 /**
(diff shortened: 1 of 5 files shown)


controllers/api/index.js
```diff
@@ -1,106 +1,98 @@
...
+
+    if (validationResult.error) {
+        res.status(400).json({
+            status: 'error',
+            message: messageError,
+            err: validationResult.error.message
+        });
+    } else {
+        try {
+            const result = feedbackService.storeFeedback({
```
mattandrews: Should this be `await feedbackService...`?

controllers/api/index.js
```diff
@@ -1,106 +1,98 @@
...
+router.post('/survey', async (req, res) => {
+    const schema = Joi.object({
+        choice: Joi.string()
+            .valid(['yes', 'no'])
+            .required(),
+        path: Joi.string().required(),
+        message: Joi.string().optional()
+    });
+
+    const valdationResult = schema.validate(req.body, {
```
mattandrews: Typo on `valdationResult`...

controllers/api/index.js
davidrapson: Ack, good catch. I've updated this and also added some integration tests for both API endpoints to assert the value of the result in the response.

PR #1865, opened 2019-04-15
Travis merge jobs

davidrapson:
Following on from https://github.com/biglotteryfund/blf-alpha/pull/1864 this PR explores (re)merging the separate build and deploy jobs back together.

I don't know if Travis have adjusted their build lifecycles or if it's something else but the deploy dependencies no longer seem to get installed if the build is not deployable (i.e. merge/PR build) so on first pass this looks like it will speed up the builds even further as we a) no longer have spin up two jobs for `master` builds and b) avoid double building and can do a single production build up front.

We'll need to merge and run a test build on `master` to confirm the speed improvements are real, but looks promising.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,6 @@ env:
 
 dist: xenial
 
-# Build stage conditions
-stages:
-  - test
-  # https://github.com/travis-ci/travis-ci/issues/9614#issuecomment-388798396
-  - name: build
-    if: (NOT type IN (pull_request)) AND (branch = master)
-
 # Defaults
 language: node_js
 node_js: '10'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -30,53 +23,47 @@ cache:
...
-          secret_access_key: $AWS_SECRET_ACCESS_KEY
-          bucket: blf-deployments
-          key: build-$TRAVIS_BUILD_NUMBER.zip
-          bundle_type: zip
-          application: BLF_WWW
-          deployment_group: Test_Fleet
-          on:
-            tags: false
+script:
+  - npm run build-production
```
davidrapson: Small note here: Arguably unit tests (`npm test`) should come before the build but we have a single unit test (RC email tests) which require a build due to the way they assert email templates. This might be an appropriate trade-off but feels like a bit of smell, something to revisit.

PR #1867, opened 2019-04-15
Rework processor mail tests

davidrapson:
This PR reworks the (flaky) reaching communities email tests. Rather than snapshotting the whole node mailer object inline styles and all I've reworked the tests to use the mock JSON transport which allows us to snapshot the text version of the email and then perform a few key assertions on the html output.

This should make this test more reliable without sacrificing too much confidence in the email output.

(Note: For now I've removed the equivalent digital fund tests as this programme is no longer open)


mattandrews:
Excellent, this should help me get https://github.com/biglotteryfund/blf-alpha/pull/1857 merged!

PR #1869, opened 2019-04-16
Form order

davidrapson:
This restructures the order of steps when creating a  new application so that the eligibility checker is available without logging in, and when completed as part of  a new application is shown _before_ any questions are asked rather than after the project name / country questions. All questions are now part of the main form steps.

The first step now looks like this:

Screenshot 2019-04-16 at 10 16 25

There's a question here as to whether we should split this out into two steps but for now I've kept everything on a single screen.


controllers/apply-next/form-router.js
```diff
@@ -116,32 +139,60 @@ function initFormRouter(formModel) {
...
-     * Set common locals
+     * Route: Eligibility checker
      */
-    router.use(async (req, res, next) => {
-        // Translate the form object for each request and populate it with current user input
-        const form = enhanceForm({
-            locale: req.i18n.getLocale(),
-            baseForm: formModel
-        });
+    router
```
davidrapson: The eligibility step now no longer requires people to be logged in. It also no longer stores progress in the session, so acts as a soft barrier to entry rather than being hard enforced.

controllers/apply-next/form-router.js
```diff
@@ -170,80 +221,54 @@ function initFormRouter(formModel) {
...
-        .post(async (req, res) => {
-            const validationResult = formModel.schema.validate(req.body, {
-                abortEarly: false,
-                stripUnknown: true,
-                escapeHtml: true
+    /**
+     * New application
+     * Create a new blank application and redirect to first step
+     */
+    router.get('/new', async function(req, res) {
```
davidrapson: Rather than rendering an interstitial page the `new` route creates a new empty application and proceeds to the first step.

cypress/integration/integration_spec.js
```diff
@@ -142,37 +205,21 @@ describe('user', () => {
...
-
-Enim provident necessitatibus ipsa ad autem aliquam ducimus minima delectus exercitationem, minus blanditiis molestias quas eaque ullam ab aperiam assumenda.`;
-
-describe.skip('awards for all', function() {
-    it('happy path application', () => {
-        const eligibilityStep = () => {
-            cy.getByLabelText('Yes').click();
-            cy.getByText('Continue').click();
-        };
+describe('awards for all', function() {
```
davidrapson: This is the main change to these tests. Including the awards for all test rather than skipping it. It only goes part way through the form but even with this addition (and a little streamlining of some of the other tests) the CI test run is still <5 minutes, with the integration tests taking about 1m30 of that so I think we can get some decent coverage of the form here without bloating the build time too much.

models/application.js
```diff
@@ -23,22 +23,17 @@ module.exports = function(sequelize, DataTypes) {
...
-            },
             application_data: {
                 type: DataTypes.JSON,
                 allowNull: true
             },
             status: {
                 type: DataTypes.ENUM,
-                values: ['ineligible', 'eligible', 'pending', 'complete'],
-                allowNull: true
+                values: ['pending', 'complete'],
```
mattandrews: This might involve a manual step when merging to `TEST` as I don't know if Sequelize is able to automatically migrate this kind of thing, just FYI.

mattandrews: Easiest thing might be to rename the old column, manually make a new one with the right config, then merge this PR so it doesn't fail/complain? Though if it works locally then ignore me.

davidrapson: Sorry should've called out that I removed the `freezeTableName` setting here so this actually creates a _new_ table called `applications` rather than the current `application`. I checked and the former doesn't exist on either the test or live databases so this would create a brand new table.

This is a good broader point though that we'll soon need a better strategy for migrating database schema changes, but for now this is essentially starting from scratch for this table.

PR #1870, opened 2019-04-16
Concurrent tests in CI

davidrapson:
Updates the build process to run the asset build concurrently as well as running unit and integration tests in parallel in CI. This seems to shave off another ~20-30 seconds from the build, running for about ~4m15s. Needs a couple more test runs to confirm.

As we're soon going to push the build times up by expanding the coverage for the awards for all form I'm keen to squeeze the baseline build down as far as I can so we can still aim for a ~5 minute PR build time even with the new forms.


davidrapson:
Ran a couple of builds now and seems to generally hit the 4m15/4m20 mark so I think we're good.

PR #1872, opened 2019-04-16
Add a User dashboard

mattandrews:
Quite a bit going on here.

### Adds breadcrumbs to all the user screens, eg:

image

### Sends all emails using our HTML format:
image

### Adds new functionality
image
(eg. changing password/email address)

Plus a general refactor (eg. merging password controllers and adding a helper function for HTML emails)


views/layouts/emails.njk
```diff
@@ -5,15 +5,12 @@
...
-        {% else %}
-            <link rel="stylesheet" type="text/css" href="/build/latest/stylesheets/emails.css">
-        {% endif %}
+        <link rel="stylesheet" type="text/css" href="/build/latest/stylesheets/emails.css">
     </head>
     <body>
         <div class="logo">
-            <img src="{% if appData.isDev %}/assets{% endif %}/images/site-logo.svg" alt="" />
+            {# We hard-code this path to avoid having to pass functions in for assembling a local path, and to prevent HTML parsers from inlining this image as a data URI #}
+            <img src="/images/logos/email.gif" alt="The National Lottery Community Fund logo" />
```
mattandrews: This caused me lots of debugging pain. The [changes I introduced last week](https://github.com/biglotteryfund/blf-alpha/commit/003d8fc8a860ac85c0cb1ef7c8275693c8553eb0#diff-a8ac64883753b8f2e84be36c6cdf979eR16) broke this, d'oh. Basically, `juice`, which inlines our CSS for emails, does some magic to rewrite paths (so the CSS change wasn't needed), but I discovered that when it inlines SVGs, this doesn't work on GMail... or apparently [most other providers](https://css-tricks.com/a-guide-on-svg-support-in-email/#svg-email-inline). I opted instead for a gif which should always work.

controllers/user/dashboard.js
```diff
@@ -2,13 +2,104 @@
...
+        formValues: data,
+        errors: errors
+    });
+}
+
+/**
+ * Route: Generic user dashboard
+ */
 router.get('/', (req, res) => {
+    let alertMessage;
```
davidrapson: :neckbeard: It'd be nice to extract this into a function and add a test for it.

controllers/user/dashboard.js
```diff
@@ -2,13 +2,104 @@
...
+            renderUpdateEmailForm(req, res, validationResult.value, errors);
+        } else {
+            try {
+                const { username } = validationResult.value;
+                const existingUser = await userService.findByUsername(username);
+                if (existingUser) {
+                    throw new Error('A user tried to update their email address to an existing email address');
+                } else {
+                    const userId = req.user.userData.id;
+                    await userService.updateNewEmail({
```
davidrapson: Does the update method return the user object? Do we need to call findById again?

controllers/user/index.js
```diff
@@ -10,13 +11,28 @@ router.use(noCache, noindex, (req, res, next) => {
...
+router.use((req, res, next) => {
+    let crumbs = [
+        {
+            label: 'Your account',
+            url: req.baseUrl
+        }
+    ];
+    if (req.user) {
+        crumbs = concat(crumbs, {
+            label: req.user.userData.username
```
davidrapson: Not sure we should put the email address in the breadcrumb,

controllers/user/password.js
```diff
@@ -0,0 +1,291 @@
+'use strict';
```
davidrapson: Nice call putting all the password reset code into one controller.

services/mail.js
```diff
@@ -62,6 +62,34 @@ function generateHtmlEmail({ template, templateData }) {
...
+ * Utility function to generate an HTML email and send it.
+ *
+ *
+ * @param {object} options
+ * @param {string} options.template
+ * @param {object} options.templateData
+ * @param {object} mailParams
+ * @return {object}
+ */
+async function sendHtmlEmail({ template, templateData }, mailParams) {
```
davidrapson: Nice, could you add a test for this? It might be more a case of reworking some of the current ones to test this higher-level function instead of the constituent parts.

views/layouts/emails.njk
davidrapson: 👍 Makes sense and good catch. Should/could this be a JPG instead of a GIF?

controllers/user/views/update-email.njk
```diff
@@ -0,0 +1,44 @@
...
+            {% endif %}
+
+            <form action="" method="post" autocomplete="off">
+                {{ formErrors(errors = errors, title = "There was a problem with your submission") }}
+
+                {% if csrfToken %}
+                    <input type="hidden" name="_csrf" value="{{ csrfToken }}">
+                {% endif %}
+
+                {{ formField({
```
davidrapson: Is this enough to ask for to change the email address? What's an appropriate pattern here, should we be asking to for password confirmation to do this? We can come back to this in a later PR.

controllers/user/dashboard.js
mattandrews: Annoyingly it [doesn't](https://stackoverflow.com/questions/38524938/sequelize-update-record-and-return-result) (except for PostgreSQL), hence the `findById` call.

davidrapson: 👍

controllers/user/index.js
mattandrews: Good call, moved it to a macro in all the templates instead.
image

services/mail.js
mattandrews: Done – give it a look now.

davidrapson:
Closed in favour of https://github.com/biglotteryfund/blf-alpha/pull/1918

PR #1876, opened 2019-04-18
Add a function to alert when the form data has changed but is not saved

mattandrews:
This change reuses some of the `beforeunload` functionality to warn the user if they attempt to navigate away from a form page when they've made changes to the data.

warning

### DIFF
--- a/assets/js/modules/forms.js
+++ b/assets/js/modules/forms.js
@@ -4,7 +4,7 @@ import { trackEvent } from '../helpers/metrics';
 function handleBeforeUnload(e) {
     // Message cannot be customised in Chrome 51+
     // https://developers.google.com/web/updates/2016/04/chrome-51-deprecations?hl=en
-    trackEvent('Apply', 'Review step abandonment check', 'message shown');
+    trackEvent('Apply', 'User warned before abandoning form changes', 'message shown');
     const confirmationMessage = 'Are you sure you want to leave this page?';
     e.returnValue = confirmationMessage; // Gecko, Trident, Chrome 34-51
     return confirmationMessage; // Gecko, WebKit, Chrome <34
@@ -27,7 +27,7 @@ function handleAbandonmentMessage(formEl) {
     formEl.addEventListener('submit', removeBeforeUnload);
 
     window.addEventListener('unload', function() {
-        recordUnload && trackEvent('Apply', 'Review step abandonment check', 'left page');
+        recordUnload && trackEvent('Apply', 'User warned before abandoning form changes', 'left page');
     });
 }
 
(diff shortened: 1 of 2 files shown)


assets/js/modules/forms.js
```diff
@@ -1,16 +1,17 @@
 import $ from 'jquery';
 import { trackEvent } from '../helpers/metrics';
 
+function handleBeforeUnload(e) {
+    // Message cannot be customised in Chrome 51+
+    // https://developers.google.com/web/updates/2016/04/chrome-51-deprecations?hl=en
+    trackEvent('Apply', 'Review step abandonment check', 'message shown');
```
davidrapson: We might want to change the `Review step abandonment check` message to be something more generic now

assets/js/modules/forms.js
```diff
@@ -30,6 +31,28 @@ function handleAbandonmentMessage(formEl) {
...
+ * Warns the user before leaving a page if the form's state is different from when the page was loaded.
+ *
+ * Caveats: form must have the class below and the field must lose focus to trigger the change() event.
+ * */
+function warnOnUnsavedChanges() {
+    let formHasChanged = false;
+    $('.js-form-warn-unsaved').each(function() {
+        const $form = $(this);
+        const initialState = $form.serialize();
+        $(':input', $form).change(function() {
```
davidrapson: Curious if rather than adding a change listener to every form element is it possible to serialise the form _within_ `beforeunload` method and do the check there, or does the return value for `beforeunload` always get used for the prompt?

assets/js/modules/forms.js
```diff
@@ -30,6 +31,28 @@ function handleAbandonmentMessage(formEl) {
...
+            const newState = $form.serialize();
+            formHasChanged = newState !== initialState;
```
davidrapson: Do we need the `formHasChanged` variable, could we inline the expression into the if statement?

mattandrews: Done!

assets/js/modules/forms.js
mattandrews: 👍

assets/js/modules/forms.js
mattandrews: Tried this – was kind of possible (`beforeunload` can return `false` which stops the message appearing), but meant we had to always bind `beforeunload` (eg. as soon as the page loads) which Chrome complains about, and meant immediately overriding it (eg. for form submits etc) so felt a bit circular.

davidrapson: I did wonder when I asked that if it was even possible, beforeunload is a thorny one, thanks for checking though.

PR #1890, opened 2019-04-25
Update validation-rules.md

scottoakley:
Hope this is all ok, but give me a shout if there's any questions. 

I've been using the GOV.UK design guide for error messages as a reference for this: https://design-system.service.gov.uk/components/error-message/

Also, I think the 'Total amount is over £10,000' error message might sit more naturally beside the 'Total' field in the form - otherwise it appears next to the 'Too many rows' error message if both appear at the same time, and is less visible. Just a suggestion!

### DIFF
--- a/controllers/apply-next/docs/validation-rules.md
+++ b/controllers/apply-next/docs/validation-rules.md
@@ -16,10 +16,10 @@ This file documents validation rules, conditions, and error message text for all
 
 **Type**: date
 
-| Rule                              | Message                                        |
-| --------------------------------- | ---------------------------------------------- |
-| Required field                    | Enter a date                                   |
-| Must be a valid date              | Enter a real date                              |
+| Rule                              | Message                                                                    |
+| --------------------------------- | -------------------------------------------------------------------------- |
+| Required field                    | Enter a date                                                               |
+| Must be a valid date              | Enter a real date                                                          |
 | At least 12 weeks into the future | Date you start the project must be after 25 July 2019/12 WEEKS INTO FUTURE |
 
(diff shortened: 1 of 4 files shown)


davidrapson:
> Also, I think the 'Total amount is over £10,000' error message might sit more naturally beside the 'Total' field in the form - otherwise it appears next to the 'Too many rows' error message if both appear at the same time, and is less visible. Just a suggestion!

Yeah, I was thinking the same thing, good call.

controllers/apply-next/docs/validation-rules.md
```diff
@@ -16,11 +16,11 @@ This file documents validation rules, conditions, and error message text for all
...
-| Rule                              | Message                                        |
-| --------------------------------- | ---------------------------------------------- |
-| Required field                    | Enter a date                                   |
-| Must be a valid date              | Enter a real date                              |
-| At least 12 weeks into the future | Date must be at least 12 weeks into the future |
+| Rule                              | Message                                                                    |
+| --------------------------------- | -------------------------------------------------------------------------- |
+| Required field                    | Enter a date                                                               |
+| Must be a valid date              | Enter a real date                                                          |
+| At least 12 weeks into the future | Date you start the project must be after 25 July 2019/12 WEEKS INTO FUTURE |
```
davidrapson: Is the "12 WEEKS INTO FUTURE" part meant to be included / in all-caps? Or is this just indicating that the date is dynamic?

scottoakley: Glad to hear it. The all caps is just to indicate that the date is dynamic

PR #1891, opened 2019-04-26
Form model restructure 

davidrapson:
As we've been iterating on the form we've been identifying more and more places where the form model (conditions, help text, validations etc.) are dependant on the in-progress form-data.

We've previously been handling this using an `enrichForm` method which did a mix of work to localise messages, apply conditions, and filter data.

We also have a mixed bag of code which uses the "base" form model and code which uses the "enriched" form model and both end up doing disparate work.

The more I looked at the problem the more natural it felt to have a single form model which is function of the locale and existing form data. This allows us to vary any part of the form based on previous answers and also has the side benefit of letting us handle form translations, rather than pass fields around for later translation.

It's also helped with resolving what level to write tests at. Previously I had some tests for the fields and some tests for the model as a whole and a set of tests like "shouldShow" which crossed the boundaries, and none for "enrichForm". Having a single model allows us to test the full form state directly.


controllers/apply-next/lib/enrich-form.js
```diff
@@ -0,0 +1,56 @@
...
+const { cloneDeep, find, flatMap, reject } = require('lodash');
+const { formatterFor } = require('./formatters');
+
+/**
+ * Enriches a form object
+ *
+ * @param {Object} baseForm
+ * @param {Object} [data]
+ */
+module.exports = function enrichForm(baseForm, data = {}) {
```
davidrapson: This method now does a lot less. In fact it only does two things:

1. Assigns values to fields
2. Filters out skippable steps

Even the second part could arguably live in the controller, or be handled in the new richer form model so we might be able to boil this down to a method which is solely concerned with field values.

controllers/apply-next/form-router.js
```diff
@@ -554,6 +434,125 @@ function initFormRouter(formModel) {
...
+            } else {
+                res.redirect(req.baseUrl);
+            }
+        };
+    }
+
+    /**
+     * Routes: Form sections
+     */
+    router
```
davidrapson: Outside of the form model structure this is the other larger change. Rather than looping through the steps we instead use route parameters to look things up. We need to do this as we're always determining the full form model using `formBuilder` based on the current data (if any).

This doesn't change a huge amount of the router code but does mean we need to do some extra work to lookup the active section and step based on the current form model.

controllers/apply-next/simple/form.js
```diff
@@ -0,0 +1,1270 @@
...
+            ]
+        };
+
+        return { ...defaultProps, ...props };
+    }
+
+    const allFields = {
+        projectName: {
+            name: 'project-name',
+            label: localise({ en: 'What is the name of your project?', cy: '' }),
```
davidrapson: This `localise` method is an alias for `get(locale)` using `lodash/fp/get`. Mostly just for syntax sugar over directly calling `{ en: 'What is the name of your project?', cy: '' }[locale]`.

Upshot is that we're doing the localisation inline here, rather than in two steps. As we call it everywhere there are locale strings in this file I wanted to make sure it read nicely.

controllers/apply-next/simple/form.js
```diff
@@ -0,0 +1,1270 @@
...
+            explanation: localise({ en: 'The project name should be simple and to the point', cy: '' }),
+            type: 'text',
+            isRequired: true,
+            schema: Joi.string().required(),
+            messages: [{ type: 'base', message: { en: 'Enter a project name', cy: '' } }]
```
davidrapson: Messages are the one place where we are not (currently) translating the strings inline as this happens in a different module and the changes here were already substantial enough. I think it would be natural to update these to behave the same way as other strings in this file.

controllers/apply-next/simple/form.test.js
```diff
@@ -0,0 +1,892 @@
...
+            'not-for-profit-company': ['company-number', 'charity-number']
+        };
+
+        map(mappings, (expected, type) => {
+            const result = fieldNamesFn({ 'organisation-type': type });
+            expect(result).toEqual(expected);
+        });
+    });
+
+    test('exclude address and date of birth for schools or statutory bodies', () => {
```
davidrapson: These tests were previously handled in multiple places via the `shouldShow` method. Now we are able to dynamically return fields based on form data directly in the model we can test this logic for the whole step, rather than only for the discreet fields. Gives us better guarantees that we're validating what is actually show, and is hopefully easier to understand what the rules are.

PR #1901, opened 2019-04-29
New field macros

davidrapson:
This PR creates a new version of the field macros focussed on the fields we are using for new forms. This enables us to iterate more freely on new form fields without the risk of introducing changes to current fields. I've left granular commits to show the stages to this process:

- Wholesale clone the form-fields file to form-fields-next and update the new forms to use this new file.
- Make some simplifications to the new macros to remove the requirement for `fieldCopy` and to make some desired changes like not showing * required for all fields (only marking when fields are optional)
- Adds some experimental tests for the new field macros
- Remove any work-in-progress or new field types from the original macros file. Stripping it back to only what is used on in-production forms.

### DIFF
--- a/modules/test-renderers.js
+++ b/modules/test-renderers.js
@@ -1,0 +1,32 @@
+'use strict';
+
+const nunjucks = require('nunjucks');
+nunjucks.configure('views');
+
+function renderComponentMacro(componentPath, macroName, params) {
+    /**
+     * Mock any global translation strings.
+     * Wherever possible you should avoid requiring context for components
+     */
+    const context = {
+        __(key) {
+            return `[i18n:${key}]`;
+        }
+    };
+
+    const macroParams = JSON.stringify(params, null, 2);
+    const macroString = `{%- from "${componentPath}" import ${macroName} with context -%}{{- ${macroName}(${macroParams}) -}}`;
+    return new Promise((resolve, reject) => {
+        nunjucks.renderString(macroString, context, (err, res) => {
+            if (err) {
+                reject(err);
+            } else {
+                resolve(res);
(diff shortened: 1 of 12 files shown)


modules/test-renderers.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const nunjucks = require('nunjucks');
+nunjucks.configure('views');
+
+function renderComponentMacro(componentPath, macroName, params) {
```
davidrapson: This is slightly experimental but explores writing snapshot tests for individual form macros. It has a big caveat that anything relying on context data outside of explicitly passed props needs to be mocked. As component macros should _largely_ be self-contained this feels like a reasonable trade-off but we'll find out as we add more tests.

modules/test-renderers.js
```diff
@@ -0,0 +1,32 @@
...
+     * Wherever possible you should avoid requiring context for components
+     */
+    const context = {
+        __(key) {
+            return `[i18n:${key}]`;
+        }
+    };
+
+    const macroParams = JSON.stringify(params, null, 2);
+    const macroString = `{%- from "${componentPath}" import ${macroName} with context -%}{{- ${macroName}(${macroParams}) -}}`;
```
davidrapson: This approach is based on https://github.com/alphagov/govuk-frontend/blob/master/lib/jest-helpers.js#L37 so all credit to the GDS team for this approach.

davidrapson: This approach is based on https://github.com/alphagov/govuk-frontend/blob/master/lib/jest-helpers.js#L37 so all credit to the GDS team for this approach.

PR #1902, opened 2019-04-30
Address history

davidrapson:
address-history

### DIFF
--- a/assets/js/modules/forms.js
+++ b/assets/js/modules/forms.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import forEach from 'lodash/forEach';
 import { trackEvent } from '../helpers/metrics';
 
 function handleBeforeUnload(e) {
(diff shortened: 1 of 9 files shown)


assets/js/modules/forms.js
```diff
@@ -81,6 +82,45 @@ function handleExpandingDetails() {
     });
 }
 
+function handleConditionalRadios() {
+    forEach(document.querySelectorAll('.js-conditional-radios'), el => {
+        function setAttributes(radioEl) {
+            var inputIsChecked = radioEl.checked;
+            var conditionalEl = el.querySelector(`#${radioEl.getAttribute('aria-controls')}`);
+            if (conditionalEl) {
+                conditionalEl.setAttribute('aria-expanded', inputIsChecked);
```
davidrapson: Not 100% sure that combining `aria-expanded` with `hidden` is the right semantics, but something to test.

PR #1905, opened 2019-04-30
Update dependencies based on npm-audit

davidrapson:
Update dependencies based on `npm audit` results. Includes some updates to client-side dependencies so will need some extra cross-browser checks before this goes out.

### DIFF
--- a/controllers/apply/reaching-communities/processor.test.js
+++ b/controllers/apply/reaching-communities/processor.test.js
@@ -9,14 +9,20 @@ const jsdom = require('jsdom');
 const { JSDOM } = jsdom;
 
 const { flattenFormData, stepsWithValues } = require('../helpers');
+const form = require('./form-model');
+const processor = require('./processor');
 
-function createMockForm() {
-    return {
+function mockProcessor() {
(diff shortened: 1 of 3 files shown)


controllers/apply/reaching-communities/processor.test.js
```diff
@@ -9,14 +9,20 @@ const jsdom = require('jsdom');
 const { JSDOM } = jsdom;
```
davidrapson: These test updates are based on a warning by the `jest` upgrade. This test was doing set up in an idiomatic way which Jest now flags as a warning.

package.json
```diff
@@ -157,12 +158,13 @@
     "eslint-plugin-vue": "5.1.0",
     "faker": "4.1.0",
     "husky": "1.3.1",
-    "jest": "24.1.0",
+    "jest": "24.7.1",
     "jest-serializer-html": "6.0.0",
+    "jquery": "3.4.0",
```
davidrapson: Moving `jquery` to `devDependencies` as we bundle it in the build so doesn't need to be in the artefact after that.

PR #1917, opened 2019-05-07
Refine summary

davidrapson:
Not a huge amount of visual changes going on here but this PR refines the summary and dashboard screen templates. Notably adding some basic responsive styles but also reworking the CSS to break the components down into more granular pieces rather than one mega page-level component. Also does a little work to start moving shared copy into the translation files.


controllers/apply-next/views/dashboard.njk
```diff
@@ -1,7 +1,134 @@
...
+            {% if application.application_data['project-postcode'] %}
+                {{ application.application_data['project-postcode'] }}
+            {% else %}
+                <span class="application-data-missing">not yet completed</span>
+            {% endif %}
+        </dd>
+    </dl>
+{% endmacro %}
+
+{% macro applicationCard(application) %}
```
davidrapson: I've inlined this back in to this template rather than living as a component file for now. These cards rely on quite a lot of state information and context data and is only used on this page so made sense to keep it in here for now. We can always extract it out again afterwards if it makes sense in future.

views/components/details/macro.njk
```diff
@@ -0,0 +1,6 @@
+{% macro details(title, isOpen = false, extraClasses) %}
```
davidrapson: Added a common macro for this to abstract our some of the common styles, notably needing to wrap the contents in an extra `<div>` for IE support.

PR #1918, opened 2019-05-07
Add a User dashboard

davidrapson:
Cleaned up version of https://github.com/biglotteryfund/blf-alpha/pull/1872 building on from @mattandrews work.

image


controllers/routes.js
```diff
@@ -64,6 +65,11 @@ const toplevel = {
             path: '/search',
             router: require('./search')
         },
+        {
+            path: '/user',
```
davidrapson: These routes should be available in welsh so moving into routes rather than as a standalone import.

PR #1921, opened 2019-05-09
Upgrade test dependencies

davidrapson:
Upgrades `jest` and `eslint` related dependencies ✨


services/__tests__/__snapshots__/mail.test.js.snap
```diff
@@ -3,11 +3,9 @@
 exports[`buildMailOptions build mail options for a html email 1`] = `
```
davidrapson: This is from upgrading `jest-serializer-html` which trims the output more than before.

controllers/apply/reaching-communities/processor.test.js
```diff
@@ -9,14 +9,23 @@ const jsdom = require('jsdom');
 const { JSDOM } = jsdom;
```
davidrapson: The set up for this test was un-idiomatic and the new version of jest flagged this as an error, these changes fix the setup code to run per test rather than inline in the `describe` block.

PR #1926, opened 2019-05-13
Dashboard cleanup

davidrapson:
image

### DIFF
--- a/controllers/apply-next/form-router.js
+++ b/controllers/apply-next/form-router.js
@@ -2,6 +2,7 @@
 const express = require('express');
 const path = require('path');
 const Raven = require('raven');
+const moment = require('moment');
 const {
     concat,
     findIndex,
@@ -11,6 +12,7 @@ const {
     includes,
     isEmpty,
     omit,
+    partition,
     set,
     unset
 } = require('lodash');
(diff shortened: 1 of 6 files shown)


controllers/apply-next/form-router.js
```diff
@@ -106,29 +108,58 @@ function initFormRouter({
...
-            applications: applicationsWithProgress,
-            form: form
-        });
+        try {
+            const applications = await applicationsService.getByForm({
+                userId: req.user.userData.id,
+                formId: id
+            });
+
+            const [submittedApplications, inProgressApplications] = partition(
```
davidrapson: Groups the dashboard into in progress and submitted applications.

controllers/apply-next/simple/form.js
```diff
@@ -555,22 +556,49 @@ module.exports = function({ locale, data = {} }) {
...
+                    return acc;
+                },
+                {}
+            )
+        );
+
+        return computed;
+    }
+
+    function summary() {
```
davidrapson: Adds a summary method to the form model. This function provides a summary of the current application, rather than hard-coding assumptions about the form into the template or database model.

models/application.js
```diff
@@ -1,53 +1,31 @@
...
+    return sequelize.define('application', {
+        id: {
+            primaryKey: true,
+            type: DataTypes.UUID,
+            defaultValue: DataTypes.UUIDV4,
+            allowNull: false
         },
-        {
-            getterMethods: {
-                grantAmount() {
```
davidrapson: This now exists on the `summary` method in the form builder rather than being tied into the database mode (which doesn't know what fields are available)

models/application.js
```diff
@@ -1,53 +1,31 @@
...
-                    if (this.application_data && this.application_data['project-budget']) {
-                        return sumBy(this.application_data['project-budget'], item => parseInt(item.cost || 0));
-                    }
-                    return false;
-                },
-                expiryDate() {
```
davidrapson: I've removed this from here for now as this should probably exist as a field in the database so we can query against it.

PR #1928, opened 2019-05-13
Update debug prefix to tnlcf

davidrapson:
📝

### DIFF
--- a/modules/appData.js
+++ b/modules/appData.js
@@ -3,7 +3,6 @@ const fs = require('fs');
 const path = require('path');
 const config = require('config');
 const { get } = require('lodash');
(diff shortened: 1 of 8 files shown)


modules/appData.js
```diff
@@ -19,7 +18,6 @@ function getDeploymentData() {
 
 const deploymentData = getDeploymentData();
 const environment = config.util.getEnv('NODE_ENV'); // Defaults to 'development'
-debug(`App environment is ${environment}`);
```
davidrapson: Logging this in `bin/www` rather than is this file which gets imported everywhere as a general utility.

PR #1930, opened 2019-05-14
Feedback cleanup

davidrapson:
Spotted whilst reviewing our site-feedback that we were encouraging duplicate responses in places due to the unused `enablePrompt` flag. This flag referred to an old pop-over prompt message (the kind we've used to promote surveys) which is no longer active.

Instead we want to disable the global _site survey_ (e.g. "Did you find what you were looking for?") on pages that have inline feedback.

This PR switches the flags and removes the unused prompt code. Whilst this features has been useful for promoting surveys in the past the position now conflicts with the cookie message so we'd need to redesign this if we ever wanted to show messages like this again anyway, so feels safe to remove for now.


controllers/archived.js
```diff
@@ -23,22 +27,16 @@ legacyPagePaths.forEach(urlPath => {
...
 /**
  * Archived files
- * Requests for legacy files (eg. stored on Sitecore) will be shown an error message
+ * Requests for legacy files will be shown an error message
  * along with a feedback form to explain what they were looking for.
- * We also log all requests for these files to ensure we can update anything missing.
  */
 router.get(legacyFilesPath, (req, res) => {
-    const filePath = req.originalUrl;
-    metrics.count({
```
davidrapson: I'm removing this because this is also getting tracked as a page-view in analytics which is more readable. The cloudwatch metrics aren't very readable in this case.

PR #1933, opened 2019-05-14
Beneficiary questions continued

davidrapson:
image


controllers/apply-next/simple/fields.js
```diff
@@ -1115,6 +1116,158 @@ module.exports = function fieldsFor({ locale, data = {} }) {
...
         },
+        beneficiariesEthnicBakground: {
+            name: 'beneficiaries-ethnic-background',
+            label: localise({ en: `Ethnic background`, cy: '' }),
+            explanation: localise({
+                en: `You have indicated that your project mostly benefits people from a particular ethnic background, please select which one(s).`,
+                cy: ``
+            }),
+            type: 'checkbox',
+            optgroups: [
```
davidrapson: `optgroups` isn't _technically_ correct for a name as it's not a real select `<optgroup>` but performs the same semantics (grouping options by a heading) so I kept the same name.

controllers/apply-next/lib/validators.js
```diff
@@ -18,6 +18,21 @@ function multiChoice(options) {
         .single();
 }
 
+/**
+ * @see https://github.com/hapijs/joi/issues/622
+ */
+function conditionalOnMultiChoice({ ref, match, schema }) {
```
davidrapson: The previous logic for these conditional multi-choice fields wasn't validating correctly. Based on the github issue linked here this is the way to do this. A little verbose so I added a wrapper function.

PR #1934, opened 2019-05-15
Fixup flexbox styling in IE

davidrapson:
Fixes a small flexbox 🐛 on the dashboard screen. The flexbox syntax was a little wonky causing the cards to expand outside the page slightly.

### DIFF
--- a/assets/sass/components/_applications.scss
+++ b/assets/sass/components/_applications.scss
@@ -226,8 +226,6 @@ $statuses: (
     .application-card__summary {
         padding: $spacingUnit;
         margin-right: $spacingUnit;
-        flex-shrink: 0;
-        flex-basis: 60%;
     }
 
     .application-card__progress {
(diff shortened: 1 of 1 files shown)


assets/sass/components/_applications.scss
```diff
@@ -242,14 +240,21 @@ $statuses: (
...
     @include mq('medium') {
         display: flex;
-        justify-content: space-between;
+
+        .application-card__summary {
+            flex: 1 1 auto;
+        }
+
+        .application-card__progress {
+            flex: 0 0 40%;
```
davidrapson: This is the replacement for `flex-basis: 60%;` above, setting the opposite column instead.

assets/sass/components/_applications.scss
```diff
@@ -242,14 +240,21 @@ $statuses: (
 
     @include mq('medium') {
         display: flex;
-        justify-content: space-between;
+
+        .application-card__summary {
+            flex: 1 1 auto;
```
davidrapson: This is the line that was missing, needed a flex statement on the second column

assets/sass/components/_applications.scss
```diff
@@ -242,14 +240,21 @@ $statuses: (
...
+            flex: 0 0 40%;
+        }
 
         .application-card__progress {
             display: flex;
             flex-direction: column;
         }
         .application-card__progress__status {
-            flex: 1 1 0;
+            flex: 1 1 auto;
```
davidrapson: Needed to be auto to avoid text collapsing.

PR #1943, opened 2019-05-20
Add initial salesforce connector

davidrapson:
This adds a really basic salesforce connector to send application data to the apex rest endpoint for application submissions.

We need to test the performance profile further to know if this is enough or if we should look a processing queue. For now we can take the simplest approach possible.

### DIFF
--- a/modules/secrets.js
+++ b/modules/secrets.js
@@ -7,7 +7,8 @@ const { getSecret } = require('./parameter-store');
  * Primary session secret
  * We allow overriding through an environment variable for CI
  */
-const SESSION_SECRET = process.env.SESSION_SECRET || getSecret('session.secret', true);
+const SESSION_SECRET =
+    process.env.SESSION_SECRET || getSecret('session.secret', true);
 
 /**
  * Database connection uri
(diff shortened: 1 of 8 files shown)


modules/secrets.js
```diff
@@ -51,23 +55,39 @@ const SENTRY_DSN = getSecret('sentry.dsn');
...
     MS_CLIENT_SECRET: getSecret('ms.auth.tnlcf.clientSecret'),
-    MS_REDIRECT_URL: process.env.MS_REDIRECT_URL || getSecret('ms.auth.redirectUrl')
+    MS_REDIRECT_URL:
+        process.env.MS_REDIRECT_URL || getSecret('ms.auth.redirectUrl')
+};
+
+/**
+ * Salesforce authentication
+ */
+const SALESFORCE_AUTH = {
```
davidrapson: This will come from parameter store once we've confirmed the API user we are connecting with.

services/salesforce.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+const request = require('request-promise-native');
+const { SALESFORCE_AUTH } = require('../modules/secrets');
+
+class Salesforce {
```
davidrapson: The service returns an instance of this class when calling `authorise`. Adds a little wrapper to handle passing the oath token along with the  request. We create a new connection each time we submit a form.

davidrapson:
This will fail in CI at the moment, we will want to skip submission in CI.

services/salesforce.js
mattandrews: Ooh our first `class`. Nice!

davidrapson:
Added a flag to only enable this in dev (and outside of test servers), allows us to merge this and configure the final credentials later.

PR #1948, opened 2019-05-21
Validate form helper

davidrapson:
1. Our current field validation tests rely on asserting agains the underlying `joi` error message, which is not ideal
2. We have to do some lookups inside the router in order to get messages, it would be good if we could get _our_ messages back alongside the validation

This PR adds a new `validate-form` helper to do that, and updates a couple of the field tests by way of example to show how we can now test against the the normalised error messages.

I've also made a change to save the **full** submission when submitting a step, rather than just the valid fields. We revalidate at multiple steps so it's OK to temporarily store impartial date and results in a better user experience.


davidrapson:
🤓

PR #1950, opened 2019-05-22
Remove custom email in favour of salesforce email

davidrapson:
Waiting on confirmation but it's likely that we'll explore sending confirmation emails directly from Salesforce on submission to both main and senior contact. This would allow us to delete our current email code, which currently only sends to main contact.

Consider this pending until we get confirmation.


davidrapson:
Seems pretty likely that we're going to try this from the Salesforce end, so I reckon we can merge this (we can always reinstate later)

PR #1953, opened 2019-05-23
Country questions

davidrapson:
Adds two country specific questions to the beneficiaries questions. These fields are required if the project country is the relevant country.

welsh-language
ni-community

### DIFF
--- a/controllers/apply-next/simple/fields.test.js
+++ b/controllers/apply-next/simple/fields.test.js
@@ -264,40 +264,6 @@ describe('fields', () => {
         });
     }
 
-    describe('companyNumber', () => {
-        test('conditionally required based on organisation type', () => {
-            assertValid(fields.companyNumber, 'CE002712');
-            assertValid(fields.companyNumber, undefined);
-
-            assertRequiredForOrganistionTypes(fields.companyNumber, [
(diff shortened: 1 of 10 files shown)


controllers/apply-next/simple/fields.test.js
```diff
@@ -264,40 +264,6 @@ describe('fields', () => {
         });
     }
 
-    describe('companyNumber', () => {
```
davidrapson: These tests have been combined with the related ones in `form.test.js` where we can test the normalised error messages, not just `joi` behaviour.

views/components/form-fields-next/macros.njk
```diff
@@ -152,7 +152,7 @@
 
     <select class="ff-select" id="field-{{ field.name }}" name="{{ field.name }}">
         {% if field.defaultOption %}
-            <option disabled value{% if not field.value %} selected{% endif %}>{{ field.defaultOption }}</option>
```
davidrapson: This fixes a bug when changing country where the first item in the list would get selected, rather than defaulting to "select a country"

davidrapson:
[Add country specific questions to beneficiaries screen](https://trello.com/c/XyOQPiwf/363-add-country-specific-questions-to-beneficiaries-screen)

PR #1957, opened 2019-05-24
Set rate limit zone in nginx for user routes

davidrapson:
Adds basic nginx rate limiting to user endpoints. Limits only on POST to login and register. Limits to 1r/s, with burst of 20. Essentially queues up requests and processes them a 1r/s if more than 20 requests come in at once the 21st with be rate limited. Based on https://www.nginx.com/blog/rate-limiting-nginx/

Given the type of rate limiting we are doing using the request_uri is appropriate, no need to worry about the client IP as for this use case we are concerned about the overall number of requests at once to the endpoints not specifically one client.

Tested locally with `siege` but will need deploying to the test site and some more tests running.


davidrapson:
[Configure rate-limiting on user account endpoints](https://trello.com/c/8dnbWBOm/351-configure-rate-limiting-on-user-account-endpoints)

deploy/server.conf
```diff
@@ -21,6 +31,10 @@ server {
     gzip_http_version 1.1;
     gzip_types text/plain text/css application/json application/javascript text/xml;
 
+    location ~ ^/user {
```
mattandrews: My regex-fu may be failing me but should this be modified to include `/welsh/user`, too (when it eventually exists)? Don't underestimate the power of a gwadu gwasanaeth wedi'i ddosbarthu attack.

davidrapson: Ah good point!

davidrapson:
Yeah, sends a 503 error which in our cloudfront config gets mapped to show the standard error page, but we could map this one to a custom one if we need to.

PR #1960, opened 2019-05-28
Sentry upgrade

davidrapson:
We're running a good few versions behind on our Sentry version. This PR attempts to bring us up to date. There are two parts to this:

**Node upgrade**
This is fairly straightforward: migrates to `@sentry/node`, updates references.

**Browser upgrade**
A bit more involved, took some trial and error to get it working with Vue. Ended up initialising this alongside Vue in the main bundle and made the previous "common" code a code-split bundle to keep the size down.

Happy to split this out into two PRs if we're more comfortable getting the Node change out first as that's the simpler one.


views/includes/metaHead.njk
```diff
@@ -1,24 +1,14 @@
...
+<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=HTMLPictureElement,Promise,Object.assign,Number.isNaN,String.prototype.includes,Array.prototype.find,Element.prototype.closest"></script>
 <script id="script-config">
 (function() {
-    var AppConfig = {
+    window.AppConfig = {
         environment: '{{ appData.environment }}',
-        isNotProduction: {{ appData.isNotProduction }},
         locale: '{{ locale }}',
         localePrefix: '{{ localePrefix }}',
-        isOldIE:
```
davidrapson: I've inlined these checks into the modules they are actually used in

views/includes/metaFooterJS.njk
```diff
@@ -1,10 +1,4 @@
-<script src="https://www.google-analytics.com/cx/api.js"></script>
```
davidrapson: Removing this as we're not using the google experiments / ab testing code anymore so we can avoid loading all this.

server.js
```diff
@@ -47,24 +47,24 @@ const vanityMiddleware = require('./middleware/vanity');
...
-Raven.config(SENTRY_DSN, {
+Sentry.init({
+    dsn: SENTRY_DSN,
     logger: 'server',
     environment: appData.environment,
-    release: appData.commitId,
-    autoBreadcrumbs: true,
-    dataCallback(data) {
-        // Clear installed node_modules
-        delete data.modules;
```
davidrapson: node_modules are no longer passed by default.

modules/secrets.js
```diff
@@ -47,7 +47,7 @@ const JWT_SIGNING_TOKEN =
 /**
  * Sentry DSN for error reporting
  */
-const SENTRY_DSN = getSecret('sentry.dsn');
+const SENTRY_DSN = getSecret('sentry.publicDsn');
```
davidrapson: Sentry updated their requirements a while ago so that all SDKs just use the public DSN, no more private DSN needed. I've added a new value to parameter store and we can remove the old one once this change is out.

controllers/grants/index.js
```diff
@@ -2,12 +2,17 @@
 const { concat, clone, pick, isEmpty, get } = require('lodash');
```
davidrapson: I think this is prettier doing some auto-formatting…

server.js
mattandrews: 🙏

PR #1961, opened 2019-05-29
Model restructure

davidrapson:
**Move models to structure recommended by `sequelize-ci`**
This takes steps towards using `sequelize-cli` by moving the models under a `db` directory with a `database-config.js` file which is used by the cli for migrations etc. 

This doesn't include the cli yet, just gets the structure ready for it.

**Migrate away from "services"**
The split between "services" and "models" was starting to bug me a little. The models felt like they could be doing more heavy lifting, currently they only define fields. Thankfully, we can lift many of the find/create methods that live in 'services' into the  model file. More custom code I've inlined into the controllers that actually use it, this is pretty much always code that's used by tools to power summary screens so makes sense to live there.

I've only done this for the Feedback and Survey models for now as a demonstration but the idea would be to update all models to this style.

**Rename modules to common**
The "modules" directory is a bit of a dumping ground of generic code. "modules" is also a _really_ generic name that means another thing in node world. In an attempt to make intent clearer I've renamed this "common". Not the most inventive name, but feels a little clearer.

As a result I've moved some code that was not "common" to the controllers that used it (some single-use past grants code in particular).

Thoughts wanted 💭


controllers/tools/surveys.js
```diff
@@ -1,16 +1,104 @@
...
+const times = require('lodash/times');
+const features = require('config').get('features');
 
-const appData = require('../../modules/appData');
-const surveysService = require('../../services/surveys');
+const { SurveyAnswer } = require('../../db/models');
 
 const router = express.Router();
 
+function voteDataFor(responses) {
```
davidrapson: This code is specific to this tools page, so makes sense to live here.

db/database-config.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
davidrapson: This structure is used by `sequelize-ci` to determine environments when running migrations. For now it just lets us clean up the config a little bit.

db/models/feedback.js
```diff
@@ -0,0 +1,48 @@
+// @ts-nocheck
```
davidrapson: This is the main change to 👀 at. Updates the Feedback and Survey models to use classes which is now the default in v5 (which we should upgrade to soon, but this works already in v4).

db/models/index.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+/**
+ * Register models and associations
+ */
+const db = {
+    Users: sequelize.import('./user'),
+    Staff: sequelize.import('./staff'),
+    Feedback: Feedback.init(sequelize, Sequelize),
+    SurveyAnswer: SurveyAnswer.init(sequelize, Sequelize),
```
davidrapson: When using classes for models you call `.init` on the model, rather than `sequelize.import`

PR #1964, opened 2019-05-29
Common cleanup

davidrapson:
On a bit of a restructuring bent with the last few PRs. Another one to clean up some of our common code.

**Update CSP directives  to be more testable**

Extracts the code to generate our CSP directives into a plain module and adds snapshot testing to cover each scenario (default, with hotjar, and with localhost debug urls enabled)

**Split robots.txt, sitemap.xml, and status page routers**

- Status is now mounted directly in `server.js`
- Robots.txt is at `controllers/robots`
- Sitemap is at `controllers/sitemap`. Also updates the non-CMS routes to be a static whitelist (rather than ineffectively trying to generate them from `routes.js`

 **Remove redundant code from routes.js**

- Removes a set of unused parameters
- Inlines the handful of remaining `lang` and `heroSlug` references into their respective routers

**Aside:** Interestingly this leaves the objects in `routes.js` as only a list of `path`s and `routers`, and is only included in `server.js` so we _could_ do away with this file completely and mount each section as an explicit router directly from `server.js` 💭


common/csp-directives.js
```diff
@@ -0,0 +1,97 @@
...
+        'ajax.googleapis.com',
+        'cdnjs.cloudflare.com',
+        'platform.twitter.com',
+        'sentry.io',
+        'syndication.twitter.com',
+        'www.google-analytics.com',
+        'use.typekit.net'
+    ];
+
+    if (allowLocalhost) {
```
davidrapson: This was previously two blocks of code: one for local images,  and another for browser-stack, I've renamed this to be an `allowLocalhost` feature flag as this more directly describes what we are doing here.

controllers/sitemap.js
```diff
@@ -0,0 +1,90 @@
...
+ */
+async function getCanonicalRoutes() {
+    /**
+     * Static routes
+     *
+     * Used for pages that are outside of the CMS, either:
+     * - Hard-coded pages (like homepage and 10k pages)
+     * - Landing pages (where child pages are powered by the CMS, but landing is custom)
+     */
+    const staticRoutes = [
```
davidrapson: This list was previously generated from `routes.js`. The biggest problem with this is that it only generated entries for the _mount-path_ of a section, i.e. didn't account for sub-pages like `/funding/programmes/all`.

Given that only a handful of pages live outside the CMS this feels simpler to hard-code.

server.js
```diff
@@ -219,40 +262,24 @@ app.use('/', require('./controllers/archived'));
...
-     * Apply page level middleware and mount router if we have one
+     * Mount page router
      */
     section.pages.forEach(function(page) {
-        router
-            .route(page.path)
-            .all(
-                injectCopy(page.lang),
-                injectHeroImage(page.heroSlug),
-                (req, res, next) => {
```
davidrapson: This was my favourite middleware to remove

```
(req, res, next) => {
    next();
}
```

That's right it just calls `next` 🤷‍♂

server.js
```diff
@@ -219,40 +262,24 @@ app.use('/', require('./controllers/archived'));
...
         next();
     });
 
     /**
-     * Page-level logic
-     * Apply page level middleware and mount router if we have one
+     * Mount page router
      */
     section.pages.forEach(function(page) {
-        router
```
davidrapson: These "helper" middleware for `injectCopy` and `injectHeroImage` are now all called directly from the routes that need them.

controllers/sitemap.js
mattandrews: 👍 and a great motivator (and helpful list) to get most of the remaining stuff into the CMS, too.

server.js
mattandrews: wow.

davidrapson: thank u, next

PR #1967, opened 2019-05-30
Contact strip bug

davidrapson:
Fixes a 🐛 excellently spotted by @mattandrews where contact details (date of birth and address) were getting stripped when the organisation type wasn't present. This information should only be stripped when an organisation type is explicitly school or statutory body.

After some debugging it turns out we were missing an `.exist` on the `joi` validation.

- Adds failing tests
- Addes `.exist` to `.when` clause

### DIFF
--- a/controllers/apply/awards-for-all/fields.test.js
+++ b/controllers/apply/awards-for-all/fields.test.js
@@ -288,37 +288,6 @@ describe('fields', () => {
         });
     }
 
-    function testContactDateOfBirth(field, minAge) {
-        test(`must be at least ${minAge} years old`, () => {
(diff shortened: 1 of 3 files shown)


controllers/apply/awards-for-all/fields.test.js
```diff
@@ -288,79 +288,6 @@ describe('fields', () => {
         });
     }
 
-    function testContactDateOfBirth(field, minAge) {
```
davidrapson: Moved these tests to `form.test.js`, aiming to migrate all these tests over to `form.test.js` soon.

PR #1970, opened 2019-05-30
Downgrade to cypress 3.2

davidrapson:
Noticed that our tests have been running ~2 minutes slower since upgrading to Cypress 3.3

image

Downgrading for now and tracking https://github.com/cypress-io/cypress/issues/4313 (same symptoms although with far larger number of tests)


mattandrews:
Hmm, interesting. 👍

PR #1971, opened 2019-05-30
Login register tweaks

davidrapson:
- Convert alertMessage to plain function and move to user lib
- Update register to create account, add copy to locales

image


davidrapson:
[Rename register to "Create an account"](https://trello.com/c/7KCEaQTM/408-rename-register-to-create-an-account)

PR #1972, opened 2019-05-31
Salesforce cleanup

davidrapson:
**Add `forSalesforce` function to form builder**

Rather than having this separate from where the form is designed, I thought it made sense to move this to the form builder, similar to how the summary is constructed.

**Inline processor into form-router**
Until we build more forms we don't really need the separate processor file here. I think it's reasonable to assume for the moment that these forms will go into salesforce, removing this layer of abstraction until we know we need it.

**Update dateRange key names**
There was a problem with the salesforce processing when using `start` and `end`, these are reserved names apparently. It's easy for us to rename the fields rather than working around this so I've named them `startDate` and `endDate`.


davidrapson:
@mattandrews Added the placeholder `contentVersion` method in here too, commented out, but gives a reference of what information is needed.

PR #1974, opened 2019-06-04
Remove head.js, add handler for web-fonts

davidrapson:
The recent Sentry upgrade now tracks unhandled promise exceptions, this is reporting cases where webfonts fail to load. The stack trace for this is pretty noisy. We want to track this but it doesn't need to be an error. This PR reworks how are fonts are loaded so we can capture an info level message if they fail to load.

Pros:

- Removes the need for a relatively large inlined `head.js`
- Allows us to capture explicit info events when web fonts fail to load
- Allows us to simplify our webpack config and only build a single entry point

Cons:

- Web fonts are loaded slightly later on first load. Not necessarily a con, our fallbacks are very similar and we are already doing async loading
- Details polyfill is loaded later. Need to check impact of this on IE 11.


postcss.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     plugins: [
         autoprefixer(),
         foutWithAClass({
-            families: ['Poppins', 'Roboto', 'caecilia', 'caecilia-sans-text'],
```
davidrapson: Last reference to the old typefaces.

davidrapson:
(Ran some tests in IE and there was no noticeable change from the `details` polyfill move)

PR #1975, opened 2019-06-04
Split application models

davidrapson:
As discussed this PR splits the application model into two separate models

**PendingApplications**

Concerned with pending/in-flight applications. This is essentially the original model with the addition of an `expirtyDate` and `submissionAttempts` field.

**SubmittedApplications**

This model stores a representation of the form at the time of submission. The main focus is to avoid this being tied to the active form model. This table should contain all the information needed to show a summary of the application at the time of submission. It also stores a reference to the Salesforce ID and a copy of the JSON submitted.

I'll call out specific changes through comments


davidrapson:
[Split application model into PendingApplication and SubmittedApplication](https://trello.com/c/bYEzMVPr/429-split-application-model-into-pendingapplication-and-submittedapplication)

bin/www
```diff
@@ -1,75 +1,72 @@
...
-};
+}
 
-// sync database models
-try {
-    models.sequelize.sync().then(function() {
+/**
+ * Sync database models and start application
+ */
+sequelize
```
davidrapson: I noticed whilst working on this that the try/catch syntax here was incorrect (this is async) so needed to be a `.catch` otherwise database errors were never logged.

config/development.json
```diff
@@ -6,6 +6,6 @@
     "enableAllowLocalhost": false,
     "enableAwardsForAllApplications": true,
     "enableSeeders": true,
-    "enableSalesforceConnector": true
+    "enableSalesforceConnector": false
```
davidrapson: I'll turn this back on once this is ready to merge

controllers/apply/form-router-next/views/delete.njk
```diff
@@ -5,19 +5,25 @@
 {% block content %}
     <main role="main" id="content">
         <div class="content-box u-inner u-inner-wide-only">
-            <div class="s-prose u-constrained-wide">
-                <h1>Are you sure you want to delete your application?</h1>
-
-                <h2 class="t2 t--underline">This can't be undone.</h2>
-                <p>If you're sure you want to delete your application, then you can proceed below &ndash; please note this this cannot be recovered.</p>
+            <div class="form-header">
```
mattandrews: Worth adding the breadcrumbs in here too as another navigation option?
image

db/models/applications.js
```diff
@@ -0,0 +1,238 @@
...
+            .toDate();
+
+        return this.create({
+            userId: userId,
+            formId: formId,
+            applicationData: null,
+            expiresAt: expiresAt
+        });
+    }
+    static saveApplicationState(id, data) {
```
mattandrews: Should this method effectively reset the `expiresAt` date each time it's called?

PR #1978, opened 2019-06-06
Add file uploads

mattandrews:
This took a while to get working but I think it's in a fairly good state now. Here's how it works:

- We mark some form steps as `isMultipart`, eg. they contain `file` inputs. We could do this automatically but it's manual for now.
- When one of these steps is encountered, its `<form>` tag gains ` enctype="multipart/form-data"`
- When these steps are submitted, we do some form body parsing using `Formidable`, which handles file inputs as well as text. 
- During validation, we compare the received files to what the form model expected, and then validate them against the size/type constraints.
- If the files pass, we upload them to S3 in a secure, encrypted bucket, keyed by the application and form IDs.
- If not, we delete the entered file data and return the user to the form.
- If they return to a successfully-uploaded file step, they'll see a link to download the file, which actually proxies it from S3, secured against the user's login status and only allowing access to the file currently stored for that form field (eg. it's not possible to request another user's files).

Oof.


assets/js/vue-apps/form-components/address-lookup.vue
```diff
@@ -180,7 +180,8 @@ export default {
         formIsValid() {
             const VALIDATION_REGEX = /^[a-z]{1,2}\d[a-z\d]?\s*\d[a-z]{2}$/i;
             return (
-                this.postcode && this.postcode.match(VALIDATION_REGEX) !== null
+                this.postcode &&
+                this.postcode.trim().match(VALIDATION_REGEX) !== null
```
mattandrews: This was a cheeky—and unrelated—fix for a postcode issue someone spotted the other day, which I fixed here before I forgot.

common/secrets.js
```diff
@@ -89,6 +89,8 @@ const DIGITAL_FUND_EMAIL =
 const POSTCODES_API_KEY =
     process.env.POSTCODES_API_KEY || getSecret('postcodes.api.key');
 
+const S3_KMS_KEY_ID = process.env.S3_KMS_KEY_ID || getSecret('s3.kms.key.id');
```
mattandrews: Debating now whether this itself should be a secret as it's just an ID.

controllers/apply/form-router-next/lib/validate-model.js
```diff
@@ -81,6 +81,38 @@ module.exports = function validateModel(formModel) {
...
+        'date',
+        'date-range',
+        'day-month'
+    ];
+
+    // Check that we haven't used a multipart field on a step with array values
+    // because we use a custom body parser (Formidable) which doesn't handle these
+    // in the same way as body-parser (eg. using qs library). Throws an error if
+    // we attempt to add an array field to one of these steps.
+    formModel.sections = formModel.sections.map(section => {
```
mattandrews: This feels a bit janky but I like the reassurance of it failing the build if we forget about these constraints.

controllers/apply/form-router-next/index.js
```diff
@@ -51,6 +54,36 @@ function initFormRouter({
...
+                        Sentry.captureException(err);
+                    } else {
+                        req.body = fields;
+                        req.files = files;
+                    }
+                    next();
+                });
+                formData.on('error', err => {
+                    Sentry.captureException(err);
+                    next();
```
davidrapson: How come we call `next` without passing the error here? Is this not a fatal error at this point?

controllers/apply/form-router-next/index.js
```diff
@@ -69,10 +73,44 @@ function initFormRouter({
         next();
     }
 
-    router.use(csrfProtection, injectCopy('applyNext'), setCommonLocals);
+    router.use(
+        // Populate req.body for multipart forms before CSRF token is needed
+        (req, res, next) => {
```
davidrapson: It might be worth moving this to the step handler, just to really limit where this is needed, but doesn't need to happen in this PR.

controllers/apply/form-router-next/steps.js
```diff
@@ -132,23 +137,90 @@ module.exports = function(formBuilder) {
...
+                const [invalidFiles, validFiles] = partition(
+                    filesToUpload,
+                    file => {
+                        return errorsForStep.find(
+                            _ => _.param === file.fieldName
+                        );
+                    }
+                );
+
+                invalidFiles.forEach(file => {
```
davidrapson: I reckon we could avoid mutating the value here using `pickBy`

```js
const newData = pickBy(validationResult.value, field => invalidFiles.map(field => file.fieldName).includes(filed.name) === false)
```

Or something similar to that

controllers/apply/form-router-next/steps.js
```diff
@@ -132,23 +137,90 @@ module.exports = function(formBuilder) {
...
+                        s3.uploadFile({
+                            formId: formId,
+                            applicationId: currentlyEditingId,
+                            fileMetadata: file
+                        })
+                    )
+                ).catch(rejection => {
+                    Sentry.captureException(rejection.error);
+                    // Manually create a form error and send the user back to the form
+                    errorsForStep = concat(errorsForStep, {
```
davidrapson: Wondering if we can avoid this assignment by directly rendering the step in this catch block.

```js
renderStep(req, res, validationResult.value, [{ 
  msg: `There was an error uploading your file - please try again`,
  param: rejection.fieldName
}]);
```

controllers/apply/form-router-next/views/step.njk
```diff
@@ -7,7 +7,11 @@
 {% block content %}
     <main role="main" id="content">
         <div class="content-box u-inner-wide-only">
-            <form action="" method="POST" class="js-form-warn-unsaved">
+            <form action=""
+                  method="post"
+                  class="js-form-warn-unsaved"
+                  {% if stepIsMultipart %}enctype="multipart/form-data"{% endif %}
```
davidrapson: I think we can call `step.isMultipart` here and avoid the extra view variable.

views/components/form-fields-next/macros.njk
```diff
@@ -227,7 +227,7 @@
                                     <input
                                         type="checkbox"
                                         id="{{ fieldId }}"
-                                        name="{{ field.name }}"
+                                        name="{{ field.name }}[]"
```
davidrapson: Is this still needed?

views/components/form-fields-next/macros.njk
```diff
@@ -495,6 +495,34 @@
...
 {% endmacro %}
 
+{% macro inputFile(field) %}
+
+    {% if field.value and field.displayValue %}
+        <div class="existing-data u-margin-bottom">
+            <h3>Existing file</h3>
+            <p>{{ field.displayValue }}</p>
+            <p class="u-no-margin">
+                <a href="../download/{{ field.name }}/{{ field.value.filename }}">Download this file</a>
```
davidrapson: We might need to define this download route somewhere on the model? This relative path might not make sense depending on where/how this macro us rendered. Doesn't need to be solved now though.

controllers/apply/form-router-next/steps.js
```diff
@@ -139,110 +149,117 @@ module.exports = function(formId, formBuilder) {
...
-                    fieldNamesForStep.includes(item.param)
-                );
+            /**
+             * Validate form against combined application data
+             * currentApplication data with req.body
+             * and file metadata mixed in.
+             */
+            const combinedApplicationData = {
+                ...applicationData,
+                ...fileValues()
```
mattandrews: 💯

PR #1979, opened 2019-06-07
Activation flow and common password checking

davidrapson:
Updates the log in and registration flow to introduce:

- Common password checking
- Enforcing activation flow

**Common passwords**

Screenshot 2019-06-07 at 14 28 26

This adds a check against a list of common passwords (top 100,000 filtered by our min-length; currently 10 characters). It doesn't add any kind of password strength checking yet but does include a link to https://www.cyberaware.gov.uk/blog/three-random-words about the "three random words" approach to secure passwords.

**Activation flow**

Screenshot 2019-06-07 at 14 27 47

We're currently enforcing this in front of the awards for all form only. You can still get to the account management pages before activating as it feels important to be able to change your email address without activating (so you can correct it if you mistyped your email). Will be interesting to see how this tests.

The copy on this will need reviewing by @scottoakley &co but the flow is in place. 

---

From the technical side this PR also merges the user model and service together. Between this and https://github.com/biglotteryfund/blf-alpha/pull/1977 we now only have one "service" left for orders so I might do a little clean up as a follow up to merge this last service into the models too  👍 👍


controllers/user/lib/account-schemas.js
```diff
@@ -0,0 +1,161 @@
...
+
+const passwordConfirmationSchema = Joi.string()
+    .valid(Joi.ref('password')) // Must match password
+    .required();
+
+const MESSAGES = {
+    emailInvalid(locale) {
+        return get(locale)({
+            en: 'Enter a valid email address',
+            cy: ''
```
mattandrews: The `/user/register` endpoint is public right now (!) so perhaps we should get these labels translated now, independent of the AFA launch (or we might forget)?

mattandrews: ... Thinking about it, the whole thing is English-only so I guess we should raise a ticket for the next sprint to get this translated (perhaps once Lily and Scott have finalised the copy for the account emails?)

controllers/user/lib/common-passwords.txt
```diff
@@ -0,0 +1,2344 @@
...
+sagittarius
+ronaldo123
+rockbottom
+Rjycnfynby
+reymysterio
+qwertyuiop12345
+qwertyuiop12
+qwerty1992
+qaz26101778
+progressive
```
mattandrews: I think you accidentally left your password in this list @davidrapson

middleware/authed.js
```diff
@@ -42,34 +73,34 @@ function requireUserAuth(req, res, next) {
...
-    } else if (req.session.redirectUrl) {
-        redirectUrl = req.session.redirectUrl;
-        delete req.session.redirectUrl;
+/*
+ * Required note staff auth
+ * Middleware to require that the visitor is NOT logged in as a staff user
+ */
+function requireNotStaffAuth(req, res, next) {
+    if (req.isAuthenticated() && isStaff(req.user)) {
+        res.redirect('/tools');
```
mattandrews: Wonder if we should show a message to staff here that they tried to access something only meant for regular accounts? eg. append `?s=regularOnly` or something on the URL?

controllers/materials/helpers.js
```diff
@@ -2,7 +2,17 @@
 const { reduce } = require('lodash');
 const { check } = require('express-validator/check');
 
-const { errorTranslator } = require('../../common/validators');
+function errorTranslator(prefix) {
```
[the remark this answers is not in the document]
davidrapson: I have deliberately inlined this because this is related to `express-validator` which we are moving away from. It shouldn't be common as that encourages reuse we want to avoid.

davidrapson:
Closing in favour of https://github.com/biglotteryfund/blf-alpha/pull/1982

PR #1980, opened 2019-06-10
Simplify dependencies

davidrapson:
Simplify dependencies a little by removing some things that are either unused or we could do without.

**Remove old Sentry packages**

`raven-js` was still hanging about, this was moved to `@sentry/browser`

**Remove `ms` package**

We were using this to generate times for `cache-control` headers, but involved some hoop jumping to convert from natural language -> milliseconds -> to seconds where we can easily just write the value in seconds.

**Remove `browser-sync`**

I'm not a fan of having this loaded in `bin/www` for something that's only used optionally in development. It turns out you can get the same behaviour by running a one-line command locally:

```
npx browser-sync start --no-open --proxy 'http://localhost:3000' --files 'public'
```

I've added this to the README. Personally I rarely use browser-sync, so this is a call for you @mattandrews as to how disruptive this change would be.

**Remove husky**

I'm curious to see if removing this has any affect on us. I nearly always have a test watcher running locally so most of the time the pre-push check just slows things down. I've also seen how many times these hooks confuse Tower (or whichever git gui you use @mattandrews). The CI step covers things anyway so thinking we can get away without this.


davidrapson:
Percy seems to have got stuck, even when rerunning the travis builds.

davidrapson:
> Husky makes my commits/pushes super slow

Honestly, I forgot how quick a git push is normally.

PR #1983, opened 2019-06-10
Reinstall @percy/cypress

davidrapson:
Something, possibly in https://github.com/biglotteryfund/blf-alpha/pull/1980, caused `axios` (a dependency of `@percy/cypress`) to be derefrenced. Uninstalling and reinstalling fixes things.


mattandrews:
Weird! 👍

PR #1990, opened 2019-06-11
Form model

davidrapson:
@mattandrews Resurrected this work from last week I mentioned to model the form in a class, allows us to move things like validation, progress,  pagination methods into the form and also add some helper methods like `form.getCurrentFields()` and `form.getCurrentFieldsForStep` to avoid awkward flat-mapping everywhere.

I've kept the commits fairly granular so you can hopefully see the steps taken.

I think this even avoids calculating validation so many times. We previously did this once to work out your progress and again in `validateForm` where now we can store the validation state when building the form as `form.validation` and re-use that for progress.


controllers/apply/awards-for-all/fields.js
```diff
@@ -2067,29 +2067,5 @@ module.exports = function fieldsFor({ locale, data = {} }) {
         }
     };
 
-    const schema = Joi.object(
```
davidrapson: This is now computed by the form model rather than needing to do this per form and pass properties all the way down.

controllers/apply/awards-for-all/mocks.js
```diff
@@ -1,121 +0,0 @@
-'use strict';
```
davidrapson: This has been inlined into `forms.test` as it's only used there.

PR #1991, opened 2019-06-14
Update organisation types

mattandrews:
This changes the org type fields a little:

- Splits out College/University from School
- Tweaks some of the Senior Contact Roles based on org types
- Adds a new `organisationSubType` field specifically for Statutory Bodies
- Customises the Role field to be free text if the `organisationSubType` contains specific values

This last item has a slight edge case side effect – if you choose a regular org type (eg. not Statutory Body), you get a list of Senior Contact Roles like this:

image

If you then pick one of these, but later go back and change to a Statutory Body and then pick one of the three free text options (eg. Prison Service, Fire Service or Police Authority), you get this when you return to the Senior Contact Role field:

image

Not sure what we can do about this besides making the free text input a different field `name` altogether and merging it at the end of the submission (eg. during the enrichment process). This feels a bit hacky though.

There's some slightly repetitive logic in here but I couldn't think of a better way of handling it in case we need other org sub-types in the future with their own logic. Thoughts very welcome!


controllers/apply/awards-for-all/fields.js
```diff
@@ -428,27 +441,23 @@ module.exports = function fieldsFor({ locale, data = {} }) {
     }
 
     function seniorContactRoleField() {
-        function rolesFor(organisationType) {
+        function rolesFor(organisationType, organisationSubType) {
             const ROLES = {
```
mattandrews: Main changes here are to add/remove a few, but also to alphabetise the object to make future editing a bit easier.

controllers/apply/awards-for-all/fields.js
```diff
@@ -653,17 +746,9 @@ module.exports = function fieldsFor({ locale, data = {} }) {
...
-                    } ${
-                        this.settings.minStart.units
-                    } from when you plan to submit your application. If your project is a one-off event, please tell us the date of the event.</p>
-                    <p><strong>For example: ${
-                        this.settings.minDateExample
-                    }</strong>.</p>
-                    <p>The project end date must be within ${
-                        this.settings.maxDurationFromStart.label
-                    }
+                    en: `<p>The start date needs to be at least ${this.settings.minStart.amount} ${this.settings.minStart.units} from when you plan to submit your application. If your project is a one-off event, please tell us the date of the event.</p>
```
mattandrews: We possibly have some sort of discrepancy between our editorconfigs here because these changes got made automatically when I saved...?!

controllers/apply/awards-for-all/fields.js
```diff
@@ -1731,9 +1800,75 @@ module.exports = function fieldsFor({ locale, data = {} }) {
             })
         }),
         organisationType: organisationTypeField(),
+        organisationSubTypeStatutoryBody: {
```
mattandrews: My thinking here was that we could add more fields like this if other org types needed sub-types, and just give them the same field name (eg. `organisationSubType`).

mattandrews:
@davidrapson I've tweaked this now to use a simplified free text logic.

PR #1992, opened 2019-06-14
Allow user to continue without using a looked-up address

mattandrews:
Fixes this bug:

postcode-error

eg. if you look up an address but don't find the one you want, you should be able to manually enter one without the address dropdown still being `required` – the fix was simply to apply the same requirement logic as the postcode input field itself.


mattandrews:
@davidrapson good call, just made it a computed property.

PR #1994, opened 2019-06-14
Update contacts section

mattandrews:
This swaps the order of a few fields, tweaks some labels/explanations, and convert the list of communication needs into a free text box.


controllers/apply/awards-for-all/form.js
```diff
@@ -452,11 +452,54 @@ module.exports = function({ locale, data = {} }) {
                             en: 'Who is your senior contact?',
                             cy: ''
                         }),
-                        introduction: localise({
-                            en: `<p>Please give us the contact details of a senior member of your organisation.</p>
-                            <p>Your senior contact must be at least 18 years old and is legally responsible for ensuring that this application is supported by the organisation applying, any funding is delivered as set out in the application form, and that the funded organisation meets our monitoring requirements.</p>`,
-                            cy: ``
-                        }),
+                        get introduction() {
```
davidrapson: Wonder if something like this makes it easier to see the different parts of this copy:

```js
get introduction() {
    function roleText() {
        let result;
        switch (currentOrganisationType) {
            case ORGANISATION_TYPES.UNINCORPORATED_REGISTERED_CHARITY:
            case ORGANISATION_TYPES.CIO:
            case ORGANISATION_TYPES.NOT_FOR_PROFIT_COMPANY:
                result = localise({
                    en: `<p>This person must be a member of your board or committee.</p>`,
                    cy: ''
                });
                break;
            case ORGANISATION_TYPES.SCHOOL:
                result = localise({
                    en: `<p>If you are a school, this person must be the headteacher.</p>`,
                    cy: ''
                });
                break;
            default:
                result = localise({
                    en: `<p>This person is usually a senior leader, or a member of your board or committee.</p>`,
                    cy: ''
                });
                break;
        }

        return result;
    }

    return [
        localise({
            en: `<p>Please give us the contact details of a senior member of your organisation.</p>`,
            cy: ``
        }),
        roleText(),
        localise({
            en: `<p>Your senior contact must be at least 18 years old and is legally responsible for ensuring that this application is supported by the organisation applying, any funding is delivered as set out in the application form, and that the funded organisation meets our monitoring requirements.</p>`,
            cy: ''
        })
    ].join('\n');
}
```

`currentOrganisationType` is already defined at the top of this file so we can avoid the extra get(  'organisationType')(data)`.

mattandrews: Good call, thanks! Just pushed that change now.

PR #1995, opened 2019-06-14
Make T&Cs a first-class section of the form

mattandrews:
This is still WIP – the tests need updating to handle this.


controllers/apply/awards-for-all/terms.js
```diff
@@ -0,0 +1,87 @@
+'use strict';
+module.exports = {
```
davidrapson: As this is only copy this could probably move back to the locale files

davidrapson:
Mind popping these extra fields in the schema documentation? At the least the name and role might get used.

PR #2000, opened 2019-06-17
Make sure lint runs in ci

davidrapson:
Spotted that `eslint` wasn't running in CI, we don't have many rules here and it runs elsewhere but better to include it here.


davidrapson:
🎉 Pull request 2000

PR #2002, opened 2019-06-19
Fix AFA snags

mattandrews:
Quite a few things in here so will call them out individually.

## Fixes issues with "other" beneficiaries
It adds this line of copy (eg. to flag up the "other" field on the next page)
image

... then on the subsequent page, it removes the requirement for a checkbox to be ticked when the "other" box is completed (otherwise this page can never be completed).

## Modifies error appearances
Errors now show the label text as well as the message, and make the hyperlinks more obvious:
image

## Adds new navbar
image

## Adds minimum budget size
image

## Adds `organisationStartDate` field
This is used to decide whether or not to show the accounting date/total questions later on.
image

## Adds AFA link to user dashboard
image

## Misc/minor fixes
- Makes the "Are you sure you wish to leave this page?" popup/check work with the Vue character count fields
- Adds some missing legal/DP text to the T&Cs page by introducing a `footer` for fieldsets
- Fixes a [brilliant typo](https://github.com/biglotteryfund/blf-alpha/commit/e83dd3875c25e066cff6a839d90ca712cd852c21#diff-51b0f487ce60156d9cc446cfcca28af0L65)
- Logs in users after changing password
- Adds Main/Senior contact duplicate validation (email/name/address)
- Adds new bank account details validation
- Extracts first/last name fields into a combined component to allow shared validation (eg. main/senior contacts)


controllers/apply/form-router-next/joi-extensions/month-year.js
```diff
@@ -0,0 +1,79 @@
...
+                },
+                validate(params, value, state, options) {
+                    const minDate = moment().subtract(
+                        params.amount,
+                        params.units
+                    );
+                    const isBeforeMin = valueToDate(value).isSameOrBefore(
+                        minDate
+                    );
+                    value.isBeforeMin = isBeforeMin;
```
mattandrews: This feels a bit off – I'm effectively making this a computed property so we can use it in validation later. Otherwise I'm not sure how we can validate a non-date object (eg. `{ month: 3, year: 2000 }`) in Joi directly.

controllers/apply/form-router-next/joi-extensions/bank-numbers.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+const { isString } = require('lodash');
+
+module.exports = function bankNumbers(joi) {
```
davidrapson: Not 100% sure this needs to be written as a custom validator. Curious if `joi.string().replace(/\D/g, '').length(5)` has the same effect? At the least both of the custom rules are repeating what length does, so it feels like there should be a way to avoid that duplication.

davidrapson:
The integration test will fail because we were previously randomly picking a country to run through the test, this will need to be set to scotland.

PR #2005, opened 2019-06-24
 Expand form navigation

mattandrews:
This builds on https://github.com/biglotteryfund/blf-alpha/pull/2002 and makes the component more flexible and less tied to the form. Adds it to the User views in place of breadcrumbs and the "user status" widget.

image


controllers/apply/form-router-next/views/dashboard.njk
```diff
@@ -2,6 +2,7 @@
 {% from "components/content-box/macro.njk" import contentBox %}
 {% from "components/icons.njk" import iconTick, iconBin %}
 {% from "components/user-status/macro.njk" import userStatus with context %}
+{% from "components/user-navigation/macro.njk" import formNavigation with context %}
```
davidrapson: `formNavigation` should probably be `userNavigation` to match the directory / class names.

controllers/tools/views/survey.njk
```diff
@@ -67,13 +67,13 @@
             {% call contentBox() %}
                 <h2>How can we improve this page?</h2>
                 {% for response in survey.noResponses %}
-                    <artible id="msg-{{ response.id }}">
+                    <article id="msg-{{ response.id }}">
```
davidrapson: 🙈

PR #2007, opened 2019-06-24
AFA copy changes

mattandrews:
Loads of things in here but really it's just working through Scott and Lily's copy change tickets in Trello.


views/components/user-status/macro.njk
```diff
@@ -1,17 +1,8 @@
 {% macro userStatus(user) %}
     {% if user %}
-        <div class="user-status">
```
davidrapson: Where has this part gone? Is this in the new user nav now?

mattandrews: Yeah, that nav defaults to this same info (logged-in username and logout link).
image

PR #2009, opened 2019-06-25
Add a message to key AFA pages to clarify where to go

mattandrews:
image

image


controllers/apply/awards-for-all/form.js
```diff
@@ -96,7 +96,8 @@ module.exports = function({ locale, data = {} }) {
                         legend: localise({ en: 'Project country', cy: '' }),
                         fields: [fields.projectCountry]
                     }
-                ]
+                ],
+                showAfaWarning: true
```
mattandrews: This is a bit rubbish but couldn't think of another way to consistently show the same style of info/message (rather than dumping it in the explanation etc).

davidrapson: We could maybe have a `message` key, so something like:

```
{
  title: "Applying for a project in England, Northern Ireland or Wales?",
  body: "<a href="https://apply.tnlcommunityfund.org.uk">You’ll need to use this form instead</a>. Only applicants in Scotland can apply through our new online form at the moment. We’re working on making this available for the rest of the UK."
}
```

Could share that as a variable within the form builder. Am I right that the "full" message is only used on account screens and this shorter one is used in the form?

mattandrews: Yep that's right.

PR #2013, opened 2019-06-26
Small screen tweaks

davidrapson:
A few small screen improvements. Mostly tweaks to avoid layout wrapping, but notably improves the display of the new user navigation:

image

This PR also updates how the user navigation links are modelled so they are shared in controllers rather than defined each time in templates.


davidrapson:
[Review form display on smaller screens](https://trello.com/c/SJf9tinm/490-review-form-display-on-smaller-screens)

PR #2022, opened 2019-06-26
Use Bank Account Checker API to compare sort codes / account numbers

mattandrews:
This means a valid bank account is required to use the form. A pair of non-matching sort code / account numbers will fail validation.

On the other hand, if the API call fails for any other reason (eg. downtime, token expiry, etc) we treat it as a success and proceed.

Generally this adds the concept of a `preflightCheck`, which is an async function on a `step` which returns a `Promise`. If the function throws an error (eg. returns an array of error objects) then these are rendered back to the user:

image

https://trello.com/c/wAiZeexS/534-use-bank-account-checker-to-validate-bacs-is-allowed


mattandrews:
For testing, here are some (test) account details that are valid:

```
{
    sortCode: 308087,
    bankAccount: 25337846
}
```

common/bank-api.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+const request = require('request-promise-native');
+const { BANK_API } = require('../common/secrets');
+
+const checkBankAccountDetails = (sortCode, accountNumber) => {
```
davidrapson: As this is on a singe function I'd probably export it directly

```js
module.exports = function checkBankAccountDetails(sortCode, accountNumber) { … }
```

controllers/apply/awards-for-all/form.js
```diff
@@ -620,7 +623,76 @@ module.exports = function({ locale, data = {} }) {
...
+                preflightCheck: function() {
+                    const sortCode = get('bankSortCode')(data);
+                    const accountNumber = get('bankAccountNumber')(data);
+
+                    return new Promise((resolve, reject) => {
+                        checkBankAccountDetails(sortCode, accountNumber)
+                            .then(status => {
+                                if (status.code === 'UNKNOWN') {
+                                    // If this API does anything weird, assume all is well
+                                    logger.info(
```
davidrapson: Mind adding a tag for the logger like in https://github.com/biglotteryfund/blf-alpha/pull/2020 ?

controllers/apply/awards-for-all/form.js
```diff
@@ -620,7 +623,76 @@ module.exports = function({ locale, data = {} }) {
...
+                                    ]);
+                                }
+                                // Otherwise everything is all good
+                                return resolve();
+                            })
+                            .catch(err => {
+                                logger.info(
+                                    'User bank details check: API call failed',
+                                    {
+                                        error: err
```
davidrapson: Would it be better to capture the error part in Sentry?

controllers/apply/awards-for-all/form.js
```diff
@@ -620,7 +623,76 @@ module.exports = function({ locale, data = {} }) {
...
                         ]
                     }
-                ]
+                ],
+                preflightCheck: function() {
+                    const sortCode = get('bankSortCode')(data);
+                    const accountNumber = get('bankAccountNumber')(data);
+
+                    return new Promise((resolve, reject) => {
+                        checkBankAccountDetails(sortCode, accountNumber)
```
davidrapson: Could make this async/await

controllers/apply/awards-for-all/form.js
```diff
@@ -624,12 +629,12 @@ module.exports = function({ locale, data = {} }) {
...
                 ],
-                preflightCheck: function() {
+                preflightCheck: async function() {
                     const sortCode = get('bankSortCode')(data);
                     const accountNumber = get('bankAccountNumber')(data);
 
-                    return new Promise((resolve, reject) => {
-                        checkBankAccountDetails(sortCode, accountNumber)
+                    return new Promise(async (resolve, reject) => {
+                        await checkBankAccountDetails(sortCode, accountNumber)
```
davidrapson: This syntax isn't correct, needs to be a `try/catch` if using async await.

PR #2025, opened 2019-06-27
Rewrite parameter store fetching to account for pagination

davidrapson:
Reverts biglotteryfund/blf-alpha#2024

Needs `GetParametersByPath` permissions in IAM role before this can be re-deployed.

### DIFF
--- a/bin/get-secrets
+++ b/bin/get-secrets
@@ -1,9 +1,10 @@
 #!/usr/bin/env node
 'use strict';
+/* eslint-disable no-console */
 const fs = require('fs');
 const assert = require('assert');
 const AWS = require('aws-sdk');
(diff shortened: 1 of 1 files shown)


mattandrews:
Wonder if we should start running the app locally as the same user (or with the same policies) as the web user, rather than all-access admins.

davidrapson:
Closing for now as we've pruned any unused secrets down to ~40 items so we have some headroom to fix this and would rather avoid having a branch hanging around until we're able to properly review and update IAM permissions to support this new way of fetching secrets.

Added an issue to the backlog to resolve this soon though https://github.com/biglotteryfund/blf-alpha/issues/2114

PR #2027, opened 2019-06-27
🌎 Upgrade client-side dependencies

davidrapson:
Upgrade client-side dependencies

- Babel versions
- Webpack versions
- Vue versions
- CSS compilation dependencies

Needs cross browser testing before this can be merged


webpack.config.js
```diff
@@ -26,7 +26,11 @@ module.exports = {
         ]
     },
     entry: {
-        app: './assets/js/main.js'
+        app: [
```
davidrapson: Added based on https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import#working-with-webpack-and-babel-preset-env

We load a polyfill for `Promise` via polyfill.io but this seems more explicit. If this tests OK we can remove the version from polyfill.io

mattandrews:
Seems fine to me if it passes!

davidrapson:
Ran through cross browser tests on this and all looks good.

PR #2030, opened 2019-06-28
Upgrade cypress and related dependencies

davidrapson:
A new version of Cypress has been released with fixes for the performance regressions in `v3.3`. Testing upgrading to the latest version to confirm if this updated fixes the slowdown seen in  https://github.com/biglotteryfund/blf-alpha/pull/1970


davidrapson:
Confirmed that this fixes the slow down. Had to re-run the build to account for the new binary, but once the cache has been stored the build time is the same as before.

PR #2031, opened 2019-06-28
Refine user logging

davidrapson:
Adds a `service` tag to user logs and refined error logging based on https://github.com/winstonjs/winston/blob/master/examples/errors.js

### DIFF
--- a/common/logger.js
+++ b/common/logger.js
@@ -1,5 +1,5 @@
 'use strict';
-const winston = require('winston');
+const { createLogger, format, transports } = require('winston');
 const WinstonCloudWatch = require('winston-cloudwatch');
 
 const appData = require('./appData');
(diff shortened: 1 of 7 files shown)


common/logger.js
```diff
@@ -1,27 +1,23 @@
...
 const transport = skipLogs
-    ? new winston.transports.Console({
-          format: winston.format.simple()
-      })
+    ? new transports.Console()
     : new WinstonCloudWatch({
           awsRegion: 'eu-west-2',
           logGroupName: `/tnlcf/${appData.environment}/app`,
-          logStreamName: appData.buildNumber,
+          logStreamName: `build-${appData.buildNumber}`,
```
davidrapson: Added a `build-` prefix to remind us what the log stream name means.

PR #2032, opened 2019-06-28
Step cleanup

davidrapson:
A couple of bits of step clean up.

- Wraps the file upload code in an `if` statement. This was the implied effect before as the list of upload promises would be empty but this is more explicit.
- Removes the concept of a section introduction. We introduced this briefly for the beneficiaries section but we're not using the functionality any more.
- Adds tests for file upload prep code


controllers/apply/form-router-next/index.js
```diff
@@ -427,24 +425,22 @@ function initFormRouter({
...
     router.route('/download/:fieldName/:filename').get((req, res, next) => {
         const { currentlyEditingId, currentApplicationData } = res.locals;
 
-        // Check that this application has data for the requested field name
         const fileData = currentApplicationData[req.params.fieldName];
+        const matchesField =
+            fileData && fileData.filename === req.params.filename;
 
-        // Confirm that the requested filename matches this field's file
-        if (fileData && fileData.filename === req.params.filename) {
```
mattandrews: Do you think the comments here weren't adding very much value? I appreciate they're only mirroring what the code is doing, but for me it helps parse what's going on (eg. the intention of these checks) rather than having to read through the logic and try to remember which variables come from where etc.

davidrapson: I typically find that these kind of comments are better replaced with a variable or function name with the intent of the comment. Tried to do that by assigning it to `matchesField`.

PR #2036, opened 2019-07-02
Add content type "People Stories" to News section

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/176

This shares the blogpost logic to expand into the Updates content type formerly known as "News", now renamed People Stories. They display the same as blogposts (minus the category field).

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -480,7 +480,7 @@ news:
   readMoreAboutProgramme: Read more about this programme
   blogpost:
     viewAllBlogPosts: View all blog posts
-  peopleStories:
+  people-stories:
     viewAllPeopleStories: View all People Stories
   pressRelease:
     region: Region
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -506,7 +506,7 @@ news:
     blog: >
       Looking for colour, context and perspective about the work of The National Lottery Community Fund, its impact on communities and topics currently in the news? It’s all here. This blog will explain, explore, showcase and examine the ways in which civil society, and as a result the sector, is changing. Read about what people in communities across the UK and sector thought leaders have to say and form your own opinion.
     people-stories: >
-      TBD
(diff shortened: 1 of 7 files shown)


controllers/updates/index.js
```diff
@@ -6,7 +6,11 @@ const { concat, isArray, pick, get } = require('lodash');
 const { buildPagination } = require('../../common/pagination');
```
mattandrews: Most of the changes here are whitespace, FYI

controllers/updates/views/view-helpers.njk
```diff
@@ -11,7 +11,7 @@
     {%- endfor %}
 {% endmacro %}
 
-{% macro blogTrail(entry, promoted = false, showAuthor = true) %}
+{% macro blogTrail(entry, promoted = false, showAuthor = true, updateType = 'blog') %}
```
davidrapson: We could maybe rename this macro at some-point to something more generic.

config/locales/en.yml
```diff
@@ -502,6 +505,8 @@ news:
       Our press releases will give you the key facts about funding announcements, new grants or research, special events, launches, which charities have been awarded funding, innovative projects, best practice in the sector and more.
     blog: >
       Looking for colour, context and perspective about the work of The National Lottery Community Fund, its impact on communities and topics currently in the news? It’s all here. This blog will explain, explore, showcase and examine the ways in which civil society, and as a result the sector, is changing. Read about what people in communities across the UK and sector thought leaders have to say and form your own opinion.
+    people-stories: >
```
davidrapson: This is `peopleStories` earlier but `people-stories` here, is camel cased for both possible?

mattandrews: Ideally the other way around, as I'm using the lowercase one (supplied via URL params) as a variable everywhere else.

PR #2042, opened 2019-07-03
Refactor postcode lookup

mattandrews:
This switches things around logic-wise so that most of the decision-making happens within the Vue components rather than wiring them up with existing HTML. This means duplicated markup for non-JS versus JS (a pattern we've used on the past grants components), but avoids the complexity introduced by this around fields being required/validated when not onscreen.


assets/js/vue-apps/form-components/address-line.vue
```diff
@@ -0,0 +1,33 @@
...
+    computed: {
+        required() {
+            return this.isRequired === true;
+        }
+    }
+};
+</script>
+
+<template>
+    <div class="ff-address__field">
```
davidrapson: Minor nit,  but having a child `ff-address__field` as the root class suggests this should be it's own top-level class name, or move the class to `address-lookup.vue`  (doesn't need changing now though)

davidrapson:
I'm not sure why snyk is complaining. Mentions a vulnerability in `nunjucks`. There is an update available so maybe worth making that to see if it sorts things, otherwise I think it's safe to ignore for this PR and we can review it separately as you're not touching any dependencies here.

davidrapson:
Think CI is failing because of a small lint error

PR #2043, opened 2019-07-03
Update test deps

davidrapson:
Updates `eslint` to the latest version. Includes some additional lint rules. Notably this one https://eslint.org/docs/rules/no-async-promise-executor. Main change here is to clean up the bank API lookup syntax to fix this warning.


assets/js/vue-apps/past-grants.js
```diff
@@ -232,7 +268,7 @@ function init(STORAGE_KEY) {
                 this.trackUi('Sort results', newSort);
             },
 
-            updateResults(urlPath, queryString = null) {
+            updateResults(urlPath, currentQueryString = null) {
```
davidrapson: This is the change here, renaming `queryString` to `currentQueryString` to avoid shadowing the query string module above. The rest is prettier autoformatting.

PR #2044, opened 2019-07-03
Merge loggers

davidrapson:
For our use case `winston` and `debug` are preforming similar purposes. This PR moves the handful of `debug` calls we have  to use `winston`. Adds a default console formatter to get readable output in development

Screenshot 2019-07-03 at 11 37 24

### DIFF
--- a/common/logger.js
+++ b/common/logger.js
@@ -16,7 +16,8 @@ function transport() {
         });
     } else {
         return new transports.Console({
-            silent: !!process.env.TEST_SERVER === true
+            silent: !!process.env.TEST_SERVER === true,
+            format: format.simple()
         });
     }
 }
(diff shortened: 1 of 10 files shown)


common/logger.js
```diff
@@ -16,13 +16,14 @@ function transport() {
...
+            silent: !!process.env.TEST_SERVER === true,
+            format: format.simple()
         });
     }
 }
 
 module.exports = createLogger({
-    level: 'info',
+    level: isDev ? 'debug' : 'info',
     format: format.combine(format.errors({ stack: true }), format.json()),
```
mattandrews: You could add `format.colorize()` to the `combine()` call here to get something similar to what `debug` did:
image

PR #2045, opened 2019-07-03
Add nested Strategic Programme pages

mattandrews:
This pairs with https://github.com/biglotteryfund/craft-dev/pull/177 and allows the rendering of child pages within Strategics, similar to Funding Programmes.

It also makes a tweak to Funding Programmes to avoid URL spoofing, eg. currently it's possible to rewrite this:

https://www.tnlcommunityfund.org.uk/funding/programmes/empowering-young-people/meet-our-young-people

to this:

https://www.tnlcommunityfund.org.uk/funding/programmes/any-old-slug/meet-our-young-people

because it's not possible to filter on parent slugs in Craft (without looking up the parent entry directly). 

The changes here mean we can make top-level funding/strategic programme pages which are plain "Content Pages", as well as making them children of strategics.

### DIFF
--- a/controllers/programmes/index.js
+++ b/controllers/programmes/index.js
@@ -1,5 +1,5 @@
 'use strict';
-const { concat, get, groupBy, head, map, set } = require('lodash');
+const { concat, get, groupBy, head, map, set, startsWith } = require('lodash');
 const express = require('express');
 const path = require('path');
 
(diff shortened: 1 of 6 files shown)


controllers/programmes/index.js
```diff
@@ -208,51 +208,75 @@ router.get(
...
-                {
-                    breadcrumbs: concat(res.locals.breadcrumbs, [
+        const { currentPath, fundingProgramme } = res.locals;
+
+        /* Only render if not using an external path */
+        if (
+            fundingProgramme &&
+            startsWith(currentPath, fundingProgramme.linkUrl)
+        ) {
+            if (get(fundingProgramme, 'entryType') === 'contentPage') {
```
mattandrews: All of the logic inside this statement is duplicates in the Strategics controller – I tried to extract it somewhere shared but couldn't work out where to put it... `/common`?

controllers/strategic-investments/index.js
```diff
@@ -1,28 +1,84 @@
...
     injectStrategicProgramme,
-    injectStrategicProgrammes
+    injectStrategicProgrammes,
+    setCommonLocals
 } = require('../../middleware/inject-content');
 
 const router = express.Router();
 
-router.get('/', injectListingContent, injectBreadcrumbs, injectStrategicProgrammes, function(req, res) {
+router.use(injectBreadcrumbs, (req, res, next) => {
```
mattandrews: I nicked this idea from the funding programmes controller, meaning that the API now [doesn't need to supply breadcrumbs](https://github.com/biglotteryfund/craft-dev/pull/177/files#diff-0c5616765ab2e0289f6948e349af3e37L25), which allows non-child pages to still get a valid section crumb (which they didn't otherwise).

davidrapson: Nice, when I first added that I was half thinking that the API could supply breadcrumbs but in practice it's almost always made more sense for the app to do it. Thanks for cleaning this up.

PR #2048, opened 2019-07-03
A11y fixes

davidrapson:
Fixes two of the simpler accessibility audit suggestions:

- High Priority: (DAC-KEYBOARD-03) - Add focusable=false attribute to SVG icons
- High Priority: (DAC-TITLES-01) - Add step number to page titles


davidrapson:
[High Priority: (DAC-TITLES-01) - Add step number to page titles](https://trello.com/c/OMbbuOZr/619-high-priority-dac-titles-01-add-step-number-to-page-titles)

davidrapson:
[High Priority: (DAC-TITLES-01) - Add step number to page titles](https://trello.com/c/OMbbuOZr/619-high-priority-dac-titles-01-add-step-number-to-page-titles)

davidrapson:
[High Priority: (DAC-KEYBOARD-03) - Add focusable=false attribute to SVG icons (do globally)](https://trello.com/c/tMSCX6mG/616-high-priority-dac-keyboard-03-add-focusablefalse-attribute-to-svg-icons-do-globally)

controllers/apply/awards-for-all/constants.js
```diff
@@ -17,7 +17,8 @@ const ORGANISATION_TYPES = {
     NOT_FOR_PROFIT_COMPANY: 'not-for-profit-company',
     SCHOOL: 'school',
     COLLEGE_OR_UNIVERSITY: 'college-or-university',
-    STATUTORY_BODY: 'statutory-body'
+    STATUTORY_BODY: 'statutory-body',
+    FAITH_GROUP: 'faith-group'
```
mattandrews: This change (and the related ones) isn't strictly part of the a11y fixes but I thought it was easier to batch up with the others to avoid merge conflict hell later.

davidrapson:
"I wasn't sure what we meant about "unique row labels", the report only mentions the buttons. Would we want to add some sort of label to the <div> containing the budget row itself?"

I was reading this part "The data on this page is difficult for screen reader users to understand. Adding text to the elements which uniquely identifies the elements on each row would aid users to complete the form without becoming disorientated." in the report as being generally about the rows, assuming that we might need "Item (Row 1) | Amount (Row 1)" for the input labels but I think you are right it's specific to the delete buttons. This looks good.

PR #2050, opened 2019-07-05
AFA application stats dashboard

mattandrews:
More we can do here, but here's a basic dashboard for now.

blf local_3000_tools_applications_awards-for-all

Allows filtering by country and date.

### DIFF
--- a/controllers/tools/applications.js
+++ b/controllers/tools/applications.js
@@ -120,6 +120,39 @@ function titleCase(str) {
     });
 }
 
+function getColourForCountry(countryName) {
+    let colour = '';
+    switch (countryName) {
+        case 'England':
+            colour = '#f95d6a';
+            break;
+        case 'Northern Ireland':
+            colour = '#2f4b7c';
+            break;
+        case 'Scotland':
+            colour = '#a05195';
+            break;
+        case 'Wales':
+            colour = '#ffa600';
+            break;
+        default:
+            colour = '#e5007d';
+            break;
+    }
+    return colour;
+}
+
+function addCountry(row) {
+    // Convert Sequelize instance into a plain object so we can modify it
+    const data = row.get({
+        plain: true
+    });
+    data.country = data.applicationCountry
+        ? data.applicationCountry
+        : get(data, 'applicationData.projectCountry');
+    return data;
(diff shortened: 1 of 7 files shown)


controllers/tools/applications.js
```diff
@@ -0,0 +1,235 @@
...
+    return minMaxAvg(appDurations);
+}
+
+function measureWordCounts(data) {
+    const wordCounts = data.map(
+        d =>
+            d.applicationSummary
+                .map(_ => _.value)
+                .join(' ')
+                .split(' ').length
```
mattandrews: This is a bit of fun really as it doesn't capture every field (eg. the budget here gets output as `Object object`) but it's a rough approximation, which I like.

controllers/tools/applications.js
```diff
@@ -0,0 +1,235 @@
...
+            let appCountry;
+
+            if (appType === 'pending') {
+                appCountry = get(item, 'applicationData.projectCountry');
+            } else {
+                const rowCountry = item.applicationSummary.find(
+                    _ =>
+                        _.label ===
+                        'What country will your project be based in?'
+                );
```
mattandrews: This is a bit ugly and makes me wonder if this field should be moved to the `applicationOverview` field so we can more easily make queries like this.

controllers/tools/applications.js
```diff
@@ -0,0 +1,235 @@
...
+        };
+    });
+
+    return dayData;
+}
+
+function minMaxAvg(arr) {
+    const sorted = arr.slice().sort((a, b) => a - b);
+    return {
+        lowest: sorted[0] || 0,
```
davidrapson: You could maybe use lodash `min` and `max` to avoid work this out by manually sorting

controllers/tools/applications.js
davidrapson: Love this as long as we're clear that it's approximate.

controllers/tools/applications.js
davidrapson: As discussed you can do `item.applicationCountry` instead.

mattandrews:
Updated to include country-specific charts:

image

PR #2051, opened 2019-07-05
Add GA tracking for various Insights Document links

mattandrews:
We've had some requests for an analytics dashboard about Insights Documents so this adds some specific tracking for document downloads (as opposed to the generic document tracker which is harder to isolate to a specific section).

I also tried adding site search tracking via the `q` parameter in GA to try to work out what searches people use, otherwise we have to use a gnarly regex (`/insights/documents\?(.*)q=[\w\d%]+`) to find pageviews with populated querystrings (and even then it's still part of a URL) – will leave this a few days to see if it picks anything up.

### DIFF
--- a/controllers/insights/views/insights-detail.njk
+++ b/controllers/insights/views/insights-detail.njk
@@ -56,7 +56,9 @@
                     {% if entry.documents %}
                         {{ documentsCard(
                             title = extraCopy.documents,
-                            documents = entry.documents
+                            documents = entry.documents,
+                            gaCategory = 'Insight Documents',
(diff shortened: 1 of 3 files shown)


mattandrews:
Looks like the `q` parameter tracking might already be working: https://analytics.google.com/analytics/web/#/report/content-site-search-pages/a637620w1357527p1354362/explorer-table.plotKeys=%5B%5D&explorer-table.secSegmentId=analytics.pagePath&_r.drilldown=analytics.searchStartPage:www.tnlcommunityfund.org.uk~2Finsights~2Fdocuments

PR #2055, opened 2019-07-08
Disallow other options if everyone is selected

davidrapson:
Adds unit tests and new schema logic to show an error if more than one option is selected when "open to everyone" is selected.

image

PR #2057, opened 2019-07-09
Beneficiaries screener

davidrapson:
To avoid the selection state issues highlighted in https://github.com/biglotteryfund/blf-alpha/pull/2055 we're switching back to a screener question but with revised wording to make sure the question makes sense in context for assistive devices.

image

Screenshot 2019-07-09 at 10 45 01


cypress/support/commands.js
```diff
@@ -113,22 +113,7 @@ Cypress.Commands.add('checkA11y', () => {
...
         .then(window => {
-            return window.axe.run(
-                {
-                    include: window.document,
-                    exclude: [['iframe']]
-                },
-                {
-                    rules: {
-                        // @TODO: Review and re-enable this
-                        'color-contrast': { enabled: false },
```
davidrapson: I've removed the default excludes in favour of overriding on a per-case basis. We should only be excluding the colour contrast check for known exceptions.

cypress/support/commands.js
```diff
@@ -113,22 +113,7 @@ Cypress.Commands.add('checkA11y', () => {
...
-                        // New ARIA complementary rule. Review and then re-enable
-                        'landmark-complementary-is-top-level': {
```
davidrapson: I'd suppressed this rule when it was first introduced as I needed to review the recommendations first. The gist is our use of `<aside>` is broadly incorrect. Not a major accessibility issues but can cause some confusion. `<aside>` can be used outside `<main>` e.g. for related posts / additional content. But shouldn't be used inside `<main>` for "callout" style content.

views/components/content-tabs/macro.njk
```diff
@@ -3,10 +3,10 @@
 {% macro contentTabs(tabs) %}
     {# data-breakpoint must match breakpoint used for styles #}
     <div class="js-tab-container" data-breakpoint="(min-width: 620px)">
-        <ul role="tablist" class="js-tabset content-tabset">
```
davidrapson: This was being flagged as an invalid combination of roles when I removed the rule exclusions in the aXe config. Doing a bit of digging it sounds like you have to be careful with the `tablist` role and it's better to leave as a plain list.

cypress/integration/awards-for-all.js
```diff
@@ -0,0 +1,481 @@
+const faker = require('faker');
```
davidrapson: It was getting to be a pain to constantly add an `.only` in these tests to run just the awards for all spec so I've split it out into a separate file which Cypress treats as a dedicated option in the UI.

I'd originally kept everything in a single file for performance reasons but in testing it looks like more recent versions of Cypress have improved the delay between suites so we can stand to break things out a little more.

davidrapson:
@mattandrews Mind giving this another review? Revised the copy a little bit and made some testing improvements including a couple of accessibility test tweaks.

PR #2060, opened 2019-07-10
Validation logging

davidrapson:
Adds logging to steps for any validation errors. Allows us to report on what validation messages are most often seen.

image


controllers/apply/form-router-next/views/step.njk
```diff
@@ -23,7 +23,7 @@
                   method="post"
                   class="js-form-warn-unsaved"
                   {% if errors | length > 0 %}data-form-has-errors="1"{% endif %}
-                  {% if stepIsMultipart %}enctype="multipart/form-data"{% endif %}
+                  {% if step.isMultipart %}enctype="multipart/form-data"{% endif %}
```
davidrapson: We can avoid the extra view variable by calling `step.isMultipart` here.

PR #2062, opened 2019-07-11
Warnings and messages

davidrapson:
This PR does some work to improve how we handle messages and warnings in the form model.

**Merge role type messages**

Some warning text for the senior contact role was handled in the step introduction and the rest was handled in the field explanation.

duplication-conditions

Added the concept of `warnings` on fields. Rationalised the logic for these messages into a single list of warnings and added tests.

**Move contact surname warning to field**

Also, moves the contact surname to the new `warnings` property and adds tests.

**Simplify featured errors logic**

All errors in the current allow list have `includeBaseError` set to `false`. This means that for now we can make the assumption that all errors in this allow-list will exclude the base error. Keen to avoid having a code path that is not being executed.

I also spotted that we can avoid the mutation and regular filtering required by setting an isFeatured flag on the message by returning a dedicated featuredMessages property from the validation function.

### DIFF
--- a/controllers/apply/awards-for-all/fields.js
+++ b/controllers/apply/awards-for-all/fields.js
@@ -564,109 +564,117 @@ module.exports = function fieldsFor({ locale, data = {} }) {
             return options;
         }
 
+        /**
+         * Statutory bodies require a sub-type,
+         * some of which allow free text input for roles.
+         */
+        function isFreeText() {
+            return (
+                currentOrganisationType === ORGANISATION_TYPES.STATUTORY_BODY &&
+                [
+                    STATUTORY_BODY_TYPES.PRISON_SERVICE,
+                    STATUTORY_BODY_TYPES.FIRE_SERVICE,
+                    STATUTORY_BODY_TYPES.POLICE_AUTHORITY
+                ].includes(currentOrganisationSubType)
+            );
+        }
+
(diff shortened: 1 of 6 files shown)


cypress/integration/integration_spec.js
```diff
@@ -176,7 +176,8 @@ describe('interactions', () => {
     it('should perform common interactions', () => {
         cy.visit('/funding/programmes/national-lottery-awards-for-all-england');
         cy.get('.cookie-consent button').click();
-        cy.checkA11y();
+        // Temporarily disable until heading-level is fixed
+        // cy.checkA11y();
```
davidrapson: There's a heading-level warning from a recent update to this page. To avoid a misleading failure here I've temporarily disabled this but have requested an edit to the page to fix the warning so will re-enable before merging this.

We might want to keep an eye on this and determine if we're OK with this kind of warning failing the build. Accessibility of the site is an important enough principle that I think it's worth being bullish on, but tricky when the content comes from the CMS and can block a build.

mattandrews: Yeah, I'd be inclined to not let this fail the build (for our own productivity) but perhaps to make it a warning somehow so we can flag it? Maybe there's a way to make Craft respect these settings too...

PR #2069, opened 2019-07-15
Alias cleanup

davidrapson:
Tiny bit of alias cleanup. Lifts the alias query logic up into `content-api`. Adds a unit test for `pathCouldBeAlias`


common/content-api.js
```diff
@@ -250,10 +275,7 @@ function getMerchandise(locale, showAll = false) {
 }
 
 module.exports = {
-    mapAttrs,
```
davidrapson: `mapAttrs` and `mergeWelshBy` are never used outside of this file so we can avoid the awkard export comment

PR #2070, opened 2019-07-15
Fix issue with beneficiaries groups validations

davidrapson:
Spotted an issue where the beneficiary groups section was not showing the correct errors on the summary screen making it hard to tell what you had to complete. The submit button check works as expected, but the logic to determine the correct messages wasn't quite right.

Added a failing unit test to validate this and then updated the field schema. The source of the issue was the conditional validation for `beneficiariesGroupsOther`

### DIFF
--- a/controllers/apply/form-router-next/lib/form-model.js
+++ b/controllers/apply/form-router-next/lib/form-model.js
@@ -86,7 +86,7 @@ class FormModel {
 
             function sectionStatus() {
                 const fieldNames = fieldsForSection().map(f => f.name);
-                const fieldData = pick(validation.value, fieldNames);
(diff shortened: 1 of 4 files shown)


controllers/apply/form-router-next/lib/form-model.js
```diff
@@ -86,7 +86,7 @@ class FormModel {
 
             function sectionStatus() {
                 const fieldNames = fieldsForSection().map(f => f.name);
-                const fieldData = pick(validation.value, fieldNames);
+                const fieldData = pick(data, fieldNames);
```
davidrapson: Using `validation.value` here was preventing some conditional messages from showing, we want to use the original data here.

PR #2071, opened 2019-07-15
Redundant settings

davidrapson:
Does a little work to remove some redundant settings from fields. I've kept the commits pretty granular so you can follow along with what's being changed.

**Remove unused recommendedWords**

This was no longer used to we can safely remove it

**Replace use of settings for date field**

The way `settings` was being used on the date range field was a little counter to it's intention. For this field the values in settings were largely being used inside the field definition so were acting as temporary variables. We can avoid some indirection and getters and setters by using regular variables here. I've done this by extracting a `fieldProjectDateRange` function that can house the temporary variables. Everything left in `settings` is now real field settings that are passed on to the view.

**Simplify dateRange validator**

Added unit tests for `dateRange`, reworked `enDateLimit` to be `maxDate` and match the format of `minDate` as it's doing the same validation and inlined `futureEndDate` into the main validator as it's a sensible default.

### DIFF
--- a/controllers/apply/awards-for-all/fields.js
+++ b/controllers/apply/awards-for-all/fields.js
@@ -1,7 +1,8 @@
 'use strict';
 const moment = require('moment/moment');
+const flatMap = require('lodash/flatMap');
 const get = require('lodash/fp/get');
-const { flatMap, includes, values, concat, has } = require('lodash');
+const has = require('lodash/has');
 
 const Joi = require('../form-router-next/joi-extensions');
 const locationsFor = require('./locations');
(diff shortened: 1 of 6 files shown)


controllers/apply/awards-for-all/fields.js
```diff
@@ -300,7 +293,324 @@ module.exports = function fieldsFor({ locale, data = {} }) {
         return { ...defaultProps, ...props };
     }
 
-    function organisationTypeField() {
+    function fieldProjectDateRange() {
```
davidrapson: I reckon it would make sense to extract all the fields into `field*` functions like we do with the `step*` functions in `form.js` but I'd much rather leave that until after we've added the welsh translation and we can see how noisy this file gets.

PR #2073, opened 2019-07-16
Allow previewing awards for all through access cookie

davidrapson:
Adds the ability to preview the new awards for all form in production by settings

Rationale: We might typically use staff-auth to preview access to certain sections but it's not possible to be logged in as both staff and a regular user which makes previewing the awards for all forms impossible. Allowing staff to be logged in as both types of users at the same time would add a lot of complexity so the simplest option for this short-lived preview is to look for a secret cookie value to allow access.

### DIFF
--- a/common/secrets.js
+++ b/common/secrets.js
@@ -8,6 +8,9 @@ const { getSecret } = require('./parameter-store');
 const SESSION_SECRET =
     process.env.SESSION_SECRET || getSecret('session.secret', true);
 
+const AWARDS_FOR_ALL_SECRET =
+    process.env.AWARDS_FOR_ALL_SECRET || getSecret('awardsForAll.secret');
+
 /**
  * Database connection uri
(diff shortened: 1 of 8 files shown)


common/secrets.js
```diff
@@ -8,6 +8,9 @@ const { getSecret } = require('./parameter-store');
 const SESSION_SECRET =
     process.env.SESSION_SECRET || getSecret('session.secret', true);
 
+const AWARDS_FOR_ALL_SECRET =
+    process.env.AWARDS_FOR_ALL_SECRET || getSecret('awardsForAll.secret');
```
davidrapson: This parameter store value needs creating before this can go live.

mattandrews:
Looks like this snippet will do what we need in the address bar (in Chrome at least, not tried other browsers):

`javascript:void(document.cookie="cookiename=value");`

davidrapson:
I've added the secret to parameter store and updated this branch so this is ready to go now.

PR #2076, opened 2019-07-16
Misc AFA fixes

mattandrews:
A few things in here:

- There was an issue where the Address lookup thought an address had already been supplied (because we changed the default address to contain `null` fieldnames). This fixes things so it knows it's not been provided.

- Budget totals now have commas removed automatically, and this also fixes a totalling error where an invalid line budget wasn't being summed properly (eg. 12 + 34 = 1234).

- Fixes an issue with file upload mimetypes which caused a 500 error when the file's type wasn't known (saw this via the pen test data)

- Re-adds the eligibility questions into `/questions`

- Updates templates for `/questions` and adds `conditionalFields()` method to output all fields for a step, regardless of conditions (eg. for the question list):

image

### DIFF
--- a/assets/js/vue-apps/form-components/address-lookup.vue
+++ b/assets/js/vue-apps/form-components/address-lookup.vue
@@ -50,7 +50,7 @@ export default {
         this.$root.$on('update:conditionalRadio', value => {
             if (value === 'yes') {
                 this.currentState = states.NotRequired;
-            } else if (this.fullAddress) {
+            } else if (this.fullAddressPreview !== '') {
                 this.currentState = this.states.AlreadyAnswered;
             } else {
                 this.currentState = states.NotAsked;
--- a/controllers/apply/form-router-next/joi-extensions/budget-items.js
+++ b/controllers/apply/form-router-next/joi-extensions/budget-items.js
@@ -44,7 +44,10 @@ module.exports = function budgetItems(joi) {
                     maxBudget: joi.number().required()
                 },
                 validate(params, value, state, options) {
-                    const total = sumBy(value, item => item.cost);
+                    const total = sumBy(
+                        value,
+                        item => parseInt(item.cost) || 0
+                    );
                     if (total > params.maxBudget) {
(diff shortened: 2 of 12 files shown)


assets/js/vue-apps/form-components/address-lookup.vue
```diff
@@ -50,7 +50,7 @@ export default {
         this.$root.$on('update:conditionalRadio', value => {
             if (value === 'yes') {
                 this.currentState = states.NotRequired;
-            } else if (this.fullAddress) {
+            } else if (this.fullAddressPreview !== '') {
```
davidrapson: Good catch

controllers/apply/form-router-next/joi-extensions/budget-items.js
```diff
@@ -44,7 +44,10 @@ module.exports = function budgetItems(joi) {
                     maxBudget: joi.number().required()
                 },
                 validate(params, value, state, options) {
-                    const total = sumBy(value, item => item.cost);
+                    const total = sumBy(
+                        value,
+                        item => parseInt(item.cost) || 0
```
davidrapson: parseInt needs a radix of `10` as the second argument. Same for budget-total-costs.js

controllers/apply/form-router-next/lib/formatters.js
```diff
@@ -130,9 +130,9 @@ function formatBudget(value) {
 
 function formatFile(value) {
     if (value) {
-        return `${value.filename} (${mime
-            .extension(value.type)
-            .toUpperCase()}, ${filesize(value.size, { round: 0 })})`;
+        const mimeType = mime.extension(value.type) || 'File';
+        const fileSize = filesize(value.size, { round: 0 });
+        return `${value.filename} (${mimeType.toUpperCase()}, ${fileSize})`;
```
davidrapson: Should it be allowed through if the mimetype is unknown in the first place?

davidrapson:
👍 Changes look good

PR #2081, opened 2019-07-18
Add AV checking by examining file tags (added by ClamAV lambda)

mattandrews:
This relies on the lambda function I've set up which scans all our form S3 buckets (dev, test and prod) and tags suspected files as infected. If these tags are found, the file is blocked from being downloaded by users, and not uploaded to Salesforce.

Update: new error message.
Screenshot 2019-07-22 11 03 44

### DIFF
--- a/controllers/apply/form-router-next/index.js
+++ b/controllers/apply/form-router-next/index.js
@@ -25,7 +25,11 @@ const { requireActiveUser } = require('../../../middleware/authed');
 const { injectCopy } = require('../../../middleware/inject-content');
 
 const salesforceService = require('./lib/salesforce');
-const { getObject, buildMultipartData } = require('./lib/file-uploads');
+const {
+    getObject,
+    buildMultipartData,
+    checkAntiVirus
+} = require('./lib/file-uploads');
 
 function initFormRouter({
     formId,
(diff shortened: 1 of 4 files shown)


controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -109,6 +109,31 @@ function uploadFile({ formId, applicationId, fileMetadata }) {
...
+function getObjectTagging(key) {
+    return s3
+        .getObjectTagging({
+            Bucket: S3_UPLOAD_BUCKET,
+            Key: key
+        })
+        .promise();
+}
+
+function checkAntiVirus({ formId, applicationId, filename }) {
```
davidrapson: You shouldn't need to construct a `new Promise` here or use `async` as you are already returning a promise. Could try something like:

```suggestion
function checkAntiVirus({ formId, applicationId, filename }) {
    const keyName = [formId, applicationId, filename].join('/');
    return s3
        .getObjectTagging({
            Bucket: S3_UPLOAD_BUCKET,
            Key: keyName
        })
        .promise()
        .then(function(tags) {
            const tagSet = get(tags, 'TagSet', []);
            const avStatus = tagSet.find(t => t.Key === 'av-status');
            if (avStatus && get(avStatus, 'Value') === 'INFECTED') {
                logger.error(`Infected file found at ${keyName}`);
                throw new Error('Infected file suspected');
            } else {
                return avStatus;
            }
        });
}
```

controllers/apply/form-router-next/index.js
```diff
@@ -458,39 +473,51 @@ function initFormRouter({
...
+
+                try {
+                    await checkAntiVirus(pathConfig);
+                } catch (err) {
+                    return res.send(
+                        'This file has been blocked for security reasons.'
+                    );
+                }
+
+                getObject(pathConfig)
```
davidrapson: I'd possibly move this block inside the `try`

controllers/apply/form-router-next/lib/file-uploads.js
mattandrews: Good call, done

controllers/apply/form-router-next/index.js
mattandrews: 👍

mattandrews:
Now updated so we show an error/warning after submission:

Screenshot 2019-07-19 12 02 29

Screenshot 2019-07-19 12 03 44

PR #2086, opened 2019-07-19
Password strength check

davidrapson:
Adds a basic strength check using [zxcvbn](https://github.com/dropbox/zxcvbn). Set to require a minimum score of **2** which equates roughly to "somewhat guessable: protection from unthrottled online attacks". 

Screenshot 2019-07-19 at 16 17 09

Screenshot 2019-07-19 at 16 17 25


config/locales/en.yml
```diff
@@ -1085,9 +1085,10 @@ user:
     passwordRequirements: >
       <p>Must be at least 10 characters long <br>
       Must not be the same as your email address<br>
-      A good tip for a secure password is to
-      <a href="https://www.cyberaware.gov.uk/blog/three-random-words">
-      use three random words<span class="u-visually-hidden"> National Cyber Security Centre - External link</span></a>.</p>
+      <a href="https://www.ncsc.gov.uk/collection/top-tips-for-staying-secure-online/use-a-strong-and-separate-password-for-email">
```
davidrapson: We've updated this as the previous blog post we were linking to has been removed/archived. This page exists in it's place. It still mentions "three random words" but covers some other general advice too so we've reworded the link text to "Read tips on how to create a strong password" to better reflect the broader content.

davidrapson:
@scottoakley Tagging you in here as this involves a error message copy change

scottoakley:
Thanks David. I suggest changing the two error messages to:

1. Password is too weak, try another password
2. Password can't be the same as your email address 

For 1. I just removed 'please' so it didn't sound optional. For 2. I thought 'email address' was clearer than username, as their username is always their email address (correct?) and we don't refer to username on the screen as far as I can see. 

Let me know if there's anything I've missed.

PR #2088, opened 2019-07-22
Expire session after period of inactivity

davidrapson:
With `rolling` set the `maxAge` setting is expiry based. Average session duration on the current form is ~12-15 minutes so setting a timeout of 30 minutes of inactivity is reasonable and in line with recommendations.

### DIFF
--- a/controllers/apply/form-router-next/index.js
+++ b/controllers/apply/form-router-next/index.js
@@ -1,5 +1,6 @@
 'use strict';
 const express = require('express');
+const csurf = require('csurf');
 const path = require('path');
 const Sentry = require('@sentry/node');
 const flatMap = require('lodash/flatMap');
@@ -20,7 +21,7 @@ const {
 const commonLogger = require('../../../common/logger');
 const appData = require('../../../common/appData');
 const { localify } = require('../../../common/urls');
-const { csrfProtection } = require('../../../middleware/cached');
+const { noStore } = require('../../../middleware/cached');
 const { requireActiveUser } = require('../../../middleware/authed');
 const { injectCopy } = require('../../../middleware/inject-content');
 
@@ -136,11 +137,12 @@ function initFormRouter({
      * Require active user past this point
      */
     router.use(
+        noStore,
+        requireActiveUser,
         handleMultipartFormData,
-        csrfProtection,
         injectCopy('applyNext'),
         setCommonLocals,
-        requireActiveUser
+        csurf()
     );
 
     /**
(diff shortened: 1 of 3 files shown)


davidrapson:
[PT 9.3: Reduce session expiry time](https://trello.com/c/vRaKEON2/677-pt-93-reduce-session-expiry-time)

mattandrews:
Fair enough – what happens when a user submits a step after 30 minutes have elapsed since pageload? Does the session re-initialise (eg. the `rolling` option) and they keep their data, or does it throw an error?

controllers/apply/form-router-next/index.js
```diff
@@ -136,11 +137,12 @@ function initFormRouter({
...
      */
     router.use(
+        noStore,
+        requireActiveUser,
         handleMultipartFormData,
-        csrfProtection,
         injectCopy('applyNext'),
         setCommonLocals,
-        requireActiveUser
+        csurf()
```
davidrapson: Move `requireActiveUser` handler up in the chain and initialise `csurf` later. This change ensures we don't throw an error if posting to a step after the session has expired.

middleware/cached.js
```diff
@@ -37,13 +37,15 @@ function sMaxAge(sMaxAgeValue) {
...
- * Apply csrf protection and no-cache at the same time
+ * Apply csrf protection and no-store at the same time
  */
-const csrfProtection = [csurf(), noCache];
+const csrfProtection = [csurf(), noStore];
 
 module.exports = {
     defaultMaxAge,
-    noCache,
+    // @TODO: Replace all instances of noCache with noStore
```
davidrapson: `noCache` is a misleading name as it's setting the `no-store` value (`no-cache` is a semantically different thing). Added a TODO here to remind us to renamed the other uses.

davidrapson:
> what happens when a user submits a step after 30 minutes have elapsed since page-load? Does the session re-initialise (eg. the rolling option) and they keep their data, or does it throw an error?

Short version is neither. With the latest changes it will redirect you back to the log in screen if you attempt to submit a step _after_ the session has expired.

davidrapson:
@mattandrews Increased this to 60 minutes of inactivity for now until we're able to implement some clearer user messaging.

davidrapson:
> Can you make a backlog ticket to revisit this though, to detect when a user has submitted a POST request to /apply/*/* and log it (and perhaps show a custom error?)

Done

PR #2090, opened 2019-07-22
Add sanitiser for nested request bodies

davidrapson:
Add sanitiser for nested request bodies. Accounts for body-parse urlencoded extended values where data is expanded into nested objects and arrays.

Attempts to only sanitises if the value is string-like to avoid unwanted stringify-ing of other values.

### DIFF
--- a/controllers/apply/form-router-next/steps.js
+++ b/controllers/apply/form-router-next/steps.js
@@ -5,6 +5,7 @@ const findIndex = require('lodash/findIndex');
 const Sentry = require('@sentry/node');
 
 const logger = require('../../../common/logger');
+const { sanitiseRequestBody } = require('../../../common/sanitise');
 const { PendingApplication } = require('../../../db/models');
 const { prepareFilesForUpload, uploadFile } = require('./lib/file-uploads');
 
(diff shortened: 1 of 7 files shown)


davidrapson:
[PT 9.8: Strip malicious input (rather than sanitising it) (LOW)](https://trello.com/c/hgfuQZSp/682-pt-98-strip-malicious-input-rather-than-sanitising-it-low)

controllers/apply/form-router-next/steps.js
```diff
@@ -109,9 +110,13 @@ module.exports = function(formId, formBuilder) {
                 currentApplicationData
             } = res.locals;
 
+            const sanitisedBody = sanitiseRequestBody(req.body);
+
+            console.log(sanitisedBody);
```
mattandrews: ❌

mattandrews: (though wonder if we should log the number of sanitisations made?)

davidrapson: 🙈 And good call, added debug level logging to the sanitise method:

image

PR #2092, opened 2019-07-22
Remove concept of public routes from apply forms

davidrapson:
Remove concept of public routes from apply forms. At some point we may want to (re)introduce the concept of a public eligibility checker, but for now it simplifies the flow if we put all form routes behind auth.

### DIFF
--- a/controllers/apply/form-router-next/index.js
+++ b/controllers/apply/form-router-next/index.js
@@ -2,12 +2,12 @@
 const express = require('express');
 const path = require('path');
 const Sentry = require('@sentry/node');
-const get = require('lodash/get');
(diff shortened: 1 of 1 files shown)


davidrapson:
[PT 9.6: Move eligibility checker behind login (LOW)](https://trello.com/c/MCbYP29b/680-pt-96-move-eligibility-checker-behind-login-low)

davidrapson:
I'm not sure this PR meaningfully changes this "My only issue with this is that it's going to be difficult for non-technical staff to generate a link to the application area". The main application area was already behind login?

mattandrews:
Yeah sorry you're right, just a note for the future I guess!

PR #2093, opened 2019-07-22
Add rate limiter final

mattandrews:
After a bit of back-and-forth understanding how https://github.com/animir/node-rate-limiter-flexible works, this is what I've come up with.

First we initialise the library and create some tables (in practice these will already exist, but the module throws an error if it hasn't confirmed they exist so we have to do this up front). I've made it so we can add other rate limiters in the future by extending the config, eg.

```
myNewRateLimiter: {
    maxPoints: 100,
    keyPrefix: 'general_rate_limiter,
    duration: 60 * 60 * 24 * 7, // Points will expire after 7 days
    blockDuration: 60 // Exceeding points locks user out for 1 minutes
}
```

... and then use it like this (eg. in a request handler):

```
const MyRateLimiter = await new RateLimiter(
    availableRateLimiters.myNewRateLimiter,
    req.ip // or some other identifying characteristic
).init();

// then we can use it like so:

LoginRateLimiter.isRateLimited() // returns true/false
LoginRateLimiter.hasConsumedPoints() // returns true/false (eg. have they used any attempts)
await LoginRateLimiter.clearRateLimit() // resets their usage (eg. on successful login)
await LoginRateLimiter.consumeRateLimit() // increments their usage count (eg. on failed login)
```

I'm fairly happy with the portability/usage of this now after a bit of trial and error with the construction/async setup. We can tweak the numbers – it's currently 5 failed login attempts (each failure expires after 3 hours), and if they exceed the limit then it's a 15 minute lockout.

There's also a fix in here for the nginx headers which adds the `X-Forwarded-For` which was never being passed properly. I'm uncertain though if this is right, eg. will this break caching? Though I think Cloudfront passes this to us so we shouldn't be varying on it. In any case, when I logged `req.ip()` on the TEST site (behind Cloudfront), it correctly returned my actual IP (even without these changes). The main thing is that the old `proxy_set_header` wasn't doing anything.

Thoughts welcome!

### DIFF
--- a/controllers/user/login.js
+++ b/controllers/user/login.js
@@ -8,7 +8,7 @@ const logger = require('../../common/logger').child({
 });
 
 const {
-    availableRateLimiters,
+    rateLimiterConfigs,
     RateLimiter
 } = require('../../middleware/rate-limiter');
 
@@ -62,11 +62,11 @@ router
         logger.info('Login attempted');
 
         // Key all rate-limited login attempts by the user's IP and the attempted username
-        const uniqueKey = `${req.ip}_${req.body.username}`;
+        const RATE_LIMIT_UNIQUE_KEY = `${req.ip}_${req.body.username}`;
 
         const LoginRateLimiter = await new RateLimiter(
-            availableRateLimiters.failByUsername,
-            uniqueKey
+            rateLimiterConfigs.failByUsername,
+            RATE_LIMIT_UNIQUE_KEY
         ).init();
 
         if (LoginRateLimiter.isRateLimited()) {
@@ -80,7 +80,10 @@ router
                 } else if (user) {
                     req.logIn(user, async function(loginErr) {
                         if (loginErr) {
-                            logger.error('Login failed', loginErr);
+                            logger.error(
+                                'Login failed: Error with authentication',
+                                loginErr
+                            );
                             next(loginErr);
                         } else {
                             logger.info('Login succeeded');
@@ -96,7 +99,7 @@ router
                      * User is invalid
                      * Show a generic error message here to avoid exposing account state
                      */
-                    logger.warn('Login failed: invalid credentials');
+                    logger.warn('Login failed: Unsuccessful login attempt');
                     try {
                         await LoginRateLimiter.consumeRateLimit();
                         return renderForm(req, res, req.body, [
@@ -104,9 +107,9 @@ router
                                 msg: `Your username and password combination is invalid`
                             }
                         ]);
-                    } catch (rlRejected) {
-                        if (rlRejected instanceof Error) {
-                            next(rlRejected);
+                    } catch (rateLimitRejection) {
+                        if (rateLimitRejection instanceof Error) {
+                            next(rateLimitRejection);
                         } else {
                             // User is rate limited
                             return renderRateLimitError(req, res);
--- a/controllers/user/login.js
+++ b/controllers/user/login.js
@@ -35,14 +35,11 @@ function renderForm(req, res, formValues = null, errors = []) {
 }
 
 function renderRateLimitError(req, res, minutesLeft) {
-    const errorData = {
-        minutesTillNextAllowedAttempt: minutesLeft
-    };
     res.status(429).render(
         path.resolve(__dirname, './views/error-rate-limit'),
         {
             title: 'Too many requests',
-            errorData: errorData
+            minutesLeft: minutesLeft
         }
     );
 }
--- a/controllers/user/login.js
+++ b/controllers/user/login.js
@@ -7,7 +7,10 @@ const logger = require('../../common/logger').child({
     service: 'user'
 });
 
-const { RateLimiter } = require('../../middleware/rate-limiter');
+const {
+    availableRateLimiters,
+    RateLimiter
+} = require('../../middleware/rate-limiter');
 
 const {
     injectCopy,
@@ -31,6 +34,15 @@ function renderForm(req, res, formValues = null, errors = []) {
     });
 }
 
+function renderRateLimitError(req, res) {
+    res.status(429).render(
+        path.resolve(__dirname, './views/error-rate-limit'),
+        {
+            title: 'Too many requests'
+        }
+    );
+}
+
 router
     .route('/')
     .all(
@@ -49,16 +61,17 @@ router
     .post(async (req, res, next) => {
         logger.info('Login attempted');
 
-        const LoginRateLimiter = new RateLimiter(
-            'failByUsername',
-            req.body.username
-        );
+        // Key all rate-limited login attempts by the user's IP and the attempted username
+        const uniqueKey = `${req.ip}_${req.body.username}`;
 
-        const rateLimiter = await LoginRateLimiter.getLimiter();
+        const LoginRateLimiter = await new RateLimiter(
+            availableRateLimiters.failByUsername,
+            uniqueKey
+        ).init();
 
-        if (rateLimiter.isRateLimited) {
+        if (LoginRateLimiter.isRateLimited()) {
             // User is rate limited
-            res.status(429).send('Too Many Requests');
+            return renderRateLimitError(req, res);
         } else {
             passport.authenticate('local', async function(err, user) {
                 if (err) {
@@ -71,12 +84,10 @@ router
                             next(loginErr);
                         } else {
                             logger.info('Login succeeded');
-
                             // Reset rate limit on successful authorisation
-                            if (rateLimiter.hasConsumedPoints) {
+                            if (LoginRateLimiter.hasConsumedPoints()) {
                                 await LoginRateLimiter.clearRateLimit();
                             }
-
                             redirectUrlWithFallback(req, res, '/user');
                         }
                     });
@@ -98,9 +109,7 @@ router
                             next(rlRejected);
                         } else {
                             // User is rate limited
-                            res.status(429).send(
-                                'Too Many Requests (login fail)'
-                            );
+                            return renderRateLimitError(req, res);
                         }
                     }
                 }
--- a/controllers/user/login.js
+++ b/controllers/user/login.js
@@ -7,6 +7,8 @@ const logger = require('../../common/logger').child({
     service: 'user'
 });
 
+const { RateLimiter } = require('../../middleware/rate-limiter');
+
 const {
     injectCopy,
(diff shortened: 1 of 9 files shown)


controllers/user/login.js
```diff
@@ -29,6 +34,15 @@ function renderForm(req, res, formValues = null, errors = []) {
     });
 }
 
+function renderRateLimitError(req, res) {
```
davidrapson: Could we add a log when we render this: `logger.info('User rate limited')`

controllers/user/login.js
```diff
@@ -44,33 +58,63 @@ router
...
-                req.logIn(user, function(loginErr) {
-                    if (loginErr) {
-                        logger.error('Login failed', loginErr);
-                        next(loginErr);
-                    } else {
-                        logger.info('Login succeeded');
-                        redirectUrlWithFallback(req, res, '/user');
+
+        // Key all rate-limited login attempts by the user's IP and the attempted username
+        const uniqueKey = `${req.ip}_${req.body.username}`;
```
davidrapson: As this is a fairly important key could we name this something like `RATE_LIMIT_UNIQUE_KEY`

controllers/user/login.js
```diff
@@ -44,33 +58,63 @@ router
...
+                            }
+                            redirectUrlWithFallback(req, res, '/user');
+                        }
+                    });
+                } else {
+                    /**
+                     * User is invalid
+                     * Show a generic error message here to avoid exposing account state
+                     */
+                    logger.warn('Login failed: invalid credentials');
```
davidrapson: Just realised we have `Login failed` as the log value both for this case and also when the login fails due to an error. Could we renamed this to something like `Unsuccessful login attempt` to help disambiguate these two.

controllers/user/login.js
```diff
@@ -44,33 +58,63 @@ router
...
+                    try {
+                        await LoginRateLimiter.consumeRateLimit();
+                        return renderForm(req, res, req.body, [
+                            {
+                                msg: `Your username and password combination is invalid`
+                            }
+                        ]);
+                    } catch (rlRejected) {
```
davidrapson: Mind renaming to something like `rateLimitRejection`, took me a bit to grok what `rlRejected ` was. Also is there a clearer way of determining the type of error than checking `instanceof`?

middleware/rate-limiter.js
```diff
@@ -0,0 +1,109 @@
...
+} = require('rate-limiter-flexible');
+const { startsWith } = require('lodash');
+
+const { sequelize } = require('../db/models/index');
+const appData = require('../common/appData');
+const logger = require('../common/logger').child({
+    service: 'rate-limiter'
+});
+
+const sqlDialect = startsWith(process.env.DB_CONNECTION_URI, 'sqlite://')
```
davidrapson: Rather than doing this check again could you try to read the value out of `databaseConfig` (see db/index.js), perhaps by exporting it, or you might be able to ask the `sequelize` instance for this.

middleware/rate-limiter.js
```diff
@@ -0,0 +1,109 @@
...
+        blockDuration: 60 * 15 // Exceeding points locks user out for 15 minutes
+    }
+};
+
+// We have to create all the ratelimiter db instances here
+// so the tables exist and are ready when the middleware is used.
+// We also need to be able to access the maxPoints
+// within the class below (which this API doesn't expose)
+// so we set it here as a property against the ratelimiter object
+let availableRateLimiters = {};
```
davidrapson: Couple of options to simplify this, you could either use lodash `mapValues` to map over this like

```js
const availableRateLimiters = mapValues(rateLimiterConfigs, (conf, name) => {
    return {
        id: name,
        config: conf,
        instance:
            sqlDialect === 'mysql'
                ? new RateLimiterMySQL({
                      storeClient: sequelize,
                      dbName: `rate_limiter_${appData.environment}`,
                      tableName: 'rate_limiter',
                      keyPrefix: conf.keyPrefix,
                      points: conf.maxPoints,
                      duration: conf.duration,
                      blockDuration: conf.blockDuration
                  })
                : new RateLimiterMemory({
                      keyPrefix: conf.keyPrefix,
                      points: conf.maxPoints,
                      duration: conf.duration,
                      blockDuration: conf.blockDuration
                  })
    };
});
```

Or, perhaps make this a function instead:

```js
function rateLimiterFor(conf) {
    if (sqlDialect === 'mysql') {
        return new RateLimiterMySQL({
            storeClient: sequelize,
            dbName: `rate_limiter_${appData.environment}`,
            tableName: 'rate_limiter',
            keyPrefix: conf.keyPrefix,
            points: conf.maxPoints,
            duration: conf.duration,
            blockDuration: conf.blockDuration
        });
    } else {
        new RateLimiterMemory({
            keyPrefix: conf.keyPrefix,
            points: conf.maxPoints,
            duration: conf.duration,
            blockDuration: conf.blockDuration
        });
    }
}

function makeRateLimiter(id, conf) {
    return {
        id: id,
        config: conf,
        instance: rateLimiterFor(conf)
    };
}

const rateLimiterConfigs = {
    failByUsername: makeRateLimiter('failByUsername', {
        maxPoints: 5,
        keyPrefix: 'login_fail_consecutive_username',
        duration: 60 * 60 * 3, // Points will expire after 3 hours
        blockDuration: 60 * 15 // Exceeding points locks user out for 15 minutes
    })
};
```

or some combination of the two.

middleware/rate-limiter.js
```diff
@@ -0,0 +1,109 @@
...
+            this.limiterInstance = await this.limiter.get(this.keyValue);
+        } catch (e) {} // eslint-disable-line no-empty
+        return this;
+    }
+
+    isRateLimited() {
+        const isRateLimited =
+            this.limiterInstance !== null &&
+            this.limiterInstance.consumedPoints >
+                this.limiterConf.config.maxPoints;
```
davidrapson: This can be `this.maxPoints` as you're setting it above in the constructor

middleware/rate-limiter.js
```diff
@@ -0,0 +1,109 @@
...
+    service: 'rate-limiter'
+});
+
+const sqlDialect = startsWith(process.env.DB_CONNECTION_URI, 'sqlite://')
+    ? 'sqlite'
+    : 'mysql';
+
+const rateLimiterConfigs = {
+    failByUsername: {
+        maxPoints: 5,
```
davidrapson: Is this the number of attempts the user has? If so can we set this to 10 which is the minimum value recommended in https://design-system.service.gov.uk/patterns/passwords/

controllers/user/login.js
mattandrews: Doing this in the main ratelimiter module already: https://github.com/biglotteryfund/blf-alpha/pull/2093/files#diff-f3fc288629418f5e498c36b048899b97R80

davidrapson:
> There's also a fix in here for the nginx headers which adds the X-Forwarded-For which was never being passed properly. I'm uncertain though if this is right, eg. will this break caching? Though I think Cloudfront passes this to us so we shouldn't be varying on it. In any case, when I logged req.ip() on the TEST site (behind Cloudfront), it correctly returned my actual IP (even without these changes). The main thing is that the old proxy_set_header wasn't doing anything.

We won't be varying on this (AFAIK Cloudfront won't let you) and the place where we are using this is not cached anyway so this is fine.

The behaviour before the change you made makes sense based on the docs: "If a viewer sends a request to CloudFront and does not include an X-Forwarded-For request header, CloudFront gets the IP address of the viewer from the TCP connection, adds an X-Forwarded-For header that includes the IP address, and forwards the request to the origin"

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#RequestCustomIPAddresses

controllers/user/login.js
mattandrews: This was copied straight from the docs for this module so I guess it's their way of determining that this is specifically an error with the module (eg. I triggered this by deleting one of its table rows midway through its validation process) rather than anything else. Will rename though!

mattandrews: Ahh, looking at it again, they have this example:

```
} catch (rlRejected) {
        if (rlRejected instanceof Error) {
          throw rlRejected;
        } else {
          res.set('Retry-After', String(Math.round(rlRejected.msBeforeNext / 1000)) || 1);
          res.status(429).send('Too Many Requests');
        }
      }
```

So sometimes this is an actual error (eg. something broke unexpectedly), other times it contains details of the user's current rate limit.

controllers/user/login.js
davidrapson: Ah nice  👍

middleware/rate-limiter.js
mattandrews: `sequelize.getDialect()` FTW

controllers/user/login.js
```diff
@@ -29,6 +34,19 @@ function renderForm(req, res, formValues = null, errors = []) {
     });
 }
 
+function renderRateLimitError(req, res, minutesLeft) {
+    const errorData = {
```
davidrapson: Why the extra `errorData` object instead of passing `minutesLeft` straight into the view?

cypress/integration/common.js
```diff
@@ -313,6 +314,32 @@ describe('user', () => {
...
+            cy.loginUser({
+                username: fakeEmail,
+                password: fakePassword
+            }).then(response => {
+                expect(response.status).to.eq(200);
+            });
+
+            // If we're at the end, run a final one to ensure we're blocked
+            if (i === maxAttempts - 1) {
+                cy.loginUser({
```
davidrapson: Could this be called after the `times` block rather than needing to check the index?

middleware/rate-limiter.js
```diff
@@ -0,0 +1,128 @@
...
+            });
+        }
+        return isRateLimited;
+    }
+
+    minutesTillNextAllowedAttempt() {
+        if (!this.limiterInstance) {
+            return null;
+        }
+        const retrySecs =
```
davidrapson: Should this be `Math.ceil`, i.e. always round _up_ to avoid giving the answer that's slightly low? We're talking about a difference of a second here so might not be a big deal.

mattandrews: I copied this from the docs for the library...

PR #2096, opened 2019-07-23
Add project name header

mattandrews:
Re-adds the project header, eg. shows the funding programme name and the application title.

image
image

### DIFF
--- a/assets/sass/scaffolding/_content.scss
+++ b/assets/sass/scaffolding/_content.scss
@@ -111,14 +111,26 @@
    ========================================================================= */
 
 .content-box {
+    $spacingDefault: $spacingUnit / 2;
+
     background-color: white;
     border: 1px solid get-color('border', 'base');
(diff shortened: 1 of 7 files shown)


assets/sass/scaffolding/_content.scss
```diff
@@ -111,16 +111,28 @@
...
+    padding: $spacingUnit $spacingDefault;
 
     @include mq('medium') {
         padding: $spacingUnit;
     }
 
+    &.content-box__inner {
+        border: none;
+        margin-left: -$spacingDefault;
+        width: calc(100% + #{$spacingDefault * 2});
```
mattandrews: These two lines look a bit magic – it's because they're counteracting the padding the box itself sets, which would be a pain to remove as everything else inside it needs the padding. We use a similar pattern (the `calc` usage) in a few other places.

PR #2098, opened 2019-07-23
Allow only one activation link to be current at a time

mattandrews:
Changes the user model to store a unix timestamp of the date of the most recent activation request. This number is stored in the JSON Web Token too. When a user activates their account, we check that the number in the JWT matches the one in the database. If it doesn't, we reject the activation (eg. because it's an older link).

This will require dropping/recreating the `users` table on production but I think we can get around this (or just manually adding the column) quite easily as nobody's using it yet.

### DIFF
--- a/controllers/user/update-email.js
+++ b/controllers/user/update-email.js
@@ -77,7 +77,7 @@ router
                         newEmail: username
                     });
                     const updatedUser = await Users.findById(userId);
-                    await sendActivationEmail(req, updatedUser, true);
+                    await sendActivationEmail(req, updatedUser);
                     logger.info('Update email change successful');
                     redirectForLocale(req, res, '/user?s=emailUpdated');
                 }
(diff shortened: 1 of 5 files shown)


mattandrews:
[PT 9.12: Prevent old activation links from working when superseded by new ones (LOW)](https://trello.com/c/b0wuK7PP/686-pt-912-prevent-old-activation-links-from-working-when-superseded-by-new-ones-low)

controllers/user/update-email.js
```diff
@@ -77,7 +77,7 @@ router
                         newEmail: username
                     });
                     const updatedUser = await Users.findById(userId);
-                    await sendActivationEmail(req, updatedUser, true);
+                    await sendActivationEmail(req, updatedUser);
```
davidrapson: Do we need to re-add the date of activation here too? What happens when this is `null`?

mattandrews: Ahh good catch, forgot we sent it there too. Adding it now.

PR #2099, opened 2019-07-23
Email password confirmation

davidrapson:
Adds a password confirmation check to the update email screen

image

### DIFF
--- a/controllers/user/password.js
+++ b/controllers/user/password.js
@@ -196,7 +196,7 @@ router
                         const username = req.user.userData.username;
                         const { password } = validationResult.value;
                         await Users.updateNewPassword({
-                            id: sanitise(username),
+                            id: req.user.userData.id,
                             newPassword: password
                         });
                         await sendPasswordResetNotification(req, username);
--- a/controllers/user/update-email.js
+++ b/controllers/user/update-email.js
@@ -25,62 +25,62 @@ function render(req, res, data = null, errors = []) {
     });
 }
 
+async function handleSubmission(req, res, next) {
+    const validationResult = validateSchema(
+        emailOnly(req.i18n.getLocale()),
+        req.body
+    );
+
+    if (validationResult.isValid) {
+        try {
+            const { username } = validationResult.value;
+            const existingUser = await Users.findByUsername(username);
+            const passwordMatches = await Users.checkValidPassword(
+                req.user.userData.password,
+                req.body.password
+            );
+
+            /**
+             * Allow updating if the password confirmation matches
+             * and there is not an existing user with the requested email address
+             */
+            const canUpdate =
+                passwordMatches === true &&
+                username !== get('username')(existingUser);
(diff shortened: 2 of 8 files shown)


controllers/user/password.js
```diff
@@ -196,7 +196,7 @@ router
                         const username = req.user.userData.username;
                         const { password } = validationResult.value;
                         await Users.updateNewPassword({
-                            id: sanitise(username),
+                            id: req.user.userData.id,
```
davidrapson: This fixes an important bug where for logged in users it wasn't actually updating the password.

db/models/user.js
```diff
@@ -58,13 +58,12 @@ class User extends Model {
 
     static updateNewPassword({ id, newPassword }) {
         return this.encryptPassword(newPassword).then(newEncryptedPassword => {
-            return this.update(
-                {
+            return this.findById(id).then(user => {
```
davidrapson: This change more directly queries for the specific user and then updates which returns the revised user record (the previous approach returned an obtuse `[1]` value if successful.

cypress/integration/common.js
```diff
@@ -110,6 +110,202 @@ describe('server smoke tests', function() {
...
+                /^[a-zA-Z0-9\-_]+?\.[a-zA-Z0-9\-_]+?\.([a-zA-Z0-9\-_]+)?$/
+            );
+            expect(res.body.mailParams.sendTo).to.equal(username);
+            expect(res.body.mailParams.subject).to.equal(
+                'Activate your The National Lottery Community Fund website account'
+            );
+        });
+    });
+
+    it('should be able to log in and update account details', () => {
```
davidrapson: This is the most notable addition here, making sure there are e2e tests for updating account details when logged in. The rest of the changes are streamlining the user tests.

controllers/user/update-email.js
```diff
@@ -80,25 +62,24 @@ router
                     await sendActivationEmail(req, updatedUser, true);
```
mattandrews: This will conflict with my branch but you just need to remove `true` – we previously showed a different message on the confirmation screen for new versus existing users, but it's the same now.

cypress/integration/common.js
```diff
@@ -110,6 +110,202 @@ describe('server smoke tests', function() {
...
+            expect(response.body).to.include('Too many requests');
+        });
+    });
+
+    it('should prevent registrations with invalid passwords', () => {
+        const username = `${Date.now()}@example.com`;
+
+        cy.visit('/user/register');
+
+        createAccount(username, '5555555555');
```
mattandrews: hey how did you hack my password

davidrapson:
[PT 9.7: Require password confirmation to modify email address (LOW)](https://trello.com/c/Ix0TyGLO/681-pt-97-require-password-confirmation-to-modify-email-address-low)

PR #2100, opened 2019-07-23
Prevent a budget from being submitted with over-maximum rows

mattandrews:
We didn't check server-side for maximum budget items (weirdly, as I'm sure this was there at one stage) – we also check whether we should add a new blank row when initialising the component in case there were already 10 rows.

### DIFF
--- a/controllers/apply/form-router-next/joi-extensions/budget-items.js
+++ b/controllers/apply/form-router-next/joi-extensions/budget-items.js
@@ -38,24 +38,6 @@ module.exports = function budgetItems(joi) {
         },
         rules: [
             {
-                name: 'maxRows',
-                params: {
-                    rowLimit: joi.number().required()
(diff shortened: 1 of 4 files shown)


controllers/apply/form-router-next/joi-extensions/budget-items.js
```diff
@@ -37,6 +37,24 @@ module.exports = function budgetItems(joi) {
             }
         },
         rules: [
+            {
```
davidrapson: You might not need this custom validator, as this extends array I think you can just call `.max` via `Joi.budgetItems().max(MAX_LIMIT)`

mattandrews: Ahh yeah you're right, nice spot. Will change!

PR #2102, opened 2019-07-23
Sequelize upgrade

davidrapson:
We're a major version behind on Sequelize and `npm audit` identified a potential vulnerability so making sure we are up to date. There was one notable API change in the form of switching `findById` to `findByPk` and the `operatorsAliases` is now disabled by default so we can remove that setting (logs a warning when it's left in).

I've upgraded any session related dependencies during this, notably `connect-session-sequelize` required an update to the latest version to fix their internal `findByPk` switch.


mattandrews:
Nice – I added a `findByUserId` function to `user.js` (line 53) without realising there was a built-in one.  Could you swap its usage (`jwt.js:27`) for this new method instead, since you're in here? 😇

PR #2103, opened 2019-07-23
Simplify the address lookup 

mattandrews:
This change removes the address preview step, eg. this screen:

image

The advantage of this is that it removes a bunch of confusing state edge cases, particularly around required fields. The new behaviour is:

- user looks up a postcode
- if found, the dropdown of addresses is populated
- when one is selected (and the field blurred), the form inputs appear, pre-populated
- the user can edit them manually or remove the whole lot and start again
- if the lookup fails, the form inputs appear for manual provision

Here's a demo:

postcode-new

### DIFF
--- a/assets/js/vue-apps/form-components/address-lookup.vue
+++ b/assets/js/vue-apps/form-components/address-lookup.vue
@@ -1,111 +1,107 @@
 <script>
-import $ from 'jquery';
-import compact from 'lodash/compact';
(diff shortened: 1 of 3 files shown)


davidrapson:
Code looks good so LGTM once we're happy from a user experience point of view. You'll likely need to come back and updates the integration tests too as they'll expect the previous behaviour.

PR #2106, opened 2019-07-24
Tag a bunch of situations for HotJar to aid filtering

mattandrews:
Best viewed alongside [this doc](https://docs.google.com/document/d/1mLFj-fJAWMQYBXyMI6z1sUQqspNJPh3Su161paGZNr4/edit) which outlines the intentions here.

Adds some tags to various user-initiated events (and a few other exceptions) so we can filter those recordings in Hotjar to help diagnose confusion or edge cases.

### DIFF
--- a/assets/js/common/forms.js
+++ b/assets/js/common/forms.js
@@ -1,6 +1,6 @@
 import $ from 'jquery';
 import forEach from 'lodash/forEach';
-import { trackEvent } from '../helpers/metrics';
+import { trackEvent, tagHotjarRecording } from '../helpers/metrics';
 
 function handleBeforeUnload(e) {
     // Message cannot be customised in Chrome 51+
@@ -10,6 +10,7 @@ function handleBeforeUnload(e) {
         'User warned before abandoning form changes',
         'message shown'
     );
+    tagHotjarRecording(['App: User shown page abandonment warning']);
     const confirmationMessage = 'Are you sure you want to leave this page?';
     e.returnValue = confirmationMessage; // Gecko, Trident, Chrome 34-51
     return confirmationMessage; // Gecko, WebKit, Chrome <34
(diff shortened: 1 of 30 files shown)


assets/js/common/forms.js
```diff
@@ -145,6 +146,71 @@ function handleConditionalRadios() {
     });
 }
 
+// Track occurrences of users changing their mind about radio options
+// eg. to highlight potential confusion around questions
+function trackIndecisiveness() {
+    let fields = {};
+    $('input[type="radio"]').on('click', function() {
```
davidrapson: General comment on all of these listeners: It would be good to restrict these selectors to the main form (i.e. be certain we're not listening to other forms).

assets/js/helpers/metrics.js
```diff
@@ -20,3 +20,13 @@ export function trackEvent(category, action, label) {
         ga('send', payload);
     }
 }
+
+/**
+ * Tag recording
+ * Helper to add tag(s) to the current session's Hotjar recording
+ */
+export function tagHotjarRecording(tags) {
+    if (window.hj && tags) {
```
davidrapson: Not sure we need the `&& tags` check, we're always passing this in as the function argument.

common/alert-message.js
```diff
@@ -18,6 +18,12 @@ module.exports = function alertMessage({ locale, status }) {
                 cy: ``
             });
             break;
+        case 'applicationDeleted':
```
davidrapson: I don't thinks this file should be moved into common.  This block doesn't need to be in here, should probably assign this value directly in `dashboard.js` and keep this file as a helper for the user controller. If we have more message that need to be shared for the form-router code add a similar `alert-message.js` helper in there. Keen to keep `common/` for code that is genuinely site-wide.

assets/js/helpers/metrics.js
mattandrews: True, I guess it's just a sanity check, but since it's only one parameter I guess it's safe enough.

assets/js/common/forms.js
mattandrews: Done!

common/alert-message.js
mattandrews: Fair enough, have moved it back.

PR #2109, opened 2019-07-25
Optional registration numbers

davidrapson:
Fixes a but from https://github.com/biglotteryfund/blf-alpha/pull/2108 where optional charity numbers were getting stripped. Needed an additional condition to account for the cases where charity number is optional.

Realised as part of this we need to share the config values for the different registration numbers to allow us to keep the fields, form, and tests in sync. Previously the tests were incomplete and repeating the logic, the are now generated based on shared config.


controllers/apply/awards-for-all/form.test.js
```diff
@@ -587,122 +587,156 @@ describe('Form validations', () => {
...
+                ).toEqual(data);
+            }
+        );
 
-        test('registration numbers shown based on organisation type', () => {
-            const defaultFieldNames = formBuilder({ locale: 'en' })
-                .getCurrentFieldsForStep('organisation', 2)
-                .map(field => field.name);
+        test.each(EDUCATION_NUMBER_TYPES)(
+            'eduction number required for %p',
```
mattandrews: Typo here, but only minor ("eduction")

PR #2110, opened 2019-07-25
Replace snyk with npm audit

davidrapson:
We've been trialing Snyk again for a few weeks and have found that the `snyk protect` step adds a 2-3 **minute** delay to builds. Fast build speed can enable enough good things that I think it's worth looking (again) at `npm audit`.

We've been here before: https://github.com/biglotteryfund/blf-alpha/pull/1197

Previously `npm audit` had no `--production` flag so would check `devDependencies` which causes a lot of churn when run as a blocking step in CI.

This changed in `npm` 6.10.x and we can now run `npm audit --production` to only fail the build if there is a vulnerability in one of our _production_ dependencies.

We still get email alerts from Snyk (and also from Github, there's a lot of competition in the node security space) so I think we are well covered here.

_Note_: This will probably fail until https://github.com/biglotteryfund/blf-alpha/pull/2102 is merged but that's as expected.


mattandrews:
image

PR #2117, opened 2019-07-25
AFA tools tweaks

mattandrews:
This does three things:

- Adds a Google Data Studio embed if found (via app secrets)
- Adds a link to the AFA page on the tools list
- Auto logs out a normal user if they request a staff user endpoint (and vice versa)

### DIFF
--- a/common/secrets.js
+++ b/common/secrets.js
@@ -102,10 +102,14 @@ const POSTCODES_API_KEY =
 
 const S3_KMS_KEY_ID = process.env.S3_KMS_KEY_ID || getSecret('s3.kms.key.id');
 
+const DATA_STUDIO_AFA_URL =
+    process.env.DATA_STUDIO_AFA_URL || getSecret('dataStudio.url.awardsForAll');
+
 module.exports = {
     AWARDS_FOR_ALL_SECRET,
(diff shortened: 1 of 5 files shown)


common/secrets.js
```diff
@@ -102,10 +102,14 @@ const POSTCODES_API_KEY =
 
 const S3_KMS_KEY_ID = process.env.S3_KMS_KEY_ID || getSecret('s3.kms.key.id');
 
+const DATA_STUDIO_AFA_URL =
+    process.env.DATA_STUDIO_AFA_URL || getSecret('dataStudio.url.awardsForAll');
```
davidrapson: We probably don't need the `process.env.DATA_STUDIO_AFA_URL` part as it's unlikely we'd need to override this in the environment or CI.

middleware/authed.js
```diff
@@ -77,7 +77,11 @@ function requireStaffAuth(req, res, next) {
         if (isStaff(req.user)) {
             next();
         } else {
-            res.redirect('/user');
+            // Log out regular users and send them to the staff page
+            req.logout();
```
davidrapson: 😍

PR #2119, opened 2019-07-26
Add more logs for AFA

mattandrews:
Couple of new logs to aid dashboard tracking.


controllers/apply/form-router-next/eligibility.js
```diff
@@ -50,8 +50,9 @@ module.exports = function(eligibilityBuilder, formId) {
             const { currentStepNumber, totalSteps } = res.locals;
 
             logger.info('Eligibility check', {
-                eligible: req.body.eligibility === 'yes',
-                formId: formId
+                eligible: req.body.eligibility,
```
mattandrews: This is so we can facet against a string as this doesn't seem to work for booleans.

PR #2122, opened 2019-07-29
i18n prep

mattandrews:
Continues with the process of moving hard-coded language strings into locale files in prep for translation.

This moves as much copy as I could find that wasn't already using the `localise({ en: '', cy: '' })` pattern into the locale files.


assets/js/vue-apps/form-components/budget-input.test.js
```diff
@@ -62,14 +62,12 @@ describe('BudgetInput', () => {
         addItem('My thing 9', 6000);
         checkItemCount(10);
         checkTotal('£10,500');
-        checkWarning(
-            `Costs you would like us to fund must be less than £${maxBudget.toLocaleString()}.`
-        );
+        checkWarning(`£${maxBudget.toLocaleString()}.`);
```
mattandrews: This fix (and below) is really dumb but I've spent almost half a day on it and fed up wasting time now. You can see the [original error here](https://travis-ci.org/biglotteryfund/blf-alpha/builds/564944792#L498) – no matter how I tried to fix this (passing in the `i18n` copy object from memory, passing it in explicitly with the values hard coded) it broke the component. Random things like the total would fail to compute. I couldn't work out what was causing this given the lengths I've gone to here to make the `copy` optional so I've resorted to simply checking that the error contains the number in question.

mattandrews: Going forward we could do with a better pattern both for adding translations to client-side components and for testing them. I didn't want to use the `localise({ en: '...', cy: '...'})` pattern here because these translations are shared between the server/client.

davidrapson: Yeah I can see this being a pain, personally for any vue components I like the pattern of a single property per string value (so translations happen outside the component) where we are more able to handle the variation. Although trying to use the localise pattern would be OK too, I think it's possibly overkill to try and share translations between server and client, so I'd be OK duplicating them.

controllers/apply/form-router-next/lib/form-model.js
```diff
@@ -250,7 +250,7 @@ class FormModel {
         return this.sections[currentSectionIndex + 1];
     }
 
-    previousPage({ baseUrl, sectionSlug, currentStepIndex = null }) {
+    previousPage({ baseUrl, sectionSlug, currentStepIndex = null, copy = {} }) {
```
davidrapson: For the form model I'd be inclined to pass the locale to the form model as a whole, or when constructing the form model rather than as a function argument. But that might be overkill for now so I reckon we go with this and revisit.

controllers/user/lib/account-schemas.js
```diff
@@ -20,65 +19,33 @@ const passwordConfirmationSchema = Joi.string()
...
-            cy: ''
-        });
-    },
-    passwordConfirmation(locale) {
-        return get(locale)({
-            en: 'Passwords must match'
-        });
-    }
-};
+function getTranslations(i18n) {
```
davidrapson: 💯

views/components/form-fields-next/macros.njk
```diff
@@ -351,7 +349,7 @@
         <div class="o-details__content ff-address__fields">
             <div class="ff-address__field">
                 {{ inputText({
-                    label: 'Building and street',
+                    label: copy.fields.address.fieldNames.line1,
```
davidrapson: It will be interesting to see how this feels when we build subsequent forms, i.e. it might be better to consider these bits of copy "global" with explicit `__('commonFields.address...')` style calls rather than implicit `copy` object. This is good for now though.

assets/js/vue-apps/form-components/address-line.vue
```diff
@@ -13,8 +13,7 @@ export default {
     <div class="ff-address__field">
         <label class="ff-label" :for="'field-' + name">
             <span v-html="label"></span>
-            <!-- @TODO i18n -->
-            <span v-if="!required" class="ff-label-note">(Optional)</span>
+            <span v-if="!required" class="ff-label-note">({{ copyRequired }})</span>
```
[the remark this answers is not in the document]
mattandrews: image
☺️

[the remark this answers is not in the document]
mattandrews: (fixed now)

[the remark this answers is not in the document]
davidrapson: That should be the other way around right? That field is optional? I meant the copy was correct but the variable name was wrong.

PR #2123, opened 2019-07-29
Prevent three fields from showing errors when they shouldn't be validated

mattandrews:
Starting a new application and viewing the summary gives these logging errors:

```
[error] projectLocation not included in step fields but failed validation
[error] projectLocationDescription not included in step fields but failed validation
[error] projectPostcode not included in step fields but failed validation
```

This change makes these fields conditional based on `projectCountry` (which `form.js` already does), matching the work we did last week and removing this noise from logs:

image


davidrapson:
Looks like we might have coverage for this via the current unit tests, so 👍

PR #2135, opened 2019-07-31
Remove custom metrics in favour of regular logs

davidrapson:
Removes the remaining single use of custom metrics in favour of a regular log.


common/mail.js
```diff
@@ -223,19 +221,7 @@ function sendEmail({ name, mailConfig, mailTransport = null }) {
         const transport = mailTransport ? mailTransport : createSesTransport();
 
         return transport.sendMail(buildMailOptions(mailConfig)).then(info => {
-            /**
-             * Record a send count as a CloudWatch event if enabled
-             * istanbul ignore if
```
mattandrews: lol at the also-redundant reference to `istanbul`.

davidrapson: It's not Constantinople

davidrapson:
@mattandrews Build was failing because I'd not deleted enough tests…

PR #2136, opened 2019-07-31
Add a marker for unspecified country and use a bar chart instead

mattandrews:
image

### DIFF
--- a/controllers/tools/views/applications.njk
+++ b/controllers/tools/views/applications.njk
@@ -145,7 +145,7 @@
                     </div>
                     <script>
                         new Chart(document.getElementById('{{ chartId }}'), {
-                            type: 'line',
+                            type: 'bar',
                             data: {
                                 {% if country %}
                                     datasets: [{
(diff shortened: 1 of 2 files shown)


controllers/tools/views/applications.njk
```diff
@@ -161,7 +161,8 @@
                                                 label: '{{ country.title }}',
                                                 data: {{ country.data | dump(2) | safe }},
                                                 borderColor: '{{ country.colour }}',
-                                                fill: '{{ country.colour }}'
+                                                fill: '{{ country.colour }}',
```
davidrapson: Do we still need `fill`?

davidrapson:
Side question: Should we make the submitted applications chart a bar as well?

mattandrews:
It uses the same code so it gets these changes too!

controllers/tools/views/applications.njk
mattandrews: Possibly not but I left it in in case we change back and this way it'll still work with the line format.

PR #2138, opened 2019-08-01
Form wording tweaks

mattandrews:
A few tweaks to enhance usability and fix a bug.

### DIFF
--- a/assets/js/vue-apps/form-components/address-lookup.vue
+++ b/assets/js/vue-apps/form-components/address-lookup.vue
@@ -1,254 +1,255 @@
 <script>
-    import $ from 'jquery';
-    import compact from 'lodash/compact';
+import $ from 'jquery';
+import compact from 'lodash/compact';
 
-    import AddressLine from './address-line.vue';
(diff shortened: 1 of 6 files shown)


assets/js/vue-apps/form-components/address-lookup.vue
```diff
@@ -1,254 +1,255 @@
...
+
+        const $form = $(this.$el)
+            .parents('form')
+            .find('input[type="submit"]');
+        const that = this;
+        $form.on('click', function() {
+            if (
+                that.candidates.length === 0 &&
+                that.postcode &&
+                that.currentState !== states.NotRequired
```
mattandrews: This line is the only real change here (sorry, prettier/whitespace otherwise).

It fixes the scenario where someone clicks "No" here, enters a postcode, then clicks "Yes" (eg. hides the widget) and then submits the step – they were seeing the `alert()` even though the field wasn't required anymore (although their browser still took them to the next step).

image

mattandrews:
Yep that's right!

PR #2141, opened 2019-08-02
Switch to express built-in body-parser

davidrapson:
This will need a bit of stress testing but the recent versions of `express` include `express.json` and `express.urlencoded` methods which alias `body-parser` allowing us to ditch the dependency and defer to the built-in version.


davidrapson:
Kicked the tyres on this an am fairly happy that this is a one-to-one replacement so safe to go.

PR #2142, opened 2019-08-02
Add applicant feedback to dashboard

mattandrews:
A good suggestion from @davidrapson:

image

### DIFF
--- a/db/models/feedback.js
+++ b/db/models/feedback.js
@@ -39,10 +39,27 @@ class Feedback extends Model {
             })
         ]);
     }
-    static findAllByDescription() {
-        return this.findAll({
+    static findAllByDescription(description = null) {
+        const whereClause = description
+            ? {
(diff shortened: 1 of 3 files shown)


db/models/feedback.js
```diff
@@ -39,10 +39,27 @@ class Feedback extends Model {
             })
         ]);
     }
-    static findAllByDescription() {
-        return this.findAll({
+    static findAllByDescription(description = null) {
```
davidrapson: I'd probably make this two methods, `findAllByDescription` which is the previous behaviour and `findByDescription` which is this new finder.

davidrapson: Maybe even rename `findAllByDescription` to `findAllGroupedByDescription` to make it  clear it's grouping results.

mattandrews: 👍 done

PR #2143, opened 2019-08-02
Convert form-components to use vue-i18n

davidrapson:
Converts our form components to use https://kazupon.github.io/vue-i18n/

As discussed I've settled on using a single `locales.json` file as it provides the best balance of simplicity and testability.

The main improvement is that it provides a consistent way of handling translation within vue components, avoiding some awkward prop passing and manual parsing of JSON, The main trade-off is that for components that have a basic server rendered equivalent we will have some duplicate copy between server and client locale files. I think is an acceptable trade-off to avoid the complexity of passing lots of copy props around.


assets/js/vue-apps/form-components/address-lookup.vue
```diff
@@ -82,23 +78,21 @@ export default {
...
+            return (
+                this.candidates.length === 0 &&
+                this.postcode &&
+                this.currentState !== states.NotRequired
+            );
+        }
+
+        $form.on('click', () => {
+            if (shouldWarn()) {
+                alert(this.$t('address.warningForIncomplete'));
```
mattandrews: Does, er, `this` definitely resolve properly here? I had issues with it originally (though I guess the arrow function might have solved it?).

davidrapson: Yeah resolves OK, but I introduced an error in the `shouldWarn` method so double checking this now.

PR #2162, opened 2019-08-05
Budget component improvements

mattandrews:
This fixes two things: 

- server-rendered errors (eg. budget limits exceeded) are now wrapped in `<noscript>` tags and the component renders its own errors on initialising, so they're stateful and disappear when fixed.

- a bug where new blank rows were added each time the form failed validation is fixed


assets/js/vue-apps/form-components/budget-input.vue
```diff
@@ -41,27 +53,36 @@ export default {
...
         }
     },
     methods: {
+        checkErrors() {
+            Vue.set(
+                this.error,
+                'TOO_MANY_ITEMS',
+                this.budgetRows.length === this.maxItems
+            );
+            Vue.set(
```
mattandrews: Had to use this rather than the old `this.error.TOO_MANY_ITEMS = true` syntax as this wasn't properly reactive when called in this way (eg. at pageload)

assets/js/vue-apps/form-components/budget-input.vue
```diff
@@ -20,16 +21,27 @@ export default {
     },
     data() {
         // Add a ready-to-use new row if the budget isn't over the limit already
-        const initialBudgetRows =
-            this.budgetData.length < this.maxItems
-                ? concat(this.budgetData, [{ item: '', cost: '' }])
-                : this.budgetData;
-
+        const shouldAddNewRow = () => {
```
davidrapson: We should probably move this to `methods` and then add a unit test for this as there's a fair bit of logic going on in here now.

assets/js/vue-apps/form-components/budget-input.vue
davidrapson: Hmm, this sounds like we should be setting an initial value for this in the `data()` set up step, then this could use the regular `this.` syntax.

Something like

```js
data() {
        // Add a ready-to-use new row if the budget isn't over the limit already
        const initialBudgetRows =
            this.budgetData.length < this.maxItems
                ? concat(this.budgetData, [{ item: '', cost: '' }])
                : this.budgetData;

        return {
            budgetRows: initialBudgetRows,
            error: {
                TOO_MANY_ITEMS: initialBudgetRows.length === this.maxItems,
                OVER_BUDGET: this.maxBudget && this.total > this.maxBudget
            }
        };
    },
```

PR #2163, opened 2019-08-05
Basic users dashboard

davidrapson:
image


davidrapson:
This is all-time, but just realised I probably want to cap this to a reasonable date range as we have some early users (us) that were registered over a year ago.

davidrapson:
image

PR #2167, opened 2019-08-05
Travis speed debug

davidrapson:
A bit of trial and error to chip away at build times (again)

- Tested switching to Ubuntu `bionic`, not a hugely noticeable change but doesn't hurt to keep up to date
- Determined that `concurrently` actually slows things down for our current use case. CI servers are quite low power so I'm assuming the running tasks in parallel is more expensive than running regularly.
- Re-merged cypress test suites and streamline the test a little bit
- Remove a unit test that required a client-side build which allows us to build earlier

Seems like there's still general slow down with instances booting up as it's still ~2 minutes from the VM starting to boot to it getting to any build steps so still slower than I'd like but better.

Overall cuts about 1-1.5 minutes off the build time.


controllers/apply/reaching-communities/processor.test.js
```diff
@@ -1,150 +0,0 @@
...
-    const processorOptions = {
-        data: flattenFormData(mockFormData),
-        copy: formCopy,
-        stepsWithValues: stepsWithValues(form.steps, mockFormData)
-    };
-
-    return processor(processorOptions, mockTransport);
-}
-
-describe('reaching communities processor', () => {
```
davidrapson: I feel _OK_ about removing this test as it's very fake i.e. mocks far too many things and requires a client-side build to work which is an anti-pattern but appreciate it does add _some_ value to open to looking at other ways to test this before merging. 💭

PR #2176, opened 2019-08-08
Fix a minor UX snag where conditional address radio defaulted to "yes"

mattandrews:
Suggested by @LynseyJReynolds – we default this question to "yes":

image

This isn't the case for the non-JS version so I changed it here to make it more explicit (eg. so the customer doesn't just skip over it).


assets/js/vue-apps/form-components/address-lookup.vue
```diff
@@ -52,8 +52,11 @@ export default {
                     'Apply: AFA: Contacts: User needs address history'
                 ]);
             }
+
             if (value === 'yes') {
                 this.currentState = states.NotRequired;
+            } else if (value === 'unknown') {
```
davidrapson: Not a big fan of this, could we not rely on the value being undefined rather than introducting a false value?

Would something like this work?

```js
if (value === undefined || value === 'yes') {
 this.currentState = states.NotRequired;
}
```

mattandrews: 👍 – was even simpler as `!value` does what we need.

PR #2177, opened 2019-08-08
Add a field for confirmation about different contacts

mattandrews:
As suggested by @scottoakley, this adds a new question to the Main Contact screen to ask the user to confirm the contacts aren't related.

MicrosoftTeams-image

I don't think we need to pass this field to Salesforce, though?

It's off by default and only enabled in TEST so we can verify it with the team.

### DIFF
--- a/controllers/apply/awards-for-all/docs/schema.md
+++ b/controllers/apply/awards-for-all/docs/schema.md
@@ -94,6 +94,7 @@ Each submission has two top-level keys: `meta` which contains metadata about the
             "firstName": "Nelda",
             "lastName": "Nolan"
         },
+        "mainContactIsValid": "yes",
         "mainContactDateOfBirth": "1975-10-12",
         "mainContactAddress": {
             "line1": "41465 Bashirian Oval",
@@ -427,6 +428,11 @@ Object with properties:
 | **firstName** | `string` |
 | **lastName**  | `string` |
 
+### mainContactIsValid
+
+type: `string`
+
+Will always be "yes" (required field)
 
 ### mainContactDateOfBirth
 
--- a/controllers/apply/awards-for-all/docs/schema.md
+++ b/controllers/apply/awards-for-all/docs/schema.md
@@ -94,7 +94,6 @@ Each submission has two top-level keys: `meta` which contains metadata about the
             "firstName": "Nelda",
             "lastName": "Nolan"
         },
-        "mainContactIsValid": "yes",
         "mainContactDateOfBirth": "1975-10-12",
         "mainContactAddress": {
             "line1": "41465 Bashirian Oval",
@@ -428,11 +427,6 @@ Object with properties:
 | **firstName** | `string` |
 | **lastName**  | `string` |
 
-### mainContactIsValid
-
-type: `string`
-
-Will always be "yes" (required fields)
 
 ### mainContactDateOfBirth
 
--- a/controllers/apply/awards-for-all/docs/schema.md
+++ b/controllers/apply/awards-for-all/docs/schema.md
@@ -94,6 +94,7 @@ Each submission has two top-level keys: `meta` which contains metadata about the
             "firstName": "Nelda",
             "lastName": "Nolan"
         },
+        "mainContactIsValid": "yes",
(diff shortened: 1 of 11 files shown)


controllers/apply/awards-for-all/form.js
```diff
@@ -624,16 +626,10 @@ module.exports = function({
...
-                            'seniorContactName.firstName'
-                        )(data);
-                        const seniorSurname = get('seniorContactName.lastName')(
-                            data
-                        );
-                        const seniorName =
-                            seniorFirstName && seniorSurname
-                                ? safeHtml`, <strong data-hj-suppress>${seniorFirstName} ${seniorSurname}</strong>`
-                                : '';
+                        const seniorName = getContactFullName(data, 'senior');
```
mattandrews: I extracted this into a function as initially we repeated the senior contact name in the validation error message, though this subsequently changed. Keeping it as a function though as I can see this coming back in future.

views/components/form-fields-next/macros.njk
```diff
@@ -638,7 +638,7 @@
                 {% endif %}
                 <ol class="form-field__errors">
                     {% for error in fieldErrors %}
-                        <li>{{ error.msg }}</li>
+                        <li>{{ error.msg | safe }}</li>
```
mattandrews: This is to allow the `<strong>` tags in the contact name above.

controllers/apply/awards-for-all/docs/schema.md
```diff
@@ -427,6 +428,12 @@ Object with properties:
 | **firstName** | `string` |
 | **lastName**  | `string` |
 
+### mainContactIsValid
```
davidrapson: I'd `.strip()` this from the schema completely.

controllers/apply/awards-for-all/fields.js
```diff
@@ -2057,6 +2057,38 @@ module.exports = function fieldsFor({ locale, data = {} }) {
                 }
             ]
         },
+        mainContactIsValid: {
```
davidrapson: This flag needs to be wrapped around this entry too. The validation schema is built from the `allFields` object, so this will force this field to be required even if the field isn't shown unless we wrap this in a flag.

controllers/apply/awards-for-all/fields.js
```diff
@@ -2057,6 +2057,38 @@ module.exports = function fieldsFor({ locale, data = {} }) {
...
+                    value: 'yes',
+                    label: localise({
+                        en: 'Yes',
+                        cy: ''
+                    })
+                }
+            ],
+            isRequired: true,
+            get schema() {
+                return multiChoice(this.options).required();
```
davidrapson: I'd try adding a `.strip()` to this as we don't really need/want this in the resulting schema, just for the validations.

controllers/apply/awards-for-all/lib/contacts.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+const get = require('lodash/fp/get');
+const { safeHtml } = require('common-tags');
+
+function getContactFullName(data, type) {
+    const fieldName =
```
davidrapson: Rather than varying on the type could we pass the `seniorContactName` or `mainContactName` data in directly?

cypress/integration/awards-for-all.js
```diff
@@ -477,11 +477,17 @@ describe('awards for all', function() {
             });
         }
 
-        function fillContact(contact) {
+        function fillContact(contact, type) {
             cy.getByLabelText('First name').type(contact.firstName);
 
             cy.getByLabelText('Last name').type(contact.lastName);
 
+            if (type === 'main') {
```
davidrapson: Rather than looking at the `type` it's better to add a check for if the field exists (i.e. these tests are better off working out what to do based on what's on the page, not needing to know about the underlying logic)

controllers/apply/awards-for-all/fields.js
mattandrews: This prevents submission so we can't do this!

davidrapson: Not sure I follow?

PR #2178, opened 2019-08-08
Improved Bank API error handling

NikhilNanjappa:
- Unit tests for VALID, INVALID_BACS, INVALID_ACCOUNT, bad response
- simplified API response to return a single `statusCode`.
- correctly threw an error for bad responses. eg. where the API key/pwd is invalid.
- moved the `bank-api.js` from `./common` to `awards-for-all/lib` folder as its not used elsewhere.

### DIFF
--- a/common/secrets.js
+++ b/common/secrets.js
@@ -79,8 +79,8 @@ const SALESFORCE_AUTH = {
 
 // These expire in July 2020
 const BANK_API = {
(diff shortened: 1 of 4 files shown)


davidrapson:
Nikhil and I paired on this one so would appreciate your review @mattandrews

PR #2184, opened 2019-08-12
Custom phone number validator

### LINKED ISSUE #2183: Phone number validation not working
### What
For the main/senior contact's phone number field, if we enter the value "4444", it passes validation.

Another scenario when the long number (of say 85 characters) long is added, the form on submission hangs.

### Solutions Implemented
Already tried adding the `.max` validation to the field but the behavior persisted.

### Expected
Should fail validation on all the above occasions.

davidrapson:
Early morning train…

As @NikhilNanjappa spotted, `joi-phone-number` was only parsing the number not checking it it so was lax in its validation. The code for the validation is straightforwards as it calls `google-libphonenumber` directly and write our own validator which calls `phoneUtil.isValidNumber` to improve the validation.

Fixes #2183


controllers/apply/form-router-next/joi-extensions/phone-number.js
```diff
@@ -0,0 +1,54 @@
...
+        name: 'string',
+        language: {
+            phonenumber: 'did not seem to be a phone number'
+        },
+        rules: [
+            {
+                name: 'phoneNumber',
+                validate(params, value, state, options) {
+                    try {
+                        const parsedValue = phoneUtil.parse(value, 'GB');
```
davidrapson: We only care about uk numbers so can avoid the abstraction of supporting configurable options by hardcoding.

davidrapson: We only care about uk numbers so can avoid the abstraction of supporting configurable options by hardcoding.

package.json
```diff
@@ -112,7 +112,7 @@
     "nodemailer": "^4.1.0",
     "nunjucks": "^3.1.3",
     "passport": "^0.4.0",
-    "passport-azure-ad": "4.0.0",
+    "passport-azure-ad": "4.1.0",
```
davidrapson: I didn’t mean to commit this here, but it’s safe to include.

controllers/apply/form-router-next/joi-extensions/phone-number.js
NikhilNanjappa: Makes sense!

NikhilNanjappa: +1

PR #2186, opened 2019-08-12
Local AV check using clamscan package

NikhilNanjappa:
- `provision.sh` changed to add clamav and clamd installation
- `development.json` changed to use the ec2 paths
- `cloudwatch-agent.json` changed to use the clamd logs
- Added a feature flag to toggle local and s3 AV checks
- Added a package clamscan
- Local AV check makes use of local clamd server and clamsan package
- Created a separate `scanAndUpload` and `scanFile` function for modularity.


controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -19,6 +20,22 @@ const s3 = new AWS.S3({
     region: 'eu-west-2'
 });
 
+const ClamScan = new NodeClam().init({
+    debug_mode: true,
+    scan_recursively: false,
+    clamdscan: {
+        socket: '/usr/local/etc/clamd.socket',
```
davidrapson: We'll want any paths to come from secrets or the environment so we can determine where to look depending on the environment (e.g. EC2 or locally)

NikhilNanjappa: Good shout! I will move these to the environment files as these are env related details.

controllers/apply/form-router-next/index.js
```diff
@@ -393,7 +394,9 @@ function initFormRouter({
                             };
 
                             try {
-                                await checkAntiVirus(pathConfig);
+                                if(!config.get('features.enableLocalAntivirus') && !appData.isTestServer) {
```
davidrapson: I'd wrap the whole try/catch in the check rather than just the await

controllers/apply/form-router-next/steps.js
```diff
@@ -218,13 +220,29 @@ module.exports = function(formId, formBuilder) {
...
+                                        const avStatus = await checkAntiVirusClamD(formId, currentlyEditingId, file);
+
+                                        if(avStatus.Value === 'CLEAN') {
+                                            uploadFile({
+                                                formId: formId,
+                                                applicationId: currentlyEditingId,
+                                                fileMetadata: file
+                                            });
+                                        } else {
+                                            logger.error('File upload skipped', {
```
davidrapson: We'll want to do more than just a `logger.error` here. Need to give this a little thought as to how we shout loudly if we find a virus.

NikhilNanjappa: Sure ... I followed the same behavior as the existing AV function on finding infected file.

config/development.json
```diff
@@ -1,5 +1,11 @@
 {
   "logLevel": "debug",
+  "clamdscan": {
```
davidrapson: These need switching round. `development.json` will never get run on EC2. We'd want the default settings to contain the standard ubuntu paths, and probably don't need anything in development.json

controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -19,6 +20,22 @@ const s3 = new AWS.S3({
...
+const ClamScan = new NodeClam().init({
+    debug_mode: false,
+    scan_recursively: false,
+    clamdscan: {
+        socket: config.get('clamdscan.socket'),
+        timeout: config.get('clamdscan.timeout'),
+        local_fallback: true,
+        path: config.get('clamdscan.path'),
+        config_file: config.get('clamdscan.config_file'),
+        multiscan: false,
```
davidrapson: Do we definitely need to set these subsequent params, can we rely on the defaults?

controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -19,6 +20,22 @@ const s3 = new AWS.S3({
...
 });
 
+const ClamScan = new NodeClam().init({
+    debug_mode: false,
+    scan_recursively: false,
+    clamdscan: {
+        socket: config.get('clamdscan.socket'),
+        timeout: config.get('clamdscan.timeout'),
+        local_fallback: true,
+        path: config.get('clamdscan.path'),
```
davidrapson: I'd probably set these config.get calls to be `process.env.CLAMDSCAN_PATH || config.get('clamdscan.path')` etc. so we can set these values locally. Using `development.json` for this doesn't really work as local config will be specific to our individual set ups.

controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -19,6 +20,22 @@ const s3 = new AWS.S3({
     region: 'eu-west-2'
 });
 
+const ClamScan = new NodeClam().init({
+    debug_mode: false,
```
davidrapson: Could we set `debug_mode` based on `appData.isDev`?

controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -109,28 +126,50 @@ function uploadFile({ formId, applicationId, fileMetadata }) {
     });
 }
 
+function checkAntiVirusClamD(formId, applicationId, fileMetadata) {
+    const keyName = [formId, applicationId, fileMetadata.fileData.name].join('/');
```
davidrapson: This `keyName` is the potential path on S3 only makes sense for the previous scan method. `fileMetadata.fileData.path` is the path we are interested in

controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -109,28 +126,50 @@ function uploadFile({ formId, applicationId, fileMetadata }) {
...
+        } = await clamscan.scan_file(fileMetadata.fileData.path);
+
+        if (is_infected) {
+            logger.error(`${keyName} is infected with ${viruses}`);
+            throw new Error('ERR_FILE_SCAN_INFECTED');
+        } else {
+            return { Key: 'av-status', Value: 'CLEAN' };
+        }
+    }).catch(err => {
+        logger.error(`Attempted read of unscanned file at ${keyName} - ${err.message}`);
```
davidrapson: We can write this as `logger.error(`Attempted read of unscanned file at ${keyName}`, err)`. `logger.error` takes an `Error` object as the second argument

controllers/apply/form-router-next/lib/file-uploads.js
NikhilNanjappa: Most of these values have tradeoffs to be honest. 

`local_fallback` by default is `false` ie. if local socket fails, it won't fall back to the binary and hence kept it as `true` as we might need a fallback. I'll remove `multiscan` which defaults to multi-thread scanning. I'll remove `reload_db` which by default will **not** look for updated definitions but will be faster. I'll also remove `active` and `bypass_test`.

controllers/apply/form-router-next/lib/file-uploads.js
NikhilNanjappa: Sure. Would it be better if I use the `fileMetadata.fileData.name` instead as I'm not sure if metadata eg. `/var/folders/dn/844xbhwn2gqf37zpf7_1nv2h0000gn/T/upload_eacc44e2fb52ef868b9c3a84c22143b8` would be of any use ?

davidrapson: Ah yeah good shout

controllers/apply/form-router-next/steps.js
davidrapson: Thinking about this I think we are fine here we can used cloudwatch or datadog alerts to give the level of alerting we need for this.

controllers/apply/form-router-next/lib/file-uploads.js
davidrapson: 👍

controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -5,11 +5,12 @@ const AWS = require('aws-sdk');
...
 const keyBy = require('lodash/keyBy');
 const mapValues = require('lodash/mapValues');
+const NodeClam = require('clamscan');
 
-const { isTestServer } = require('../../../../common/appData');
+const { isTestServer, isDev } = require('../../../../common/appData');
 const { S3_KMS_KEY_ID } = require('../../../../common/secrets');
 const logger = require('../../../../common/logger').child({
-    service: 's3-uploads'
+    service: 'file-uploads'
```
mattandrews: Just a note to remind us to update the dashboard(s) when this goes live as the metrics there will refer to the old name.

deploy/cloudwatch-agent.json
```diff
@@ -15,6 +15,11 @@
             "file_path": "/var/log/nginx/access.log",
             "log_group_name": "/tnlcf/APP_ENV/nginx/access",
             "log_stream_name": "{instance_id}"
+          },
+          {
+            "file_path": "/var/log/clamav/clamd.log",
+            "log_group_name": "/tnlcf/APP_ENV/clamav/clamd",
```
mattandrews: This will need to be added to the DataDog Lambda function so these logs are pushed there.

davidrapson:
[Tech debt: Revisit approach for antivirus scanning](https://trello.com/c/vvhhqYqm/738-tech-debt-revisit-approach-for-antivirus-scanning)

PR #2198, opened 2019-08-15
Simplify joi-extensions

davidrapson:
We had a code smell in our joi-extensions where we were relying on parent state in order to validate related fields, making these validations tightly coupled on the awards for all form.

This was based on the assumption that it was not possible to pass a Joi.Ref as an argument to a validation, which **is** possible after all.

This PR reworks our extensions (with additional tests where needed to back up the changes) to avoid any dependency on internal state objects


controllers/apply/awards-for-all/fields.js
```diff
@@ -1019,6 +1046,65 @@ module.exports = function fieldsFor({ locale, data = {} }) {
         };
     }
 
+    function fieldProjectTotalCosts() {
+        const budgetTotal = sumBy(
```
davidrapson: We can compute the project total here allowing us to do the min budget check using a standard `min` validation rather than needing a custom extension.

controllers/apply/awards-for-all/fields.js
```diff
@@ -1019,6 +1046,65 @@ module.exports = function fieldsFor({ locale, data = {} }) {
...
+                    getting £10,000 from another funder to cover additional costs,
+                    then your total project cost is £18,000.
+                    If you are asking us for £8,000 and there are no other costs
+                    then your total project cost is £8,000.
+                </p>`,
+                cy: ``
+            }),
+            type: 'currency',
+            isRequired: true,
+            schema: Joi.friendlyNumber()
```
davidrapson: `friendlyNumber` is a new low level extension which is solely responsible for allowing `,` and spaces in numbers and coercing it to a number

controllers/apply/awards-for-all/fields.js
```diff
@@ -2093,76 +2139,75 @@ module.exports = function fieldsFor({ locale, data = {} }) {
...
-                    .mainContact()
-                    .required()
+                return result;
             },
-            [
+            schema: Joi.object({
+                firstName: Joi.string().required(),
+                lastName: Joi.string().required()
+            })
+                .compare(Joi.ref('seniorContactName'))
```
davidrapson: We have a new low level `compare-object` extension which allows us to pass a ref to do a deep `isEqual` check

controllers/apply/form-router-next/joi-extensions/bank-numbers.js
```diff
@@ -1,55 +0,0 @@
-'use strict';
-const { isString } = require('lodash');
```
davidrapson: This whole file was never used.

controllers/apply/awards-for-all/fields.js
NikhilNanjappa: Does it also allow spaces in numbers ?

NikhilNanjappa: I have also noticed, if you pass a value like `21ab5c6.00` to friendly-number, we assume its correct and convert it to `2156` whilst not throwing any validation error. Is this intended ?

controllers/apply/form-router-next/joi-extensions/compare-object.js
```diff
@@ -0,0 +1,37 @@
...
+            isEqual: 'Objects must not match'
+        },
+        rules: [
+            {
+                name: 'compare',
+                params: {
+                    ref: joi.func().ref()
+                },
+                validate(params, value, state, options) {
+                    const refVal = params.ref(
```
mattandrews: 💯 

I was [trying to figure this out](https://stackoverflow.com/questions/56689438/comparing-two-objects-in-joi-validation-eg-to-avoid-duplicates) and couldn't, but *knew* there'd be a smart way to get the reference value. Nice job! 👍

davidrapson:
Closing this but will re-open a fresh cut of this (plenty of things have moved following the welsh translations)

PR #2203, opened 2019-08-16
Merged scanAndUpload function

davidrapson:
Had a go a merging the scan and upload methods back together. Noticed that the promise code could be simplified doing this as well. Thoughts welcome 💭 

image

### DIFF
--- a/controllers/apply/form-router-next/lib/file-uploads.js
+++ b/controllers/apply/form-router-next/lib/file-uploads.js
@@ -20,22 +20,6 @@ const s3 = new AWS.S3({
     region: 'eu-west-2'
 });
 
-const ClamScan = new NodeClam().init({
-    debug_mode: false,
-    scan_recursively: false,
-    clamdscan: {
-        socket: config.get('clamdscan.socket'),
-        timeout: config.get('clamdscan.timeout'),
-        local_fallback: true,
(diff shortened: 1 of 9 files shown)


controllers/apply/form-router-next/lib/file-uploads.js
```diff
@@ -122,48 +110,80 @@ function uploadFile({ formId, applicationId, fileMetadata }) {
...
+                throw new Error('ERR_FILE_SCAN_INFECTED');
+            } else {
+                logger.error(`Attempted read of unscanned file at ${keyName}`);
+                throw new Error('ERR_FILE_SCAN_UNKNOWN');
+            }
+        });
+}
+
+function scanAndUpload({ formId, applicationId, fileMetadata }) {
+    function shouldScan() {
```
NikhilNanjappa: Can we export/re-use this function for toggling the `await checkAntiVirus` in `index.js` or make this function usable within the `checkAntiVirus` function of `file-upload.js` ?

davidrapson: The logic for that function is different so not sure we could reuse it

NikhilNanjappa: I meant the `shouldScan` function.

PR #2211, opened 2019-08-20
Add clamav install wrapped in test group check

davidrapson:
Thinking about how we can unblock https://github.com/biglotteryfund/blf-alpha/pull/2186 and the simplest way to test clamav provisioning only in the test environment might be to wrap in a deploy group check. Is this as bad idea? 💭

### DIFF
--- a/deploy/provision.sh
+++ b/deploy/provision.sh
@@ -8,6 +8,15 @@ sh -c 'echo deb https://oss-binaries.phusionpassenger.com/apt/passenger xenial m
 apt-get update
 apt-get install -y nginx-extras passenger
 
+# Installing clamav and clamav-daemon (anti-virus scan)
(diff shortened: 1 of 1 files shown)


NikhilNanjappa:
Sure 👍 I was also thinking from the future perspective, would it be better to have another "dev" environment where we can deploy & test any such tech-debts ? (I'm guessing there might be more additions like this in the future where the need for a separate env might arise)

davidrapson:
@NikhilNanjappa I tend to take the view that additional environments should be pretty exceptional, so maybe but prefer to avoid it. The further away from production we are the more artificial things get, and it's another environment to maintain.

NikhilNanjappa:
Sure ... Just thinking out loud :)

PR #2220, opened 2019-08-22
Add a question about contact language preference

mattandrews:
Adds a new question for main and senior contacts about language preference:

image

~There's a `// @TODO` in here to add a missing translation which I've asked for~, but other than that it should be good to go. 

Next steps will be interpreting this data at the Salesforce end to use it to decide what language to email these contacts in.


controllers/apply/awards-for-all/fields.js
```diff
@@ -630,6 +630,52 @@ module.exports = function fieldsFor({ locale, data = {} }) {
         };
     }
 
+    function fieldContactLanguagePreference(props) {
```
davidrapson: Might be useful to extract this out into `fields/`, aiming to break things out gradually so would be good to do for new fields.

davidrapson:
[Determine approach for welsh language communication preferences](https://trello.com/c/d6X2f7X4/751-determine-approach-for-welsh-language-communication-preferences)

controllers/apply/awards-for-all/fields.js
mattandrews: Good call, give it a look now (plus tests).

controllers/apply/awards-for-all/form.test.js
```diff
@@ -1025,6 +1025,37 @@ describe('Contacts', () => {
...
         }
     );
+
+    test.each([
+        'mainContactLanguagePreference',
+        'seniorContactLanguagePreference'
+    ])('%p must exist and be a valid choice for Wales', function(fieldName) {
+        assertValidByKey({
+            projectCountry: 'england',
+            [fieldName]: 'wales'
```
davidrapson: This test looks a bit wonky? Should this be `null` or something else?

mattandrews: Oops this was me debugging a test, sorry! Will fix.

NikhilNanjappa:
Awesome ... looks good to me 👍

davidrapson:
@NikhilNanjappa 🚀 Ship is

PR #2221, opened 2019-08-22
Add locale to salesforce meta

davidrapson:
Add locale to salesforce meta. 🌐


davidrapson:
Cherry picked this change into https://github.com/biglotteryfund/blf-alpha/pull/2220 so we can handle welsh language schema changes together.

PR #2222, opened 2019-08-22
Travis tweaks

davidrapson:
Makes a few tweaks to the travis build that I'd tested before but we never merged.

**Remove concurrently from travis steps**
Confirmed that our use of `concurrently` actually slows things down in a CI environment

**Remove a test which requires a client-side build**
`processor.test.js` was heavily mocked and dependent on built-css to the point that I'm not convinced it actually tests anything.

**Remove percy**
Bit of a heavy heart on this one as I think Percy are a really lovely service, but relatively to the time it adds to builds I'm not convinced it gives us enough value. We have low churn on our CSS, and the tests don't cover high value things like the awards for all form due to the challenge of doing visual regression testing on dynamic content. Plus the screenshots are from Chrome only which is pretty over-served in our testing.

**Re-merge integration_spec**
Enough of a slowdown between cypress runs that it's worth the faff of having a single spec file.


davidrapson:
About ~3 minutes quicker

image

PR #2223, opened 2019-08-23
Launch AFA Wales

mattandrews:
This will need to go live when we launch in Wales and not before, but it tweaks the location message wording to clarify that both Wales and Scotland should use this new form.

image


mattandrews:
@davidrapson have tweaked this to undo the workaround in https://github.com/biglotteryfund/blf-alpha/pull/2271 and make Wales available to all, so this becomes the launch PR.

PR #2241, opened 2019-08-23
Disable html5 validations

NikhilNanjappa:
Added `noValidate` flag to all the steps except the ones that contain address-lookup (organisation details step, senior & main contact step).

As part of this, found an unrelated issue with address-lookup validation - The 'remove address' link does not clear the values from the object until another valid address is submitted (need to create a ticket to capture this).


NikhilNanjappa:
As expected, postcode lookup is the only element which behaves oddly (will attach a file replicating the behavior soon)

controllers/apply/form-router-next/views/step.njk
```diff
@@ -26,8 +26,9 @@
                   method="post"
                   class="js-form-warn-unsaved js-apply-{{ formId }}"
                   {% if errors | length > 0 %}data-form-has-errors="1"{% endif %}
-                  {% if step.isMultipart %}enctype="multipart/form-data"{% endif %}
-            >
+                  {% if step.isMultipart %}enctype="multipart/form-data"{% endif %} 
+                  {% if step.noValidate %}novalidate{% endif %}
```
NikhilNanjappa: Adding `novalidate` attribute if the flag is truthy.

controllers/apply/form-router-next/lib/validate-model.js
```diff
@@ -60,6 +60,7 @@ module.exports = function validateModel(formModel) {
 
     const stepSchema = Joi.object({
         title: Joi.string().required(),
+        noValidate: Joi.boolean().optional(),
```
NikhilNanjappa: Added `noValidate` to the schema to pass schema validation

PR #2244, opened 2019-08-23
More welsh amends

davidrapson:
The main change here is allowing formatters to be localised. This feels a tiny bit awkward and it feels like we could probably model the formatters a little differently, but as an initial approach I think this works OK.


mattandrews:
Just added a tweak from the most recent translations as it was specific to the formatters and made more sense in here than in my branch.

PR #2250, opened 2019-08-27
Session expiry extension / warnings

mattandrews:
There's a few things going on here – the main ones are these:

- If a user enters a page with a form that has a class of `js-session-form`, then any time they press a key or click anywhere, their cookie expiry is reset (back to 2 hours)

- There's also a timer checking on a 30 second interval to see if they're near expiry. When they reach 5 minutes remaining on their session, they'll see a modal window* warning them. Clicking on this window triggers the above handler and resets their expiry back to 2 hours again.

image

- If they make no action for 2 hours, their session expires and they'll be shown a modal telling them their work is lost. 

image

* At the moment, the modal functionality is disabled – I'd like to see what effect the auto-extension of the session has on the number of people being inadvertently logged out. The modals are basically a bit of a fail-safe for people who walk away from their computer for hours – they'll see a message onscreen telling them what happened, but if they'd been using the form actively they should never encounter it.

Also, the modals aren't translated yet so we'd need this before enabling them. For now I've added tracking events when they would be shown so we can monitor real-world frequency.


views/layouts/main.njk
```diff
@@ -15,6 +15,28 @@
             </ul>
         </nav>
 
+        <div class="modal js-modal u-hidden">
```
NikhilNanjappa: Do we have to meet the progressive enhancement & accessibility needs ?

As mentioned in this guidance - https://docs.google.com/document/d/16BEBhmodDoUyIeyVAfyj3rMX45Lw0Zo5rs0pWC3_5Mw/edit.

assets/js/common/session.js
```diff
@@ -0,0 +1,87 @@
...
+import $ from 'jquery';
+import debounce from 'lodash/debounce';
+import modal from './modal';
+import { trackEvent } from '../helpers/metrics';
+
+// Deploy switch - allows us to roll out this feature in phases
+const showModals = false;
+
+const expiryCheckIntervalSeconds = 30; // The interval we'll check timeouts in
+const warningShownSecondsRemaining = 5 * 60; // The time before logout we'll show a warning at
```
NikhilNanjappa: We could move these values to an env driven file.

views/layouts/main.njk
mattandrews: Great point – this would definitely need some attention before we launch it. I think we can merge this as-is now but if/when we decide to *use* the modal (we might not need to!) we should look at this.

mattandrews:
@davidrapson thoughts please!

mattandrews:
Closed in favour of https://github.com/biglotteryfund/blf-alpha/pull/2255

PR #2255, opened 2019-08-28
Auto extend session

mattandrews:
Forked from https://github.com/biglotteryfund/blf-alpha/pull/2250 – removes the modal functionality until we've had time to do it to a higher standard (when, though?).


package.json
```diff
@@ -80,7 +80,6 @@
     "config": "3.0.1",
     "connect-session-sequelize": "6.0.0",
     "connect-slashes": "^1.4.0",
-    "cookie-parser": "1.4.4",
```
mattandrews: This is no longer required for express-session, in fact they recommend not using it if you don't need it for anything else (which we don't, eg. we don't use `req.cookies` anywhere).

assets/js/common/session.js
```diff
@@ -0,0 +1,71 @@
...
+        } else if (expiryTimeRemaining <= warningShownSecondsRemaining) {
+            // The user has a few minutes remaining before logout
+            trackEvent('Session', 'Warning', 'Timeout almost reached');
+        }
+    }
+
+    const handleActivity = () => {
+        if (isAuthenticated) {
+            // Extend their session
+            $.ajax({
```
davidrapson: Does this need `withCredentials` or similar, or does `$.ajax` send through the session cookie with the request by default?

config/default.json
```diff
@@ -55,7 +55,8 @@
     "showContactConfirmationQuestion": false
   },
   "cookies": {
-    "session": "blf-alpha-session"
+    "session": "blf-alpha-session",
+    "expiryMs": 7200000
```
davidrapson: I'd move this config elsewhere, the reads like there is a cookie named `7200000`. Would be good to name it more explicitly for session expiry

cypress/integration/integration_spec.js
```diff
@@ -153,7 +153,10 @@ describe('user', () => {
             logIn(newUser.username, newUser.password);
 
             // Log out
-            cy.getByText('Log out').click();
+            cy.get('.user-nav__links')
```
davidrapson: If possible add a data-testid to this element to avoid needing a CSS hook here

package.json
davidrapson: _Interesting_ 👍 always up  for removing dependencies

views/includes/metaHead.njk
```diff
@@ -8,6 +8,7 @@
         environment: '{{ appData.environment }}',
         locale: '{{ locale }}',
         localePrefix: '{{ localePrefix }}',
+        sessionExpirySeconds: {{ appData.config.get('cookies.expiryMs') / 1000 }}
```
davidrapson: Should we set this in seconds rather than needing a calculation in a template?

assets/js/common/session.js
mattandrews: When it's using the same domain as the current page, cookies are included by default so we don't need to change anything.

cypress/integration/integration_spec.js
mattandrews: Not possible (I tried this first) – those nav links are generated from a loop/array.

views/includes/metaHead.njk
mattandrews: Perhaps – will have to just invert the calculation in the session config instead.

mattandrews: ... done!

cypress/integration/integration_spec.js
davidrapson: We control how this is generated though, so we could definitely add a hook in _somewhere_ not a big deal for now though.

PR #2256, opened 2019-08-28
Application Expiry

NikhilNanjappa:
Done:
- An endpoint which looks up all pending applications due to expire in 30 days or less.
- handle month/week/day expiry .ie. fetch email addresses + send emails + update records.
- handle expired .ie. delete records (no emails sent).
- Switching SQL statements based on dialect.
- Added email body

TODO:
- "opt-out" of emails feature ?

**Endpoint Response:**

Screenshot 2019-09-10 at 13 35 07

On error:
Screenshot 2019-09-11 at 13 31 46

**Email:**
Screenshot 2019-09-18 at 16 58 41


controllers/api/index.js
```diff
@@ -150,4 +151,51 @@ router.post('/survey', async (req, res) => {
...
+router.post('/applications/expiry', async (req, res) => {
+    try {
+        const expiryApplications = await PendingApplication.findApplicationsByExpiry();
+        const expiredReminderUserIds = [],
+        dayReminderUserIds = [],
+        weekReminderUserIds = [],
+        monthReminderUserIds = [],
+        expiredApplicationIds = [];
+        
+        expiryApplications.forEach((application) => {
```
davidrapson: Not sure how I feel about this code, S=switching on `true` is a bit confusing and feels like these conditions are queries. Could we handle this logic in database queries, i.e. query for each condition separately?

db/models/applications.js
```diff
@@ -87,6 +87,18 @@ class PendingApplication extends Model {
...
+        const endDate = moment().add('1', 'months').toDate();
+
+        return this.findAll({
+            where: {
+                expiresAt: {
+                    [Op.between]: [startDate, endDate]
+                }
+            }
+        });
+    }
```
mattandrews: Not sure if this is helpful but the following SQL query will give you the number of days until each application expires:

`SELECT DATEDIFF(expiresAt, CURDATE()) AS daysTilExpiry FROM PendingApplications`

... though I'm not sure how to translate that into Sequelize syntax...

NikhilNanjappa: Yup, I used the same query and the syntax I used is the sequelize translation of it. Though,  unfortunately I couldn't find any confirmation through their docs. Had to dig a lot to find the date compare example in their issues list.

db/models/applications.js
```diff
@@ -87,6 +87,19 @@ class PendingApplication extends Model {
             order: [['createdAt', 'DESC']]
         });
     }
+    static findApplicationsByExpiry(from, to) {
+        let rawSqlStmt = `DATEDIFF(expiresAt, CURDATE()) >= ${from} AND DATEDIFF(expiresAt, CURDATE()) <= ${to}`;
+        // const startDate = moment().toDate();
+        // const endDate = moment().add('30', 'days').toDate();
+
+        if (from === 'expired') {
```
davidrapson: It would be good to split the expired query out rather than relying on a magic string.

mattandrews:
Closing in favour of https://github.com/biglotteryfund/blf-alpha/pull/2371

PR #2267, opened 2019-08-29
Fix typo in West and East Dunbartonshire

davidrapson:
Fix typo in West and East Dunbartonshire. Need to double check if the schema change needs to be updated on the Salesforce end first, if so I'll update to make only the label change in the mean time.


davidrapson:
Will re-open this once we've confirmed the dependent salesforce change has been made

PR #2276, opened 2019-08-30
Clone the minimum project start date before modifying it

mattandrews:
Here's a funny one, spotted by Osian:

image

Logging `minDate` each time we format it, we see:

```
moment("2020-01-03T11:55:15.083")
moment("2020-01-02T11:55:15.083")
moment("2020-01-01T11:55:15.083")
moment("2019-12-31T11:55:15.083")
```

eg. each time we called `formatAfterDate()` it was removing a day from the date, which meant the validations would change each time too...!

This change calls `clone()` on `minDate` so we're not modifying it. I checked the codebase and everywhere else we use `.add()` or `.subtract()`, we're either doing it against the default `moment()` object or already cloning the original date, so we should be safe.

Afterwards:

```
moment("2020-01-03T11:56:28.774")
moment("2020-01-03T11:56:28.774")
moment("2020-01-03T11:56:28.774")
moment("2020-01-03T11:56:28.774")
```

### DIFF
--- a/controllers/apply/awards-for-all/fields/organisation-type.js
+++ b/controllers/apply/awards-for-all/fields/organisation-type.js
@@ -66,7 +66,7 @@ module.exports = function fieldOrganisationType(locale) {
                 en: oneLine`My organisation is a not-for-profit company
                     registered with Companies House, and <strong>may also</strong>
                     be registered as a charity`,
-                cy: oneLine`Mae fy sefydliad yn gwmni dielw sydd yn gofrestredig
+                cy: oneLine`Mae fy sefydliad yn gwmni di-elw sydd yn gofrestredig
                     â Thŷ’r Cwmnïau, a <strong>gall hefyd</strong> fod wedi’i
                     gofrestru fel elusen.`
             })
--- a/controllers/apply/awards-for-all/form.test.js
+++ b/controllers/apply/awards-for-all/form.test.js
@@ -329,7 +329,7 @@ describe('Project details', () => {
                     endDate: { day: 1, month: 1, year: 2030 }
                 }
             },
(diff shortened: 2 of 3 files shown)


controllers/apply/awards-for-all/fields/organisation-type.js
```diff
@@ -66,7 +66,7 @@ module.exports = function fieldOrganisationType(locale) {
                 en: oneLine`My organisation is a not-for-profit company
                     registered with Companies House, and <strong>may also</strong>
                     be registered as a charity`,
-                cy: oneLine`Mae fy sefydliad yn gwmni dielw sydd yn gofrestredig
+                cy: oneLine`Mae fy sefydliad yn gwmni di-elw sydd yn gofrestredig
```
mattandrews: Quick request from Osian so chucked it in.

controllers/apply/awards-for-all/form.test.js
```diff
@@ -329,7 +329,7 @@ describe('Project details', () => {
                     endDate: { day: 1, month: 1, year: 2030 }
                 }
             },
-            [expect.stringMatching(/Date you end the project must be within/)]
+            [expect.stringMatching(/Date you start the project must be after/)]
```
davidrapson: Not sure I follow why this has changed?

mattandrews: Discussed this in-person but the gist of it is: this bug was causing ambiguous validation rules anyway, but also because the date 1st January 2020 is now an invalid start date (as of the last day or so), so this test returns a different result than it used to. Have updated it to make the dates dynamic.

PR #2277, opened 2019-08-30
Add HotJar tag for summary

mattandrews:
Tags recordings so we can verify what people do when their form is complete

### DIFF
--- a/controllers/apply/form-router-next/summary.js
+++ b/controllers/apply/form-router-next/summary.js
@@ -76,6 +76,20 @@ module.exports = function(formBuilder) {
             }
         }
 
+        if (form.progress.isComplete) {
+            const msg = ['Apply: AFA: Summary: User shown form complete'];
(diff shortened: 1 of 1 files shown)


controllers/apply/form-router-next/summary.js
```diff
@@ -76,6 +76,18 @@ module.exports = function(formBuilder) {
             }
         }
 
+        if (form.progress.isComplete) {
+            const msg = ['Apply: AFA: Summary: User shown form complete'];
+            if (res.locals.hotJarTagList) {
```
davidrapson: I think `concat` will handle this `if` logic for you.

mattandrews: You sure? Playing about with it, it'll add `undefined` into the array if it doesn't exist?

davidrapson: Hmm. I think I was misremembering, ignore me!

PR #2279, opened 2019-08-30
Sequelize Seeding

NikhilNanjappa:
Made use of `sequelize-cli` to seed the database.

In this case, created a seed file which consumes a local json file with data and seeds the `PendingApplications` table.


.gitignore
```diff
@@ -5,6 +5,7 @@
 .idea
 .map
 .vscode/
+.sequelizerc
```
NikhilNanjappa: gitignored this as this contains db uri.

seeders/pending-applications.js
```diff
@@ -0,0 +1,29 @@
...
+  pendingApplications.data.forEach(application => {
+    application.id = uuidv4();
+    application.createdAt = moment().toDate();
+    application.updatedAt = moment().toDate();
+  });
+  pendingApplications.data[0].expiresAt = moment().add('30', 'days').toDate();
+  pendingApplications.data[1].expiresAt = moment().add('14', 'days').toDate();
+  pendingApplications.data[2].expiresAt = moment().add('2', 'days').toDate();
+  pendingApplications.data[3].expiresAt = moment().toDate();
+  pendingApplications.data[4].expiresAt = moment().subtract('5', 'days').toDate();
```
NikhilNanjappa: dynamically change the dates based on the current.

davidrapson: Id rather generate the whole thing using `faker` if we can (see form.test.js for an example) than use a static fixture

davidrapson:
(should say this is a really great start by the way 💪 )

NikhilNanjappa:
I tried using the database-config but couldn't get it working. The `process.env` values were not getting resolved (returns `undefined` without node running).

seeders/pending-applications.js
NikhilNanjappa: Sure. Thought, the prop `expiresAt` will still need adding/subtracting 'x' days as this needs to be dynamic to the current date for various scenarios .ie. seeding data for 30, 14, 2 day expiries each.

NikhilNanjappa:
Adding the `require('dotenv').config()` didn't work at the top of `database-config.js` but works fine if added in the `.sequelizerc`

db/seeders/data/pending-applications.js
```diff
@@ -0,0 +1,449 @@
...
+        line1: faker.address.streetAddress(),
+        townCity: faker.address.city(),
+        county: faker.address.county(),
+        postcode: 'B15 1TR'
+    };
+};
+
+module.exports = {
+    data: [
+        {
```
davidrapson: Lots of duplication here, is it the expiration date that is useful to change? Could we share a method to avoid duplicating all the shared code?

db/seeders/data/pending-applications.js
```diff
@@ -0,0 +1,449 @@
...
+            id: uuidv4(),
+            userId: 1,
+            formId: "awards-for-all",
+            applicationData: JSON.stringify({
```
mattandrews: Would it be possible to sort the keys alphabetically in this object, to make it a bit easier to update in future?

db/seeders/data/pending-applications.js
```diff
@@ -0,0 +1,449 @@
...
+                    endDate: toDateParts(moment().add(30, 'weeks')),
+                    startDate: toDateParts(moment().add(18, 'weeks'))
+                },
+                bankAccountNumber: faker.finance.account(),
+                projectTotalCosts: faker.random.number({ min: 301, max: 400 }),
+                seniorContactName: {
+                    lastName: faker.name.lastName(),
+                    firstName: faker.name.firstName()
+                },
+                seniorContactRole: "trustee",
```
mattandrews: This will probably be invalid based on the `organisationType` above – have a look at the logic [here](https://github.com/biglotteryfund/blf-alpha/blob/master/controllers/apply/awards-for-all/lib/roles.js#L88).

One solution would be to hardcode the org type to something that matches this role (in this case, `UNINCORPORATED_REGISTERED_CHARITY`)

mattandrews: Although not sure if this data really needs to be valid!

db/seeders/data/pending-applications.js
mattandrews: Also, could this just be a function which returns a blob of random data values each time, rather than duplicating all the field names each time?

eg. something like this:

```
const getRandomApplicationData = expiryDate => {
    return {
        projectName: faker.lorem.words(5),
        bankSortCode: faker.finance.account(6),
        ...,
        expiresAt: expiryDate
    };
};

const expiryDates = [
    moment().add('30', 'days'),
    moment().add('14', 'days'),
    moment().add('2', 'days'),
    moment(),
    moment().subtract('5', 'days')
];

module.exports = {
    data: expiryDates.map(expiryDate => {
        return {
            id: uuidv4(),
            userId: 1,
            formId: 'awards-for-all',
            applicationData: JSON.stringify(getRandomApplicationData(expiryDate.toDate()))
        };
    })
};


```

NikhilNanjappa:
This is an interesting point. I had the same doubt while thinking of creating the submitted applications seed as well. Specifically the `createFromPendingApplication` where we would need pending applications first, get the pending application id for example and use it for the submitted application.

db/seeders/data/pending-applications.js
NikhilNanjappa: Makes complete sense! @davidrapson mentioned the same as well. Will work on this change now.

db/seeders/data/pending-applications.js
```diff
@@ -0,0 +1,129 @@
...
+        totalIncomeYear: faker.random.number({ min: 1000, max: 1000000 }),
+        yourIdeaCommunity: faker.lorem.words(55),
+        yourIdeaPriorities: faker.lorem.words(55),
+        yourIdeaProject: faker.lorem.words(55)
+    };
+};
+
+const prepareSeeds = async () => {
+    const user = await Users.createUser({
+        username: faker.internet.email(),
```
davidrapson: Tweak here, this generates potentially real email addresses (ran this and got a `@gmail.com` address) for other user seeds we've used `${uuid}@example.com` for the email address.

PR #2281, opened 2019-09-02
Country tests

davidrapson:
Previously our Cypress tests only ran through a Scotland application. This PR reworks the tests to randomise the selected country during the test and adds support for country specific questions.

### DIFF
--- a/cypress/integration/integration_spec.js
+++ b/cypress/integration/integration_spec.js
@@ -410,14 +410,6 @@ describe('common pages', () => {
 
 describe('awards for all', function() {
     it('should submit full awards for all application', () => {
-        const country = 'Scotland';
-
-        function shouldDisplayErrors(errorDescriptions = []) {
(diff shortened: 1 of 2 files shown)


cypress/integration/integration_spec.js
```diff
@@ -562,51 +554,19 @@ describe('awards for all', function() {
...
-                        cy.getByLabelText('Item or activity').type(description);
-                        cy.getByLabelText('Amount').type(amount);
+                        cy.getByLabelText('Item or activity').type(
+                            `Example budget item ${index + 1}`
+                        );
+                        cy.getByLabelText('Amount').type(amount.toString());
                     });
-            }
-
-            function deleteLastRow() {
```
davidrapson: I've deleted this error code as we have coverage for this in other tests and it's the only place in these tests where we check that. Can reinstate if it's felt these have more value.

PR #2284, opened 2019-09-02
Merge locals

davidrapson:
Following on from https://github.com/biglotteryfund/blf-alpha/pull/2283 realised that the concept of "app locals"  and "request locals" for our use case are the same so we can merge a few things together.


mattandrews:
I agree with @NikhilNanjappa, makes sense to kill app locals but keeping them separate is a bit more manageable IMO.

PR #2293, opened 2019-09-03
Add additional data to validation error logging

mattandrews:
This is to help make the error logging a bit more useful:

- Adds the error type (eg. `base`, `number.max`) as well as the error message (eg. to make it easier to group types of error when the messages are dynamic)

- Adds an (anonymised) application ID to the data so we can group by these (eg. to isolate errors from one individual getting stuck on a specific field and skewing the numbers)


mattandrews:
Swapped `md5` for native `crypto`!

PR #2295, opened 2019-09-03
Remove mainContactIsValid field

davidrapson:
Based on the limited numbers of applications we've seen with possibly related contacts and revisiting approaches for this I think we are safe to remove this field.


davidrapson:
Yep think so, but let's leave this open and we can chat it through in sprint planning tomorrow.

PR #2296, opened 2019-09-04
Tweak clamav script

davidrapson:
Revised version of clamav script based on some tests on plain EC2 instances. Restarting the service after running `freshclam` seems to kick things off. Including a `service clamav-daemon status` call so we can log the status for debugging.


davidrapson:
[Install clamav on EC2 instances through provision script](https://trello.com/c/ZTFuu87m/786-install-clamav-on-ec2-instances-through-provision-script)

PR #2301, opened 2019-09-04
Remove digital fund form code

davidrapson:
The digital fund in its current incarnation has been closed for 8 months so I think we are pretty safe to remove the unused application form code now. As and when we open further rounds of applications we'll almost certainly want to revisit this approach anyway.


davidrapson:
@mattandrews Checked analytics and we haven't had a single hit to the apply urls since the rebrand so thing we are good.

PR #2302, opened 2019-09-04
Change application duration stats to use minutes for fastest application

mattandrews:
Before:

image

After:

image

(I think the changes in number comes from the greater precision involved in measuring everything in minutes then converting this into days, rather than averaging everything into days to begin with)


davidrapson:
But this is brilliant for now, the "0 days" was bugging me 💯

PR #2307, opened 2019-09-05
Install clamav in production

davidrapson:
Remove test server check from clamav script, first step in getting this change live.

### DIFF
--- a/deploy/provision.sh
+++ b/deploy/provision.sh
@@ -9,15 +9,11 @@ apt-get update
 apt-get install -y nginx-extras passenger
 
 # Install ClamAV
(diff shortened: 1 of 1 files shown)


deploy/provision.sh
```diff
@@ -9,15 +9,11 @@ apt-get update
...
-then
-    apt-get install -y clamav clamav-daemon
-    service clamav-freshclam restart
-    sleep 3m
-    service clamav-daemon start
-    service clamav-daemon status
-fi
+apt-get install -y clamav clamav-daemon
+service clamav-freshclam restart
+sleep 3m
```
mattandrews: Possibly worth adding a comment here to explain why we're doing this?

PR #2327, opened 2019-09-12
Added progress state to pending applications

NikhilNanjappa:
- Create an object in constants to hold the different states. For now has `NOT_STARTED` (default value), `PENDING` and `COMPLETE`.
- while saving the application on each step, the `currentProgressState` in db updates based on `form.progress.isComplete`.

> Note: Before we merge this we'll need to create the column manually on test/prod, so nothing errors when we deploy the change.

Screenshot 2019-09-13 at 12 44 27


db/models/applications.js
```diff
@@ -2,6 +2,7 @@
 'use strict';
 const moment = require('moment');
 const { Model, Op } = require('sequelize');
+const { APPLICATION_PROGRESS_STATES } = require('../../controllers/apply/awards-for-all/constants');
```
davidrapson: Importing this from specific router code feels like a bit of a smell. Can we define this in this file as an ENUM?

db/models/applications.js
```diff
@@ -109,9 +120,12 @@ class PendingApplication extends Model {
             expiresAt: expiresAt
         });
     }
-    static saveApplicationState(id, data) {
+    static saveApplicationState(id, data, formIsComplete) {
         return this.update(
-            { applicationData: data },
+            {
+                applicationData: data,
+                currentProgressState: formIsComplete ? APPLICATION_PROGRESS_STATES.complete : APPLICATION_PROGRESS_STATES.pending
```
davidrapson: Can we pass the actual state into here, rather than passing `formIsComplete`?

controllers/apply/awards-for-all/constants.js
```diff
@@ -109,6 +109,12 @@ const FILE_LIMITS = {
     ]
 };
 
+const APPLICATION_PROGRESS_STATES = {
```
davidrapson: I don't think this should be defined here, but rather set as a ENUM / field limit on the database.

db/models/applications.js
NikhilNanjappa: I explored around the ENUMs when I was working on the Application Expiry piece but always found inconsistencies around whether its supported or not similar to many other types like array etc. The [doc here](https://sequelize.org/master/class/lib/data-types.js~ENUM.html) says it's only compatible with postgres.

davidrapson: Oh, interesting, we are using ENUM on the survey model but I guess it just doesn't do what we think it does! Fair enough, keep it like this. Still reckon the constants should live in here rather than in the awards for all constants but otherwise looks good.

db/models/applications.js
```diff
@@ -30,6 +31,16 @@ class PendingApplication extends Model {
...
             },
 
+            /**
+             * Application's Current Progress Status
+             * e.g. PENDING
+             */
+            currentProgressState: {
+                type: DataTypes.ENUM('NOT_STARTED', 'PENDING', 'COMPLETE'),
+                allowNull: true,
+                defaultValue: 'NOT_STARTED'
```
NikhilNanjappa: Need to check if this does not replace the existing in progress applications to have `NOT_STARTED` instead of leaving it as `NULL`.

mattandrews: I think it will only take effect if we created a new application and didn't specify this field – when we add the new column manually, it'll set everything already in there to `NULL`.

db/models/applications.js
```diff
@@ -96,6 +107,11 @@ class PendingApplication extends Model {
             }
         });
     }
+    static countCompletedApplications(applications) {
```
mattandrews: I think this function should live in the application dashboard code as it's not really part of the model (eg. it's just operating on an array of application objects), and then it's in the same place as all the other statistical functions/lookups for the dashboard.

db/models/applications.js
NikhilNanjappa: Thats what I assumed but running the query `... MODIFY currentProgressState enum('NOT_STARTED', 'PENDING', 'COMPLETE') DEFAULT 'NOT_STARTED'` changed all the existing applications to `NOT_STARTED` too. My only guess is MYSQL `DEFAULT` is not the same as Sequelize `defaultValue` ?

db/models/applications.js
NikhilNanjappa: :) It was @davidrapson 's suggestion. David - you want to take this one ?

db/models/applications.js
mattandrews: Ah okay, maybe when we manually create the column we'll have to set it to `DEFAULT NULL` then immediately change the default (thoughts @davidrapson)?

db/models/applications.js
NikhilNanjappa: I do agree though, that it need not be part of the model as its essentially just a 'count' and can be directly used like `uniqBy` or `getAppsToday` methods on the previous lines without the need for a model/db call.

davidrapson: I do think that _something_ like this should be on the model, i.e. the notion of asking for some detail of the application status makes sense on the model to avoid leaking implementation details. Maybe an instance method like:

```js
isComplete() {
   return this.currentProgressState === 'COMPLETE'
}
```

Might be preferable so that we can then write something like

```js
const completedCount = applications.filter(_ => _.isComplete()).length;
```

Having a helper method helps us avoid needing to expose the implementation details of what ENUM value is used.

mattandrews: Yeah that makes sense, but keep the actual filtering/counting code with all the rest in applications.js!

db/models/applications.js
davidrapson: I think default of `null` is OK, and we could have the new applications method explicitly set NOT_STARTED?

db/models/applications.js
```diff
@@ -96,6 +106,9 @@ class PendingApplication extends Model {
             }
         });
     }
+    static countCompleted(applications) {
+        return applications.filter(app => app.currentProgressState === 'COMPLETE').length;
```
NikhilNanjappa: replaced lodash `countBy` with JS `filter` method as `countBy` stopped working after I change the type of `currentProgressState` to `ENUM` (flaky + weird?).

Should we stay away from ENUM or go ahead with this ?

mattandrews: Hmm weird – it's probably fine to leave this as you've got it here, then 👍

NikhilNanjappa:
Before we merge, I guess we need to create the column in test/prod ?

PR #2337, opened 2019-09-16
Joi upgrade alias

davidrapson:
Whilst doing a proof of concept upgrade in https://github.com/biglotteryfund/blf-alpha/pull/2336 it highlighted that there are a _lot_ of changes in joi 16 which means pretty much all the schema code needs upgrading.

It turns out npm 6.9+ supports package aliases so we can run multiple versions of joi at once to aid migration. We would still want to do the upgrade quickly but this might avoid a long running branch or 1000+ line change set in one go. As a proof of concept migrated just the very simple assistance schema to the new version.

We should do this _soon_ but may be worth waiting a week as the documentation is still a little fresh (full docs on the new extension syntax are still forthcoming although there are some basic guides and error messages help you out).

Thoughts very welcome 💭

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   },
   "dependencies": {
     "@hapi/joi": "15.1.1",
+    "@hapi/joi16": "npm:@hapi/joi@16.0.1",
     "@ideal-postcodes/core-node": "1.1.0",
     "@sentry/browser": "5.6.2",
     "@sentry/integrations": "5.6.1",
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   },
   "dependencies": {
     "@hapi/joi": "15.1.1",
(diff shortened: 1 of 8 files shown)


davidrapson:
Just used the same approach to migrate the feedback and survey API schemas and feels manageable using two versions for a (short) period. Thinking we'd start with these simpler ones then:

- Account schema
- Standard form schema (accepting some duplication of extensions to begin with)
- Simple form schema
- Re-merge extensions (if needed)
- Delete old joi version

package.json
```diff
@@ -70,6 +70,7 @@
   },
   "dependencies": {
     "@hapi/joi": "15.1.1",
+    "@hapi/joi16": "npm:@hapi/joi@16.1.1",
```
NikhilNanjappa: This flexibility is very useful! 👍

PR #2338, opened 2019-09-16
Enable noValidate in production

davidrapson:
Confirmed we can deploy this to production now, so enabling this in production and removing the temporary feature flag.


davidrapson:
[Disable default in-browser validation to increase consistency, improve accessibility, and resolve translation limitations](https://trello.com/c/LE8dx7jB/753-disable-default-in-browser-validation-to-increase-consistency-improve-accessibility-and-resolve-translation-limitations)

PR #2349, opened 2019-09-18
Standard form: Initial project country step

davidrapson:
This is the first step towards the get advice form. I've added the initial form schema and the first step within this PR. You will not be able to get all the way through the form at this point, but at least gives us the very first step:

image

I've left the commits very granular to show progress over time, will squash things down when merging. Will call out larger changes in comments.


controllers/apply/awards-for-all/form.js
```diff
@@ -9,9 +9,9 @@ const has = require('lodash/fp/has');
 const sumBy = require('lodash/sumBy');
 const { safeHtml, oneLine } = require('common-tags');
 
-const { FormModel } = require('../form-router-next/lib/form-model');
+const { FormModel } = require('../lib/form-model');
 const fromDateParts = require('../lib/from-date-parts');
-const { formatDateRange } = require('../form-router-next/lib/formatters');
+const { formatDateRange } = require('../lib/formatters');
```
davidrapson: Lifted code that is shared between the two form models into a shared `lib` directory.

controllers/apply/form-router-next/index.js
```diff
@@ -172,17 +172,23 @@ function initFormRouter({
...
+        );
+    }
 
     /**
      * Route: Start application
      * Redirect to eligibility checker
      */
     router.get('/start', function(req, res) {
-        res.redirect(`${req.baseUrl}/eligibility/1`);
+        if (eligibilityBuilder) {
```
davidrapson: The new form doesn't have an eligibility checker so making this optional and start a new application if we don't have a checker step.

controllers/apply/lib/field-types.js
```diff
@@ -0,0 +1,153 @@
+'use strict';
+const get = require('lodash/fp/get');
+const { oneLine } = require('common-tags');
+const Joi = require('./joi-extensions');
+
+class TextField {
```
davidrapson: This is the biggest change. Adds the concept of field types to give us a way of sharing common field logic. Aim is to make the standard form use these and then update the awards for all form to start using them once we've confirmed it works as an approach.

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+
+module.exports = function({ locale = 'en', data = {} } = {}) {
+    const localise = get(locale);
+
+    const allFields = {
+        projectCountry: new RadioField({
+            name: 'projectCountry',
+            label: localise({
+                en: `What country (or countries) will your project take place in?`,
+                cy: ``
```
mattandrews: Just thinking ahead for future-proofing purposes here – do you think it's worth adding a specific TODO string to these missing translations to make it easier to locate them all when it comes to translation time (rather than my gnarly regex from before)?

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+                        en: 'Select a location',
+                        cy: 'Dewiswch leoliad'
+                    })
+                }
+            ]
+        },
+        projectLocationDescription: new TextField({
+            label: localise({
+                en: 'Project location',
+                cy: ''
```
mattandrews: This is `Lleoliad y prosiect`

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+                    type: 'number.integer',
+                    message: localise({
+                        en: `Total cost must be a whole number (eg. no decimal point)`,
+                        cy: `Rhaid i’r cost fod yn rif cyflawn (e.e. dim pwynt degol)`
+                    })
+                },
+                {
+                    type: 'number.min',
+                    message: localise({
+                        en: `Must be at least £10,000`,
```
mattandrews: In AFA we've made this a constant (`MAX_BUDGET_TOTAL_GBP`) and used that in error messages, despite the extremely unlikely odds of it ever changing. Worth reusing here (or indeed adding a link to AFA for this message)?

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+                        cy: ''
+                    })
+                }
+            ]
+        },
+        projectIdea: new TextField({
+            name: 'projectIdea',
+            label: localise({
+                en: 'What would you like to do?',
+                cy: ''
```
mattandrews: `Beth yr hoffech ei wneud?`

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+                        cy: `Rhaid i’r ateb fod yn llai na 500 gair`
+                    })
+                }
+            ]
+        }),
+        organisationLegalName: new TextField({
+            name: 'organisationLegalName',
+            label: localise({
+                en: 'What is the full legal name of your organisation?',
+                cy: ''
```
mattandrews: `Beth yw enw cyfreithiol llawn eich sefydliad?`

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+            }),
+            explanation: localise({
+                en: oneLine`This must be as shown on your governing document.
+                    Your governing document could be called one of several things,
+                    depending on the type of organisation you're applying
+                    on behalf of. It may be called a constitution, trust deed,
+                    memorandum and articles of association,
+                    or something else entirely.`,
+                cy: ``
```
mattandrews: We have this in the translations too – I think it's worth getting them all in now where they exist, unless we know these labels will change (in which case sorry, ignore me!)

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+                    type: 'string.minWords',
+                    message: localise({
+                        en: `Answer must be at least 50 words`,
+                        cy: `Rhaid i’r ateb fod yn o leiaf 50 gair`
+                    })
+                },
+                {
+                    type: 'string.maxWords',
+                    message: localise({
+                        en: `Answer must be no more than 500 words`,
```
mattandrews: Would be nice to see these as `settings` variables to make it easier to share the schema and the messages.

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+                        cy: 'Rhowch gyfenw'
+                    })
+                }
+            ]
+        },
+        contactEmail: new EmailField({
+            name: 'contactEmail',
+            label: localise({
+                en: 'Email',
+                cy: ``
```
mattandrews: `E-bost`

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+            })
+        }),
+        contactPhone: new PhoneField({
+            name: 'contactPhone',
+            label: localise({
+                en: `Telephone number`,
+                cy: ``
```
mattandrews: `Rhif ffôn`

controllers/apply/get-advice/form.js
```diff
@@ -0,0 +1,398 @@
...
+            })
+        })
+    };
+
+    function stepProjectCountry() {
+        return {
+            title: localise({ en: 'Project country', cy: '' }),
```
mattandrews: `Gwlad y prosiect`

controllers/apply/lib/field-types.js
```diff
@@ -0,0 +1,153 @@
...
+        this.settings = props.settings || {};
+    }
+    localise(msg) {
+        return get(this.locale)(msg);
+    }
+}
+
+class EmailField extends TextField {
+    constructor(props, locale) {
+        super(props, locale);
```
mattandrews: 💯

controllers/apply/get-advice/form.js
davidrapson: I think it's OK, the number of files we need to translate will be smaller here (should mostly just be the form/field model)

controllers/apply/get-advice/form.js
davidrapson: I have a separate card open for this, so yes will link across to awards for all here.

controllers/apply/get-advice/form.js
davidrapson: Started doing that in a follow on PR so 👍

davidrapson:
Ignoring the comments here solely because I've started picking most of this up in a follow on PR. Going to merge this as is.

PR #2354, opened 2019-09-19
Update org type question

mattandrews:
Update org type question to reflect latest prototype, eg. switch the order of some options, make the copy briefer, and add the new CIC option.


controllers/apply/awards-for-all/constants.js
```diff
@@ -43,6 +43,7 @@ const ORGANISATION_TYPES = {
     UNINCORPORATED_REGISTERED_CHARITY: 'unincorporated-registered-charity',
     CIO: 'charitable-incorporated-organisation',
     NOT_FOR_PROFIT_COMPANY: 'not-for-profit-company',
+    CIC: 'community-interest-company',
```
mattandrews: Need to confirm this with Hyphen8.

PR #2370, opened 2019-09-24
Migrating Joi to v16: User schema

NikhilNanjappa:
- migrated password-strength extension code to the new API format.

Not a lot of changes but needs going through the new documentation/issues/internal codes to make a change.


NikhilNanjappa:
Not normalize errors to be precise (at least not yet) but the return value of `error` was changed from `null` to `undefined` if no errors.

PR #2371, opened 2019-09-24
Application expiry v2

mattandrews:
Builds on @NikhilNanjappa's great work to add the concept of an application email queue.

- When new applications are created, if the form router is passed an object of `expiryEmailPeriods` (see `constants.js`), an array of email queue items are created
- An API endpoint, when POSTed to, will look up all emails scheduled to be sent in the past, send them, and mark them as sent
- The same endpoint deletes any expired applications
- The Cypress tests make some expired applications and then verifies the email queue contains what we expect, emails them, then verifies the queue is clear

The only part that's a bit clunky is the email templating: right now it's hardcoded to AFA, but it'll be simple enough to vary on the `emailType` field in the database to change to different templates/messages in the future.

There's also some gnarly code for table relations to allow joining between email queues / applications / users. I spent a day and a half finessing this and I'm fairly sure it's right now, but Sequelize makes this stuff something of a mystery.

Feedback encouraged as there's a fair amount of complexity / key functionality here.


db/models/applications.js
```diff
@@ -97,6 +100,23 @@ class PendingApplication extends Model {
             order: [['createdAt', 'DESC']]
         });
     }
+    static findExpiredApplications() {
+        return this.findAll({
+            attributes: ['id'],
+            where: {
+                expiresAt: {
+                    [Op.lte]: moment().toDate()
```
NikhilNanjappa: Interesting! I thought Sequelize 5.x and above, we couldn't use `lte` like operators on dates.

db/models/index.js
```diff
@@ -37,9 +41,28 @@ const db = {
...
+    constraints: false // don't delete users when deleting applications
+});
+db.Users.hasMany(db.PendingApplication);
+
+db.SubmittedApplication.belongsTo(db.Users);
+db.Users.hasMany(db.SubmittedApplication);
+
+db.ApplicationEmailQueue.belongsTo(db.PendingApplication, {
+    foreignKey: 'applicationId',
+    constraints: false
```
NikhilNanjappa: Could we use `cascade` to delete emailQueue records when the appropriate pending applications are deleted ?

controllers/api/lib/application-expiry.js
```diff
@@ -0,0 +1,124 @@
...
+ */
+const deleteExpiredApplications = async expiredApplications => {
+    try {
+        logger.info('Handling expired applications');
+        const applicationIds = expiredApplications.map(
+            application => application.id
+        );
+
+        const dbStatus = await PendingApplication.deleteApplications(
+            applicationIds
```
NikhilNanjappa: Guess we will also have to delete the emailQueue records as they serve no purpose without the pending application (or does the relation take care of it ?).

db/models/applications.js
mattandrews: It's working for me! Where did you see that?

db/models/index.js
mattandrews: We could but I'm a bit wary of it as it's a bit magic and I'm scared of getting the associations wrong (the docs are fairly unclear about this) and accidentally deleting a user's account as well when we delete email queue items (or pending applications). For now I've just manually deleted them which I think is more explicit.

controllers/api/lib/application-expiry.js
mattandrews: We can keep them around for a while, eg. 3 months after they're sent – it helps for auditing purposes, eg. if someone complains that we deleted their application without warning them, we can look up the record to verify that we did email them.

controllers/apply/awards-for-all/confirmation.test.js
```diff
@@ -11,8 +15,8 @@ describe('confirmation builder', () => {
         });
 
         expect(england.body).toMatchSnapshot();
-        expect(england.body).toContain('0345 4 10 20 30');
-        expect(england.body).toContain('afe@tnlcommunityfund.org.uk');
+        expect(england.body).toContain(CONTACT_DETAILS_PHONE.england);
```
davidrapson: I'd avoid importing the constants for this test, it's good if unit tests avoid importing implementation details of the code they are testing. We could use a snapshot test here if we want to avoid duplicating values. (I know we do this elsewhere but it's good to avoid)

controllers/apply/awards-for-all/constants.js
```diff
@@ -109,6 +109,30 @@ const FILE_LIMITS = {
     ]
 };
 
+const EXPIRY_EMAIL_REMINDERS = [
+    {
+        emailTypeKey: 'AFA_ONE_MONTH',
```
davidrapson: type key feels a bit overkill, perhaps `emailKey` or `emailType`

controllers/apply/form-router-next/delete.js
```diff
@@ -2,7 +2,10 @@
 const express = require('express');
 const path = require('path');
 
-const { PendingApplication } = require('../../../db/models');
+const {
+    PendingApplication,
+    ApplicationEmailQueue
```
davidrapson: ApplicationEmailQueue doesn't look like its used here?

controllers/apply/form-router-next/index.js
```diff
@@ -127,6 +130,30 @@ function initFormRouter({
...
+     */
+    if (features.enableSeeders) {
+        router.post('/seed', async (req, res) => {
+            const application = await PendingApplication.createNewApplication({
+                formId: formId,
+                userId: req.body.userId,
+                customExpiry: req.body.expiresAt
+            });
+
+            if (expiryEmailPeriods) {
```
davidrapson: Should this code be lifted up into `createNewApplication`? Do we always create queue items when creating an application?

controllers/apply/form-router-next/lib/application-expiry.js
```diff
@@ -0,0 +1,179 @@
...
+const { sendHtmlEmail } = require('../../../../common/mail');
+const {
+    PendingApplication,
+    ApplicationEmailQueue
+} = require('../../../../db/models');
+
+const { getEmailFor, getPhoneFor } = require('../../../../common/contacts');
+
+const {
+    signTokenUnsubscribeApplicationEmails
```
davidrapson: Can we inline this code? Having a method called `signTokenUnsubscribeApplicationEmails` in common feels too specific, is it only related to this logic?

controllers/apply/form-router-next/lib/application-expiry.js
```diff
@@ -0,0 +1,179 @@
...
+        throw new Error(
+            'Missing environment variable APPLICATION_EXPIRY_EMAIL'
+        );
+    }
+
+    return await Promise.all(
+        emailQueue.map(async emailToSend => {
+            let subjectLine = '';
+
+            switch (emailToSend.emailType) {
```
davidrapson: 💭 Wondering if the `emailQueue` should also contain the subject line / email content? I think this is fine for now though.

db/models/index.js
NikhilNanjappa: Yeah I agree with the docs being unclear. Makes sense for it to explicit!

db/models/applications.js
NikhilNanjappa: I don't have the exact link but I was sure I read it on StackOverflow. Anyways as long as its working, its fine I guess!

common/contacts.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
```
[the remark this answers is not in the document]
mattandrews: Talked about this and the emails/phone numbers are generic (not AFA-specific) so leaving this where it is.

controllers/apply/form-router-next/index.js
mattandrews: This code will be deleted shortly so for now let's leave this as a separate explicit thing (and see if the Standard form will need this functionality, in which case we'll know if there's a parallel between application/queue creation).

mattandrews:
@davidrapson could you re-review with the new changes?

controllers/apply/form-router-next/views/unsubscribed.njk
```diff
@@ -0,0 +1,14 @@
+{% extends "layouts/main.njk" %}
+
+{% set bodyClass = 'has-static-header' %}
+{% block content %}
+    <main role="main" id="content">
+        <div class="content-box u-inner-wide-only">
+            <div class="s-prose">
+                <h1>{{ title }}</h1>
+                <div class="message" role="alert">{{ message | safe }}</div>
+                <p><a class="btn" href="/">{{ __('global.misc.back') | capitalize }}</a></p>
```
NikhilNanjappa: I think providing an option to "close this tab" would be good too. 💭

mattandrews:
Re-requesting a review for the most recent commits – I've tweaked it to send an email to a known secret (the digital monitoring address) even when the feature is switched off, so we can get this merged and start running it to verify it emails as we'd expect.

PR #2372, opened 2019-09-25
Organisation type field

davidrapson:
image


controllers/apply/get-advice/form.js
```diff
@@ -412,6 +412,137 @@ module.exports = function({ locale = 'en', data = {} } = {}) {
...
+    function fieldOrganisationType() {
+        return new RadioField({
+            name: 'organisationType',
+            label: localise({
+                en: 'What type of organisation are you?',
+                cy: 'Pa fath o sefydliad ydych chi?'
+            }),
+            options: [
+                {
+                    value: 'unregistered-vco',
```
davidrapson: I'm deliberately not using any constants here or sharing the code from awards for all a) we don't need to depend on organisation type for sub-questions here and b) the actual list may diverge from the simple product (e.g. different options based on country, and some different valid types) so doesn't make sense to share anything.

davidrapson:
[Ask for organisation type](https://trello.com/c/dPYdCUfw/827-ask-for-organisation-type)

controllers/apply/lib/field-types/field.js
```diff
@@ -19,34 +17,77 @@ class Field {
...
+     */
+    defaultLabel() {
+        return null;
+    }
+
+    /**
+     * Allow sub-classes to provide a default schema
+     * also allows it to be accessed by the instance
+     */
+    defaultSchema() {
```
davidrapson: Added these `default*` methods to make it easier for the sub-types to define their own without having to duplicate as much boilerplate in the constructor. See `phone.js` for a good example of this.

controllers/apply/lib/field-types/field.js
```diff
@@ -19,34 +17,73 @@ class Field {
             throw new Error('Must provide name');
         }
 
-        this.label = props.label;
+        const label = props.label ? props.label : this.defaultLabel();
+        if (label) {
+            this.label = label;
+        } else {
+            // throw new Error('Must provide label');
```
mattandrews: Should this be here/commented?

PR #2379, opened 2019-09-26
Standard form start end

davidrapson:
Adds basic start page and confirmation page, plus some temporary workarounds to get submissions working (bypassing salesforce and database saving for get advice form for now)

localhost_3000_apply_get-advice_start


controllers/apply/form-router-next/submission.js
```diff
@@ -0,0 +1,181 @@
...
+                    form: form
+                });
+            });
+        }
+
+        /**
+         * Fake submission for get advice form
+         * @TODO: SubmittedApplication assumes project title
+         */
+        if (formId === 'get-advice' && appData.isNotProduction) {
```
davidrapson: Currently the submitted application model requires an application so needs reworking. For now I've bypassed _all_ saving for the get advice form.

PR #2383, opened 2019-09-27
Add related content to listing pages

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/192

Allows listing pages to have a grid of related stories like the homepage:

image


mattandrews:
This changed a fair bit after I reworked the Craft end, so give it another look @davidrapson!

views/components/related-content/macro.njk
```diff
@@ -1,19 +1,34 @@
...
+                {% for article in articles %}
                     <li class="flex-grid__item related-content__item">
-                        {{ articleTeaser({
-                            'title': article.trailText or article.title,
-                            'linkUrl': article.linkUrl,
-                            'summary': article.summary
-                        }) }}
+                        {% if usePromo %}
+                            {{ promoCard({
+                                "title": article.title,
```
davidrapson: should `title` not also be `article.trailText or article.title`. Any rather than the `usePromo` flag could we check if we have an image to determine which style to use?

mattandrews: The new field doesn't have a `trailText` attribute (the component itself at the CMS end determines whether to use the linked entry's title or an overridden one).

I did think about checking for an image but then it'd be a style mismatch as the promo card looks like this:

image

... but the regular version like this:

image

So I think all three need to look the same. Ultimately we'll probably want to switch the above/regular version to be a promo card I guess, if we start adding images to the homepage news grid.

PR #2384, opened 2019-09-30
Standard form refinement

davidrapson:
Adds statutory body sub type to get advice field

image

Does some work to share location options better between forms and extracts country and location fields into new files for awards for all form.


davidrapson:
[Show sub-type for statutory bodies](https://trello.com/c/UAYg5K2E/843-show-sub-type-for-statutory-bodies)

controllers/apply/get-advice/fields.js
```diff
@@ -19,16 +19,16 @@ const {
...
     const localise = get(locale);
 
-    const projectCountries = getOr([], 'projectCountry')(data);
+    const projectCountries = getOr([], 'projectCountries')(data);
 
-    function fieldProjectCountry() {
+    function fieldProjectCountries() {
         return new CheckboxField({
-            name: 'projectCountry',
+            name: 'projectCountries',
```
davidrapson: Renamed this to `projectCountries` to reflect that multiple options can be selected and this is an array

PR #2385, opened 2019-09-30
England launch

davidrapson:
- Remove form messaging. Northern Ireland do not use the old online form so we can remove all messaging with limited impact.
- Add `england` to `allowedCountries`


mattandrews:
Is it not slightly confusing for NI customers, still, if they end up here and don't know where to go?

PR #2387, opened 2019-09-30
Added last succesful save time stamp

NikhilNanjappa:
Screenshot 2019-09-30 at 15 32 51
Screenshot 2019-09-30 at 15 33 18
Screenshot 2019-09-30 at 15 34 17
Screenshot 2019-09-30 at 15 34 34
Screenshot 2019-09-30 at 15 34 46


assets/sass/components/_forms.scss
```diff
@@ -904,3 +904,23 @@ $inputOutline: 3px solid get-color('border', 'outline');
         }
     }
 }
+
+.js-form-warn-unsaved.js-session-form {
```
davidrapson: Ooh. we need to re think this CSS. Best to avoid handing any styling off a `js-` prefixed class, and this is far too specific. I'd add a modifier to form actions and make the `last-save-timestamp` a child element. The text based on the mockup should be link blue, and only the timestamp emboldened. I'd suggest something like this:

```scss
.form-actions--flex {
    display: flex;
    flex-flow: row wrap;
    justify-content: space-between;
    align-items: center;

    .form-actions__primary,
    .form-actions__secondary {
        flex: 0 0 auto;
    }

    .form-actions__timestamp {
        margin-bottom: 0;
        color: get-color('links', 'base');
        font-size: 15px;

        time {
            font-weight: font-weight('body', 'semibold');
        }
    }
}
```

With the following markup:

```
<div class="form-actions form-actions--flex">
    <div class="form-actions__primary">
        <input class="btn" type="submit" value="{{ copy.continue }}"/>
    </div>
    <p class="form-actions__secondary form-actions__timestamp">
        Last successful save - <time datetime="{{ lastSaveDatetime }}">{{ lastSaveTime }}</time>
    </p>
</div>
```

controllers/apply/form-router-next/steps.js
```diff
@@ -98,6 +106,17 @@ module.exports = function(formId, formBuilder) {
         };
     }
 
+    async function calcLastSaveTime(applicationId) {
+        const lastUpdatedAt = (await PendingApplication.lastUpdatedTime(applicationId)).updatedAt;
+
+        return moment(lastUpdatedAt).calendar(null, {
```
davidrapson: This date formatting is unusual, particularly the `L [in MM/DD/YYYY format]` line. I'd high recommend trying something like `moment(application.updatedAt).locale(locale).fromNow()` which will do all the hard work for you on formatting a relative timestamp https://momentjs.com/docs/#/displaying/fromnow/

db/models/applications.js
```diff
@@ -131,6 +131,14 @@ class PendingApplication extends Model {
             { where: { id: { [Op.eq]: id } } }
         );
     }
+    static lastUpdatedTime(applicationId) {
```
davidrapson: Is this necessary? Could we use the regular find by primary key method?

assets/sass/components/_forms.scss
davidrapson: `lastSaveDatetime` would be the ISO datetime value

davidrapson:
[Last saved timestamp added to A4A form.](https://trello.com/c/xDPpiFoh/842-last-saved-timestamp-added-to-a4a-form)

db/models/applications.js
NikhilNanjappa: I just wanted to fetch the field value we needed rather than the entire record from db, especially when it contains a fairly big json object which we don't need. Was just thinking of a minor performance win.

PR #2393, opened 2019-09-30
Standard form tweaks

davidrapson:
- Makes the start button text configurable on the form model
- Adds some non-lorem-ipsum placeholder text for section summaries

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1561,8 +1561,7 @@ applyNext:
       cta: Start new application
   summary:
     title: Summary
-    introduction: >
-      <p>Here's a summary of your application. You may complete the sections in any order.</p>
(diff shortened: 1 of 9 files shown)


config/locales/en.yml
```diff
@@ -1561,8 +1561,7 @@ applyNext:
       cta: Start new application
   summary:
     title: Summary
-    introduction: >
-      <p>Here's a summary of your application. You may complete the sections in any order.</p>
```
davidrapson: Removing "Here's a summary of your application" here as it's the only text on this page that specifies this is an application and we are keen to test peoples assumptions around that without leading them. Means it will change on the awards for all form but felt acceptable as it basically repeats the title, it's already clear this is a summary.

PR #2401, opened 2019-10-01
Fix CMS preview

mattandrews:
This uses the new Craft 3 preview tokens by passing them through on all API calls (where provided) which automagically allows for shareable draft URLs and live-updating previews.


common/content-api.js
```diff
@@ -47,15 +48,14 @@ function fetchAllLocales(toUrlPathFn, options = {}) {
...
-    let previewParams = {};
-    previewParams[previewMode.mode] = previewMode.id; // eg. ?draft=123
-
-    return Object.assign({}, previewParams, params);
+function addPreviewParams(requestParams = {}, params = {}) {
+    const globalParams = pick(requestParams, [
+        'social',
+        'x-craft-live-preview',
+        'x-craft-preview',
+        'token'
```
mattandrews: I couldn't put the locale in here as it's not a query param, but if we passed around the whole `req` object I guess we could? But this seems complex as we do some unusual things with locales for some endpoints (eg. fetching both or forcing one language etc).

common/content-api.js
```diff
@@ -234,35 +234,30 @@ function getRecentFundingProgrammes({ locale, limit = 3 }) {
...
     slug = null,
-    previewMode = null,
     query = {},
+    requestParams = {},
     type = null
 }) {
     if (slug) {
         return fetch(`/v1/${locale}/research/${slug}`, {
-            qs: addPreviewParams(previewMode, { ...query })
+            qs: addPreviewParams(requestParams, { ...query })
```
davidrapson: Merging `requestParams` with `query` is a bit redundant as `query` will be a superset of `requestParams`. Not a big deal for now but we should come back and clean this up.

server.js
```diff
@@ -183,12 +184,17 @@ app.use([
             }
         }
     }),
+    (req, res, next) => {
```
davidrapson: We don't need this extra middleware, it can go into the preview middleware assuming we call the CSP stuff first.

```js
module.exports = (req, res, next) => {
    const isContentPreview =
        req.query.token &&
        (req.query['x-craft-live-preview'] || req.query['x-craft-preview']);

    if (isContentPreview) {
        res.locals.isContentPreview = true;
        res.cacheControl = { noStore: true };
        res.removeHeader('X-Frame-Options');
        requireStaffAuth(req, res, next);
    } else {
        next();
    }
}
```

davidrapson: Thinking about it we might even want to use CSPs `frame-ancestors` in favour of this https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors as `X-Frame-Options` is largely deprecated.

common/content-api.js
mattandrews: Hmm true, in this instance the only thing being passed through is the `req.query.social` param, which is now "global" via these changes – meaning I can remove it everywhere using these calls.

server.js
mattandrews: Might leave it as-is for now as we get into slightly fiddly config territory where technically the live website might be loaded into the test CMS for previews but its iframe header won't allow it (unless we configured the header to allow all CMS domains for all app environments).

PR #2409, opened 2019-10-02
Add preview querystrings to Cloudfront

mattandrews:
Allow Cloudfront to vary on these querystrings.

### DIFF
--- a/common/cloudfront.js
+++ b/common/cloudfront.js
@@ -30,8 +30,6 @@ function makeBehaviourItem({
         : ['HEAD', 'GET'];
 
     const globalQuerystrings = [
-        'draft',
-        'version',
         'social',
         'token',
         'x-craft-live-preview',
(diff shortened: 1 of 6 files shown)


common/cloudfront.js
```diff
@@ -29,7 +29,14 @@ function makeBehaviourItem({
         ? ['HEAD', 'DELETE', 'POST', 'GET', 'OPTIONS', 'PUT', 'PATCH']
         : ['HEAD', 'GET'];
 
-    const globalQuerystrings = ['draft', 'version', 'social'];
+    const globalQuerystrings = [
```
davidrapson: We can presumably remove `draft` and `version` now, or do they still exist?

mattandrews: Oh good call – yeah, let's do that.

davidrapson:
👍

PR #2416, opened 2019-10-04
Fix postcode lookup bug

mattandrews:
Fixes this bug:
postcode-bug-v2

(eg. entering a postcode but not looking it up, and then using manual entry)

The `alert()` logic didn't account for this scenario and it sounds like it crops up when browser autofill does something unexpected.


mattandrews:
I can't see an easy way to mount a test component to an existing `<form>` element (required for these tests) so will return to this, but will also merge this as the fix is important to get live.

PR #2420, opened 2019-10-04
Save and redirect without rendering validation errors

NikhilNanjappa:
- Next/prev changed to link styled submit buttons.
- On click of next/prev buttons, the form is saved (with errors) and renders the next/prev page.

### DIFF
--- a/assets/sass/components/_buttons.scss
+++ b/assets/sass/components/_buttons.scss
@@ -221,7 +221,7 @@ $buttonColor: get-color('brand', 'primary');
     text-decoration: underline;
     border: none;
     border-radius: 0;
-    cursor: inherit;
+    cursor: pointer;
     text-align: inherit;
     word-wrap: break-word;
     white-space: normal;
--- a/assets/sass/components/_buttons.scss
+++ b/assets/sass/components/_buttons.scss
@@ -221,7 +221,7 @@ $buttonColor: get-color('brand', 'primary');
     text-decoration: underline;
     border: none;
     border-radius: 0;
-    cursor: pointer;
+    cursor: inherit;
     text-align: inherit;
     word-wrap: break-word;
     white-space: normal;
--- a/controllers/apply/form-router-next/steps.js
+++ b/controllers/apply/form-router-next/steps.js
@@ -172,6 +172,23 @@ module.exports = function(formId, formBuilder) {
                 stepFields.map(f => f.name).includes(item.param)
             );
 
+            function prevOrNextBtnNotClicked() {
+                return !(req.body.previousBtn || req.body.nextBtn);
+            }
+
+            function determineRedirectUrl() {
+                const { previousPage, nextPage } = form.pagination({
+                    baseUrl: res.locals.formBaseUrl,
+                    sectionSlug: req.params.section,
+                    currentStepIndex: stepIndex,
+                    copy: res.locals.copy
+                });
+            
+                return req.body.previousBtn
+                    ? previousPage.url
+                    : nextPage.url;
+            }
+
             try {
                 let dataToStore = validationResult.value;
                 const currentProgressState = form.progress.isComplete
@@ -203,11 +220,11 @@ module.exports = function(formId, formBuilder) {
                 );
 
                 /**
-                 * If there are errors re-render the step with errors
+                 * If there are errors re-render the step with errors only if prev/next btn is not clicked
                  * - Pass the full data object from validationResult to the view. Including invalid values.
                  * Otherwise, find the next suitable step and redirect there.
                  */
-                if (errorsForStep.length > 0 && !(req.body.previousBtn || req.body.nextBtn)) {
+                if (errorsForStep.length > 0 && prevOrNextBtnNotClicked()) {
                     const renderStep = renderStepFor(
                         req.params.section,
                         req.params.step
@@ -237,24 +254,7 @@ module.exports = function(formId, formBuilder) {
                         }
                     }
 
-                    let redirectUrl;
-                    if (req.body.previousBtn) {
-                        const { previousPage } = form.pagination({
-                            baseUrl: res.locals.formBaseUrl,
-                            sectionSlug: req.params.section,
-                            currentStepIndex: stepIndex,
-                            copy: res.locals.copy
-                        });
-                        redirectUrl = previousPage.url;
-                    } else {
-                        const { nextPage } = form.pagination({
-                            baseUrl: res.locals.formBaseUrl,
-                            sectionSlug: req.params.section,
-                            currentStepIndex: stepIndex,
-                            copy: res.locals.copy
-                        });
-                        redirectUrl = nextPage.url;
-                    }
+                    const redirectUrl = determineRedirectUrl();
 
                     /**
                      * Handle file uploads if we have any for the step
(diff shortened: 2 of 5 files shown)


controllers/apply/form-router-next/steps.js
```diff
@@ -207,7 +207,7 @@ module.exports = function(formId, formBuilder) {
                  * - Pass the full data object from validationResult to the view. Including invalid values.
                  * Otherwise, find the next suitable step and redirect there.
                  */
-                if (errorsForStep.length > 0) {
+                if (errorsForStep.length > 0 && !(req.body.previousBtn || req.body.nextBtn)) {
```
davidrapson: I reckon we need to name this conditional, mind making a function something like `shouldRenderErrors` or `shouldProceedWithoutValidation`?

controllers/apply/form-router-next/steps.js
```diff
@@ -237,12 +237,24 @@ module.exports = function(formId, formBuilder) {
...
 
-                    const { nextPage } = form.pagination({
-                        baseUrl: res.locals.formBaseUrl,
-                        sectionSlug: req.params.section,
-                        currentStepIndex: stepIndex,
-                        copy: res.locals.copy
-                    });
+                    let redirectUrl;
+                    if (req.body.previousBtn) {
+                        const { previousPage } = form.pagination({
```
davidrapson: We don't need to call this twice, you can get these values from the same call. Also worth extracting into a named function like

```js
function determineRedirectUrl() {
    const { previousPage, nextPage } = form.pagination({
        baseUrl: res.locals.formBaseUrl,
        sectionSlug: req.params.section,
        currentStepIndex: stepIndex,
        copy: res.locals.copy
    });

    return req.body.previousBtn
        ? previousPage.url
        : nextPage.url;
}
```

views/components/user-navigation/macro.njk
```diff
@@ -13,26 +13,12 @@
...
                             <span class="user-nav__pagination-arrow">←</span>
-                            <a href="{{ previousPage.url }}">
-                                <span>
-                                    {{ __('applyNext.navigation.previous') }}
-                                </span>
-                                <span class="user-nav__pagination-subtitle">
-                                    {{ previousPage.label }}
-                                </span>
-                            </a>
+                            <input class="btn-link" name="previousBtn" type="submit" value="{{ __('applyNext.navigation.previous') }}"/>
```
davidrapson: Not something for now, but I'd like to revisit how this macro is structured as it currently does two things: form pagination and account links. The account link functionality could be pulled out. Currently this macro is used on pages that aren't forms so has location dependent behaviour which I'd like to fix. Your change here looks good though 👍

views/components/user-navigation/macro.njk
```diff
@@ -13,26 +13,12 @@
...
                         <li class="user-nav__pagination-link user-nav__pagination-link--next">
-                            <a href="{{ nextPage.url }}">
-                                <span>
-                                    {{ __('applyNext.navigation.next') }}
-                                </span>
-                                <span class="user-nav__pagination-subtitle">
-                                    {{ nextPage.label }}
-                                </span>
-                            </a>
+                            <input class="btn-link" name="nextBtn" type="submit" formaction="" value="{{ __('applyNext.navigation.next') }}"/>
```
davidrapson: How come this button has a `formaction` and the other doesn't? Do we need this?

controllers/apply/form-router-next/steps.js
```diff
@@ -237,12 +237,24 @@ module.exports = function(formId, formBuilder) {
...
+                            baseUrl: res.locals.formBaseUrl,
+                            sectionSlug: req.params.section,
+                            currentStepIndex: stepIndex,
+                            copy: res.locals.copy
+                        });
+                        redirectUrl = previousPage.url;
+                    } else {
+                        const { nextPage } = form.pagination({
```
mattandrews: It would be nice to turn this into a function somewhere since it's basically the same code for `nextPage`/`previousPage`.

mattandrews:
Just tried this locally and the top set of nav links don't submit as they're not inside the `<form>`

mattandrews:
Also... we've lost the field names below the buttons:
image

image

Finally, this might be more of a UX question, but the buttons just display a regular cursor now, not a hand/hyperlink... I guess this is right technically, but since they look like links, they could do with a hand?

views/components/user-navigation/macro.njk
NikhilNanjappa: My bad. I have to apply the `formaction` to the this button too.

controllers/apply/form-router-next/steps.js
NikhilNanjappa: Good shout. I will name this function appropriately.

controllers/apply/form-router-next/steps.js
NikhilNanjappa: Sure. I'll put this in a named function. Will also add the `cursor: pointer` to the .btn-link

assets/sass/components/_buttons.scss
```diff
@@ -221,7 +221,7 @@ $buttonColor: get-color('brand', 'primary');
     text-decoration: underline;
     border: none;
     border-radius: 0;
-    cursor: inherit;
+    cursor: pointer;
```
davidrapson: 👎 Sorry! Can we add the pointer to _just_ those link buttons not all buttons

controllers/apply/form-router-next/steps.js
```diff
@@ -172,6 +172,23 @@ module.exports = function(formId, formBuilder) {
                 stepFields.map(f => f.name).includes(item.param)
             );
 
+            function prevOrNextBtnNotClicked() {
```
davidrapson: Not sure about this name? Sounds like a double negative with `not` in the name, can name this what it does. i.e. shouldRenderErrors?

controllers/apply/form-router-next/steps.js
```diff
@@ -237,24 +254,7 @@ module.exports = function(formId, formBuilder) {
...
-                    } else {
-                        const { nextPage } = form.pagination({
-                            baseUrl: res.locals.formBaseUrl,
-                            sectionSlug: req.params.section,
-                            currentStepIndex: stepIndex,
-                            copy: res.locals.copy
-                        });
-                        redirectUrl = nextPage.url;
-                    }
+                    const redirectUrl = determineRedirectUrl();
```
davidrapson: You can inline references to this variable, call the method where it is used instead.

assets/sass/components/_buttons.scss
NikhilNanjappa: But would we not need the `pointer` to all the buttons which looks like links.

controllers/apply/form-router-next/steps.js
NikhilNanjappa: I didnt want to include the `errors` into this check tbh as I wanted to use them separately. I could combine them and named as `shouldRenderErrors`.

controllers/apply/form-router-next/steps.js
NikhilNanjappa: Did you mean directly call the method when needed instead of creating a variable `redirectUrl` ?

assets/sass/components/_buttons.scss
mattandrews: I think you're seeing the wrong classname here @davidrapson, this one **is** the link button class?

davidrapson: Ah fair enough!

PR #2423, opened 2019-10-07
Address requirement fix

davidrapson:
The schema lines for main and senior contact addresses were missing a `.required()`

```
Joi.ukAddress().compare(Joi.ref('seniorContactAddress'))
```

This meant that whilst an object would be validated against the expected schema if some values were present, if somehow _nothing_ was provided we'd consider this OK.

Our "empty form" snapshot test technically caught this but due to us using the same messages for all address fields it was difficult to spot; particularly as snapshot tests are based on the assumption of a _good_ first state to test against. 

I've made the original snapshot test more explicit to also include the field name in the snapshot and added an additional test to cover the field requirement.


controllers/apply/awards-for-all/form.test.js
```diff
@@ -865,6 +880,26 @@ describe('Contacts', () => {
         });
     });
 
+    test('contact address required by default', () => {
```
davidrapson: This test probably isn't _strictly_ required as it's covered by the amended empty form test but worth having an an explicit check.

PR #2427, opened 2019-10-07
Revert "Revert "Save and redirect without rendering validation errors""

NikhilNanjappa:
Reverts biglotteryfund/blf-alpha#2426

### DIFF
--- a/views/components/user-navigation/macro.njk
+++ b/views/components/user-navigation/macro.njk
@@ -13,26 +13,12 @@
                     {% if previousPage %}
                         <li class="user-nav__pagination-link user-nav__pagination-link--prev">
                             <span class="user-nav__pagination-arrow">←</span>
-                            <a href="{{ previousPage.url }}">
-                                <span>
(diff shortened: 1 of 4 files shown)


views/components/user-navigation/macro.njk
```diff
@@ -13,12 +13,23 @@
                     {% if previousPage %}
                         <li class="user-nav__pagination-link user-nav__pagination-link--prev">
                             <span class="user-nav__pagination-arrow">←</span>
-                            <input class="btn-link" name="previousBtn" type="submit" formaction="" value="{{ __('applyNext.navigation.previous') }}"/>
+                            <div class="user-nav__pagination-btn-container">
```
davidrapson: Do we need this extra `user-nav__pagination-btn-container` div?

NikhilNanjappa: Yup (or a `span`). It was a solution for steps label not wrapping below the next/prev buttons (earlier it was within `<a>` element).

PR #2432, opened 2019-10-07
Project name

davidrapson:
This adds a project name field to the get advice form. Determined it is beneficial to allow distinguishing between multiple ideas. Pending confirmation of exactly where the field should appear.


davidrapson:
[Add project name to get advice form so applicants can refer to previous submissions at a glance](https://trello.com/c/4wa4lTqr/856-add-project-name-to-get-advice-form-so-applicants-can-refer-to-previous-submissions-at-a-glance)

davidrapson:
Merging this as is, we need to decide if this is the right place to ask the question, but it's safe to merge.

PR #2440, opened 2019-10-08
Config cleanup

davidrapson:
Config simplification where settings don't change per environment

- Inline `indexable` config, only used for robots
- Merge session config and cookie name into same config
- Remove unused `domains.base`
- Remove need for `enableSiteSurvey` this can be set as default in locals
- Remove need for name change message (this is always on)
- Remove complexity around CSP config, no need for arguments
- Replace `enableCloudWatchLogs` config in favour of `isDev` check which is more direct


davidrapson:
I'm pro removing it now tbh, but yeah after a year feels more realistic.

PR #2442, opened 2019-10-08
Track clicks on email expiry prompts

mattandrews:
We can link users directly to their expiring application in the emails, and also track when they click on these links so we have some idea of how effective these reminders are. Confirmed this works as expected locally.

### DIFF
--- a/controllers/apply/form-router-next/index.js
+++ b/controllers/apply/form-router-next/index.js
@@ -238,6 +238,14 @@ function initFormRouter({
      * Route: Edit application
      */
     router.get('/edit/:applicationId', function(req, res) {
+        // If this link includes a source (s) parameter, track it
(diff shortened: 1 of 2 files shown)


controllers/apply/form-router-next/index.js
```diff
@@ -238,6 +238,14 @@ function initFormRouter({
      * Route: Edit application
      */
     router.get('/edit/:applicationId', function(req, res) {
+        // If this link includes a source (s) parameter, track it
+        // eg. to analyse usage of expiry reminder emails
+        if (get(req.query, 's') === 'expiryEmail') {
```
davidrapson: I think `if(req.query.s === 'expiryEmail')` should be safe to do without the `get`

PR #2443, opened 2019-10-08
Forgot password unregistered

davidrapson:
Identified a bug where forgotten password requests for _unregistered_ users would hang. Steps to reproduce:

- Go to https://www.tnlcommunityfund.org.uk/user/password/forgot
- Enter an unregistered email address

Expected behaviour is that a placebo version of the forgotten password confirmation screen should be shown.

Added a cypress test for unregistered accounts to confirm the bug and implemented a fix.

### DIFF
--- a/controllers/user/password.js
+++ b/controllers/user/password.js
@@ -135,6 +135,9 @@ router
                         logger.info('Password reset request succeeded');
                         renderForgotForm(req, res);
                     }
+                } else {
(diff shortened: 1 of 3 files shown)


davidrapson:
[Bug report: 504 error shown when using forgotten password form](https://trello.com/c/TMR42fn7/877-bug-report-504-error-shown-when-using-forgotten-password-form)

PR #2456, opened 2019-10-08
Tweak the names/form values of a few Scottish regions

mattandrews:
Updates the names and the form values of a few incorrect regions.

### DIFF
--- a/controllers/apply/index.js
+++ b/controllers/apply/index.js
@@ -1,6 +1,8 @@
 'use strict';
 const express = require('express');
 const path = require('path');
+const { get, includes } = require('lodash');
+const { Op } = require('sequelize');
 
 const {
     verifyTokenUnsubscribeApplicationEmails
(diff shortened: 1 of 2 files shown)


controllers/apply/index.js
```diff
@@ -71,6 +73,66 @@ router.get('/emails/unsubscribe', async function(req, res) {
...
+    ];
+
+    const invalidRegionNames = replacements.map(_ => _.from);
+
+    const applications = await PendingApplication.findAllByForm(
+        'awards-for-all'
+    );
+
+    const appsToUpdate = applications.filter(app => {
+        return includes(
```
davidrapson: FYI `Array.includes` is supported in the version of node we run so we don't need lodash/includes. Doesn't matter here though.

mattandrews: Aha!

davidrapson: image

PR #2459, opened 2019-10-09
Use new macros for forgotten password form

davidrapson:
Spotted that the forgotten password screen was the only one in the account section that wasn't using the new form macros.

**Before**
image

**After**
image


NikhilNanjappa:
Does the new form not have the "* Required" text after the label ?

davidrapson:
Yep, we flipped things so we only mark optional fields instead.

PR #2464, opened 2019-10-09
Convert feedback to common style

davidrapson:
image


controllers/tools/feedback.js
```diff
@@ -6,10 +6,15 @@ const { Feedback } = require('../../db/models');
 
 const router = express.Router();
 
-router.get('/', async (req, res, next) => {
+router.get('/', async function(req, res, next) {
```
mattandrews: Out of interest, why this change? Seeing a few more of these lately, is this a better style we should be using for these callbacks?

PR #2469, opened 2019-10-10
Login/Logout in global header

NikhilNanjappa:
https://trello.com/c/GYyjTZGX/858-universal-login-in-header


assets/js/common/forms.js
```diff
@@ -220,6 +220,19 @@ function initHotjarTracking(formId) {
     trackDetailsClicks(scopedFormClass);
 }
 
+// Update the Login link to Logout if user signs in
+function updateSecondaryNav() {
+    $.ajax({
```
mattandrews: Maybe we could make this AJAX call a shared function so we can use the same code in `session.js` – perhaps export this function from there, eg:

```
function getSession() {
    return $.ajax({
        type: 'get',
        url: '/user/session',
        dataType: 'json'
    });
}

getSession().then(response => {
        if (response.isAuthenticated) {
            $('.js-toggle-login').toggleClass('is-conditional-link');
        }
}
```

NikhilNanjappa:
@mattandrews - I was already in a process of doing something similar as your simplifying suggestion. After looking at your comment, I just renamed few classes from `js-account-link--login/logout` to `js-toggle-login/logout` to keep it consistent with other class names. I also added the `js-hidden` bit now. Removed the `js-account-link` as its not needed and added a `global-header__account-link` to manipulate the CSS only.

views/includes/global-header.njk
```diff
@@ -24,7 +24,32 @@
     <div class="global-header__inner">
         <div class="global-header__extra">
             <ul class="global-header__navigation-secondary">
-                <li><a href="{{ globalNavigation.home.url }}">{{ globalNavigation.home.label }}</a></li>
+                <noscript>
```
mattandrews: This is how this renders with JS off:

image

I should have been clearer, sorry! My solution posted here didn't require the use of `<noscript>` anymore – the `.js-hidden` classes will set `display: none` **for users with JavaScript enabled** so users without JavaScript will still see them, which is why the Log Out link has an extra class of `u-hidden` which hides it for all users (until the JS makes it visible again). 

You can delete the `<noscript>` block here :)

NikhilNanjappa: Ahh, you're right!

It was my bad tbh. I understood you earlier and knew the intention as well, just forgot to remove that bit :) Updated now!

davidrapson:
🚀 Cool, let's ship this!

PR #2471, opened 2019-10-10
Applications dashboard

davidrapson:
Scratching around the surface of rendering a basic dashboard at `/apply` with latest application across both simple and standard forms.


controllers/apply/dashboard.js
```diff
@@ -88,4 +88,42 @@ router.get(
     }
 );
 
+router.get(
+    '/all',
+    csrfProtection,
+    requireActiveUser,
+    injectCopy('applyNext'),
+    async function(req, res, next) {
+        // @TODO: can combine this and above to a common enrich function?
```
davidrapson: Yep, I reckon we should as the'll be the same details for each view

PR #2477, opened 2019-10-11
Fix date errors

mattandrews:
Few things here. 

Adds some missing example formats to date fields:

image

Simplifies things with "on or after" and a valid date, rather than invalid:
image

Fixes a typo which prevented this error from showing up for invalid start dates:

image

### DIFF
--- a/controllers/apply/awards-for-all/fields/project-date-range.js
+++ b/controllers/apply/awards-for-all/fields/project-date-range.js
@@ -97,7 +97,7 @@ module.exports = function(locale) {
                 })
             },
             {
-                type: 'datesRange.startDate.invalid',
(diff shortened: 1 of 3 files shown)


controllers/apply/awards-for-all/fields/project-date-range.js
```diff
@@ -35,7 +34,7 @@ module.exports = function(locale) {
         explanation: localise({
             en: `<p>
                 If you don't know exactly, your dates can be estimates.
-                But you need to start your project after
+                But you need to start your project on or after
```
mattandrews: I'm just waiting on a Welsh translation of this change, FYI.

PR #2485, opened 2019-10-15
Output github compare URL rather than commit list

davidrapson:
We've been finding that the github compare url e.g. https://github.com/biglotteryfund/blf-alpha/compare/COMMITA...COMMITB is far more useful than creating our own list of commits. 

image

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -2,6 +2,7 @@
 'use strict';
 require('dotenv').config();
 const config = require('config');
+const { stripIndents } = require('common-tags');
 
 const has = require('lodash/has');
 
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -121,71 +118,37 @@ function getEnvStatus(env) {
...
-
-    // get server statuses
-    return Promise.all(lookups).then(responses => {
-        let test = responses[0];
-        let prod = responses[1];
-
-        // get the commit IDs
-        let ids = {
-            test: test.COMMIT_ID,
-            prod: prod.COMMIT_ID || 'c54660f391ed5542308d0deb4b87f07ce54ec120' // last build before adding commit_id
```
davidrapson: I've removed this on the fairly safe assumption that we'll never deploy this commit as it's over two years old

PR #2487, opened 2019-10-15
data-hj-whitelist only applies to inputs

davidrapson:
According to hotjar docs `data-hj-whitelist` only applies to inputs. Numeric blacklisting is handled via an account level setting so is a global config.


mattandrews:
So do we need to blacklist things at Hotjar before we can merge this?

davidrapson:
No it already does it, this code does nothing

PR #2490, opened 2019-10-15
Fixup integration tests (Highlands -> Highland) and update schema

davidrapson:
We randomise the country for our integration tests so this was sporadically failing because of the location name change (Highlands -> Highland). Also updates the schema doc with the new values.


cypress/integration/integration_spec.js
```diff
@@ -558,11 +560,10 @@ it('should submit full awards for all application', () => {
                 };
                 break;
             case 'Scotland':
-                location = sample([
```
davidrapson: We had this mostly when we were submitting lots of random test data to Salesforce, we have easier ways of doing this now (and less of a requirement) so we can reduce the randomness here.

PR #2495, opened 2019-10-15
Application cards

davidrapson:
Splitting out the front-end component for the new applications dashboard. This pull request will contain only some new macros with example props, but not hook things up to the new screens (@NikhilNanjappa is working on the queries for that part)

### DIFF
--- a/controllers/apply/lib/form-model.js
+++ b/controllers/apply/lib/form-model.js
@@ -4,6 +4,7 @@ const find = require('lodash/find');
 const findIndex = require('lodash/findIndex');
 const findLastIndex = require('lodash/findLastIndex');
 const flatMap = require('lodash/flatMap');
+const get = require('lodash/fp/get');
 const has = require('lodash/has');
 const isEmpty = require('lodash/isEmpty');
 const pick = require('lodash/pick');
@@ -17,6 +18,8 @@ const normaliseErrors = require('./normalise-errors');
 
 class FormModel {
     constructor(props, data = {}, locale = 'en') {
+        const localise = get(locale);
+
         this.title = props.title;
         this.startLabel = props.startLabel;
         this.allFields = props.allFields;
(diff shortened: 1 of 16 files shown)


davidrapson:
Started with the question "How far can we get with minimal new CSS?". The answer is quite far:

**Rough version**:

image

**Design**

image

davidrapson:
Closer. Now with progress bar

image

davidrapson:
Got something that I think is in a position to merge (pending review). Will make a follow on pull request for the other card styles, we already have an "action-card" pattern that I think we might want to repurpose here, and is much simpler than these application cards.

controllers/apply/lib/form-model.js
```diff
@@ -137,10 +140,22 @@ class FormModel {
                 }
             }
 
+            function sectionStatusLabel() {
```
davidrapson: Added this here so avoid the logic in the templates, making it easier to work with progress in macros.

views/components/application-card/examples.njk
```diff
@@ -0,0 +1,89 @@
...
+            value: "2 December, 2020–2 December, 2020"
+        }, {
+            label: "Location",
+            value: "Sutton Coldfield"
+        }, {
+            label: "Organisation",
+            value: "The Scouts of Sutton Coldfield"
+        }],
+        progress: {
+          sectionsComplete: 2,
```
mattandrews: Could this be calculated inside the macro rather than having to (presumably) do it manually every time we use this component? eg. we could add a filter function for templates which returns this total?

davidrapson:
['Dashboard' which shows the last activity on your account for new and existing applicants](https://trello.com/c/WyaKFf1v/853-dashboard-which-shows-the-last-activity-on-your-account-for-new-and-existing-applicants)

PR #2506, opened 2019-10-17
Updated action-card style

davidrapson:
Builds off https://github.com/biglotteryfund/blf-alpha/pull/2495

One of the two card styles for the new dashboards is an "action card", we already have a pattern like this which we used for the digital fund pages. Rather than creating a completely new pattern I thought it made sense to repurpose this.

Screenshot 2019-10-17 at 15 11 45

Side-effect is that is naturally changes the digital fund pages, but I think this makes sense for consistency and is fairly low risk as that fund is closed for applications.

**Before**

image

**After**

image

### DIFF
--- a/assets/sass/components/_applications.scss
+++ b/assets/sass/components/_applications.scss
@@ -301,6 +301,7 @@ $statuses: (
    ========================================================================= */
 
 .application-card-new {
+    box-shadow: 0 2px 6px 1px rgba(0, 0, 0, 0.1);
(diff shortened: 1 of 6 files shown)


davidrapson:
['Dashboard' which shows the last activity on your account for new and existing applicants](https://trello.com/c/WyaKFf1v/853-dashboard-which-shows-the-last-activity-on-your-account-for-new-and-existing-applicants)

PR #2507, opened 2019-10-17
Randomise cypress test for salesforce testing

davidrapson:
Randomise cypress test for salesforce testing

### DIFF
--- a/cypress/integration/integration_spec.js
+++ b/cypress/integration/integration_spec.js
@@ -1064,11 +1064,13 @@ it('should submit full awards for all application', () => {
     });
 });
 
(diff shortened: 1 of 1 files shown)


controllers/apply/form-router-next/submission.js
```diff
@@ -88,7 +88,7 @@ module.exports = function(
             /**
              * Store submission in salesforce if enabled
              */
-            if (enableSalesforceConnector === true && !appData.isTestServer) {
```
davidrapson: Removed this so we can have Cypress push to Salesforce locally, will reinstate after testing and before merging this PR.

PR #2512, opened 2019-10-18
Misc AFA form fixes

mattandrews:
A few things in here based on minor niggles and bug reports:

## Changes completion text to correctly refer to the emails we send
(eg. that they go to the named contacts, not the applicant)

## Adds a bilingual logo for emails sent in Welsh
(also needed for Salesforce)

## Add navigation to Help pages
Based on this user feedback:

> "When I looked at what the bank statements requirements were it was not possible to go back to the form and I had to log in again."

**Before / after**
image

## Fix an issue with bank detail validation
This user report highlighted a bug:

>"A notice on the sortcode page probably just needed to reflect the current error where if you put in hyphens it doesn't let you proceed, even when you delete them- should save some phonecalls!"

It turned out that the bank account API checks were performed against the raw input data rather than the validated Joi-approved data. In the case of the sort code, this meant that if you entered `12-34-56`, Joi would transform this into `123456`, but the bank API would get the first (hyphenated) value, which fails. You'd be returned to the page with an error saying the sort code was invalid, but the value would be `123456` – which is now valid! So if you submitted the page again without making changes, the bank API check would succeed. This change makes sure the API call uses the Joi-validated data to do the lookup.

## Warn users when visiting the form over `file://` protocol

We see a small but steady stream of users attempting to access the form after saving it to their desktops. This fails for a variety of reasons (missing JS files, failed security policies etc), but the most important one is that the form's `action` attribute is blank, so when they submit a step it goes nowhere. The solution here is to show a prominent error when the form is accessed over `file://` URLs:

image


assets/sass/components/_buttons.scss
```diff
@@ -164,6 +164,15 @@ $buttonColor: get-color('brand', 'primary');
     }
 }
 
+/**
+ * Buttons: Warning Variant
+ */
+.btn.btn--warning {
```
davidrapson: Do we need to add a new button style here for a one time use?

controllers/apply/form-router-next/steps.js
```diff
@@ -241,7 +241,7 @@ module.exports = function(formId, formBuilder) {
                      */
                     if (step.preFlightCheck) {
                         try {
-                            await step.preFlightCheck();
+                            await step.preFlightCheck(dataToStore);
```
davidrapson: Nice catch!

views/includes/metaHead.njk
```diff
@@ -24,5 +24,10 @@
     if (sessionStorage.fontsLoaded) {
         docEl.className += ' ' + 'fonts-loaded';
     }
+
+    if (window.location.protocol === 'file:') {
+        alert("It looks like you've saved a copy of this webpage to work with on your own computer. This will not work correctly and your data will be lost if you attempt to complete this form. Please return to our website at https://www.tnlcommunityfund.org.uk to continue applying.");
```
davidrapson: Do we need this in addition to the message? Saving a web page offline isn't unreasonable, feels a bit much to alert on every page.

mattandrews: Oops, this was an initial version of this change which I thought I'd reverted! Just removed it.

assets/sass/components/_buttons.scss
mattandrews: Do you mean it should just be a new standalone style? I guess it's conceivable we'll want to put a button inside an error box at some stage...? It looked awful in the default pink (on the red border box)!

davidrapson: Ah fair enough! No makes sense to have a new variant then

davidrapson:
🚀

davidrapson:
[Bug: Support hyphens and spaces when passing bank details to checker API](https://trello.com/c/0BpFjdxD/901-bug-support-hyphens-and-spaces-when-passing-bank-details-to-checker-api)

PR #2513, opened 2019-10-18
Add new Funding section

mattandrews:
This pairs with https://github.com/biglotteryfund/craft-dev/pull/194 and handles all URLs under `/funding/*` (besides existing ones). The goal is to revamp the entire Funding section and make the URLs a bit nicer at the same time (eg. drop redundant stuff like `/funding-guidance/`).

The main change here is merging two same-but-different templates: "listing pages" (eg. grids of child pages) and "information pages" (content). This leads to some slightly thorny template code in the combined "CMS page" template (eg. to allow for the different use cases and to display children pages, if required, in different ways) but I think this represents a direction of working for the CMS into 2020 – eg. generic templates for **all** CMS-powered pages (ideally!) which support showing children in different ways.

At the moment this new section is non-PROD environments only while we prepare the content, but when we're ready we can switch it on then manually start linking to it.

Things we might want to do in future: make the entire `/funding` landing page driven dynamically off its child pages in this section – though this would involve making it possible to add child "pages" which just link to other non-CMS URLs (eg. grant search).

Thoughts welcome!

### DIFF
--- a/assets/sass/scaffolding/_grid.scss
+++ b/assets/sass/scaffolding/_grid.scss
@@ -13,8 +13,12 @@ $flexGridSpacing: $spacingUnit;
  */
 .flex-grid {
     @include reset-list();
-    // Overrides constrained content containers
-    max-width: 100% !important;
+
+    .u-constrained-content-wide & {
+        // Overrides constrained content containers
+        // eg. grids within flexible content areas
+        max-width: 100% !important;
+    }
 
     @include mq('medium-minor') {
         display: flex;
--- a/controllers/common/basicContent.test.js
+++ b/controllers/common/basicContent.test.js
@@ -1,0 +1,103 @@
+/* eslint-env jest */
+'use strict';
+const { getChildrenLayoutMode } = require('./index');
+
+describe('CMS page templates', () => {
+    it('should work out feasible child page layout mode', () => {
+        const pageGrid = {
+            title: 'foo',
+            childPageDisplay: 'grid',
+            children: [
+                {
+                    title: 'bar',
+                    trailImage: 'bar.jpg'
+                },
+                {
+                    title: 'baz',
+                    trailImage: 'baz.jpg'
+                }
+            ]
+        };
+
+        expect(getChildrenLayoutMode(pageGrid)).toEqual('grid');
+
+        const pageGridMissingImages = {
+            title: 'foo',
+            childPageDisplay: 'grid',
+            children: [
+                {
+                    title: 'bar',
+                    trailImage: 'bar.jpg'
+                },
+                {
+                    title: 'baz',
+                    trailImage: 'baz.jpg'
+                },
+                {
+                    title: 'test'
+                }
+            ]
+        };
+
+        expect(getChildrenLayoutMode(pageGridMissingImages)).toEqual('list');
+
+        const pageList = {
+            title: 'foo',
+            childPageDisplay: 'list',
+            children: [
+                {
+                    title: 'bar',
+                    trailImage: 'bar.jpg'
+                },
+                {
+                    title: 'baz',
+                    trailImage: 'baz.jpg'
+                },
+                {
+                    title: 'test'
+                }
+            ]
+        };
+
+        expect(getChildrenLayoutMode(pageList)).toEqual('list');
+
+        const pageNone = {
+            title: 'foo',
+            childPageDisplay: 'none',
+            children: [
+                {
+                    title: 'bar',
+                    trailImage: 'bar.jpg'
+                },
+                {
+                    title: 'baz',
+                    trailImage: 'baz.jpg'
+                },
+                {
+                    title: 'test'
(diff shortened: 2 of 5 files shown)


assets/sass/scaffolding/_grid.scss
```diff
@@ -13,8 +13,12 @@ $flexGridSpacing: $spacingUnit;
  */
 .flex-grid {
     @include reset-list();
-    // Overrides constrained content containers
-    max-width: 100% !important;
+
+    .u-constrained-content-wide & {
+        // Overrides constrained content containers
+        // eg. grids within flexible content areas
+        max-width: 100% !important;
```
mattandrews: In my efforts to fix this bug:

image

... I caused this one: 

image

controllers/common/views/cms-page.njk
```diff
@@ -0,0 +1,119 @@
...
+{% block content %}
+
+    {% if pageHero.image %}
+        {{ hero(title, pageHero.image) }}
+    {% endif %}
+
+    <main role="main" id="content" {% if pageHero.image %}class="nudge-up"{% endif %}>
+        {% block contentPrimary %}
+
+            {% if content.introduction %}
```
mattandrews: All of this is to allow backward compatibility with "legacy" information/listing pages (eg. [this section](https://www.tnlcommunityfund.org.uk/funding/funding-guidance/managing-your-funding)).

assets/sass/scaffolding/_grid.scss
davidrapson: This feels…very specific?

controllers/common/index.js
```diff
@@ -79,44 +79,46 @@ function basicContent({ lang = null, customTemplate = null } = {}) {
...
-                    // What layout mode should we use? (eg. do all of the children have an image?)
-                    const missingTrailImages = content.children.some(
-                        page => !page.trailImage
-                    );
-                    const childrenLayoutMode = missingTrailImages
-                        ? 'plain'
-                        : 'heroes';
-                    if (missingTrailImages) {
+                } else {
+                    let childrenLayoutMode = 'list';
```
davidrapson: Quite a lot of logic here, perhaps try and extract this out so we can unit test it?

controllers/common/views/cms-page.njk
davidrapson: Is there a way we could work out if it's "legacy" earlier on in the controller and render two different templates?

controllers/common/views/cms-page.njk
```diff
@@ -0,0 +1,119 @@
...
+                                {{ miniatureHero({
+                                    "title": page.trailText | default(page.title, true),
+                                    "linkUrl": page.linkUrl,
+                                    "image": { "url": page.trailImage if page.trailImage else page.photo }
+                                }, isLarge = true) }}
+                            </li>
+                        {% endfor %}
+                    </ul>
+                </section>
+            {% elseif childrenLayoutMode === 'list' and content.flexibleContent.length > 1 %}
```
davidrapson: Why do we need `and content.flexibleContent.length > 1`?

assets/sass/scaffolding/_grid.scss
mattandrews: I'm not really sure how else to achieve it – it's because the `s-prose` classes do things to `ul` and `li` elements which are quite difficult to override. Suggestions welcome!

controllers/common/views/cms-page.njk
mattandrews: Probably – will have a look.

controllers/common/index.js
mattandrews: Done – and ended up specifying in `routes.js` which pages get this new template, so we can drop support for the old "legacy" template fields without breaking older pages.

assets/sass/scaffolding/_grid.scss
davidrapson: Don't put these inside an s-prose?

davidrapson: As discussed on chat, this all makes sense for now, and we can come back to it to try and find a way to not make the prose class a requirement for a flexible content area.

controllers/common/basicContent.test.js
```diff
@@ -0,0 +1,103 @@
+/* eslint-env jest */
+'use strict';
+const { getChildrenLayoutMode } = require('./index');
+
+describe('CMS page templates', () => {
+    it('should work out feasible child page layout mode', () => {
```
davidrapson: ❤️ 

btw, if there is only one test block you can skip the `describe` and call `it` (or `test` as simpler alias) on its own

controllers/common/views/cms-page.njk
davidrapson: Not a big deal, might make things easier to delete but the logic isn't complicated

PR #2514, opened 2019-10-18
Dashboard integration

davidrapson:
Placeholder pull request to integrate dashboard screen


davidrapson:
Work so far:

image


Needed to tweak the queries to account for when we don't have either a pending or submitted application, and after speaking to Earl realised that the "has previous" checks can be simpler and only need to check pending applications

Biggest bit of work left to do is to work out how to prepare the data for the view on a per form model basis for both pending and submitted applications

davidrapson:
Starting to get real data in:

image

@NikhilNanjappa I'd appreciate your thoughts, particularly around how we do the "enrich step" as it's a bit hacky at the moment

NikhilNanjappa:
Worked on the 'Dashboard - All applications' view:

Screenshot 2019-10-18 at 16 45 46
Screenshot 2019-10-18 at 16 46 10

Realize its WIP, but just bringing up some minor points that I noticed so we don't forget.

- If no latest application is found, the title wont be displayed
- The title text is below the text 'Pick up where you left off ...'
- If no application(s) is found, the `actionCard`s stack one below the other rather than next to each other.

davidrapson:
Added no results states to both views:

image

image

davidrapson:
Added user navigation links and tidied up both screens:

image

controllers/apply/views/dashboard.njk
```diff
@@ -0,0 +1,54 @@
...
+{% from "components/action-card/macro.njk" import actionCard %}
+{% from "components/application-card/macro.njk" import applicationCard with context %}
+{% from "components/user-navigation/macro.njk" import userNavigation with context %}
+
+{% block content %}
+    <main role="main" id="content">
+        <section class="content-box u-inner-wide-only">
+            {{ userNavigation(userNavigationLinks) }}
+
+            <h1 class="t--underline">{{ title }}</h1>
```
NikhilNanjappa: The title (Latest Application) would not be present if latest application does not exist.

davidrapson: This part is intentional (for now). I think there should always be a `h1` on the page, so leaving this as is but plan to review with folks once we have a preview link out.

config/locales/en.yml
```diff
@@ -1556,6 +1556,25 @@ applyNext:
       title: New application
       body: You can create a new funding application for %s here
       cta: Start new application
+  dashboardNew:
+    latest:
+      title: Latest application
+    all:
+      title: All applications
```
NikhilNanjappa: Very minor changes but the dashboard links are capitalized 'Latest Application' and 'All Applications'.

I'm guessing the Welsh translation is already on the way ?

davidrapson: We always sentence case the titles (Earl possibly missed this), Welsh translation will only come once we've finalised the screens.

controllers/apply/views/dashboard.njk
NikhilNanjappa: Just confirming, is it the same for the button name being 'Continue' instead of 'Continue with this application' and Last Updated value not having the time ?

NikhilNanjappa:
Outstanding work on this one!! 💯 👍 Awesome effort!

All LGTM for now 👍 I'm also guessing the new redirections will be in place after the screens are confirmed ?

controllers/apply/views/dashboard.njk
davidrapson: For now yeah, we're a little tighter on space than was thought so sticking with a short label for now (we can always come back to that), I'm using `.calendar` for the last updated which I think only includes the time if the date is recent.

PR #2515, opened 2019-10-21
Northern Ireland country validation

davidrapson:
Include Northern Ireland in valid options so we can bypass disabled option in production to perform live tests. Still disabled to applicants


davidrapson:
[Test email sending in production for Northern Ireland applications](https://trello.com/c/V6UdgWJV/872-test-email-sending-in-production-for-northern-ireland-applications)

PR #2521, opened 2019-10-21
Updated charity or company text

davidrapson:
image

Pending content review, and then welsh translation

### DIFF
--- a/controllers/apply/awards-for-all/fields/senior-contact-role.js
+++ b/controllers/apply/awards-for-all/fields/senior-contact-role.js
@@ -40,7 +40,8 @@ module.exports = function(locale, data) {
                 localise({
                     en: oneLine`As you're a company, your senior contact needs
                         to be listed as one of your board members on Companies House.`,
-                    cy: `` // @TODO: i18n
+                    cy: oneLine`Gan eich bod yn gwmni, rhaid i’ch uwch gyswllt gael
+                        ei restru fel un o’ch aelodau bwrdd ar Dŷ’r Cwmnïau.`
                 })
             );
         }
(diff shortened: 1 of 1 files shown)


davidrapson:
[Revise content on conditional text that appears on Senior Contact page if customers select 'Not-for-profit company'](https://trello.com/c/85jji8yG/862-revise-content-on-conditional-text-that-appears-on-senior-contact-page-if-customers-select-not-for-profit-company)

controllers/apply/awards-for-all/fields/senior-contact-role.js
```diff
@@ -73,9 +73,12 @@ module.exports = function(locale, data) {
...
                     localise({
-                        en: oneLine`Your senior contact must be listed as a
-                            member of your organisation's board or committee
-                            with the Charity Commission/Companies House.`,
+                        en: oneLine`If you're a company, your senior contact
+                            must be listed as a member of your organisation's
+                            board with Companies House.
+                            If you're an unincorporated Charity, your Senior
+                            contact should be listed as a Trustee on the
+                            Charities regulator website.`,
```
mattandrews: Think Regulator should be capitalised, eg. https://www.charitiesregulator.ie/en

(could/should this be a link to that site, too?)

scottoakley: I did a quick review of the content (mainly adding subheadings for scannability and also reducing excess text). See what you think: 

**If you're a company**
Your senior contact needs to be listed as one of your board members on Companies House. 

**If you're a registered charity (that's not also a company)**
Your senior contact needs to be listed as one of your trustees with your charity regulator. 

An alternative suggestion - is it too complex to make each subheading appear only to the relevant types of organisations? So that we can reduce text on-screen. 

Then it coudl be just something like 'As you're a company, you senior contact needs to be listed as one of your board memebrs on Companies House.' and so on

davidrapson: No, that's simpler if anything, let's go with separate text

scottoakley: Great! So:

'As you're a company, your senior contact needs to be listed as one of your board members on Companies House.'

And:

'As you're a registered charity (that's not also a company), your senior contact needs to be listed as one of your trustees with your charity regulator.'

Might be good to get Kirsten to do a review too?

davidrapson:
Got the new text translated, so we can merge this once we're happy

PR #2522, opened 2019-10-21
Only allow Northern Ireland for standard form

davidrapson:
image

Need to decide on wording first, but this mirrors the previous country logic from awards for all


davidrapson:
[Decide how we want to message for other countries when launching in Northern Ireland](https://trello.com/c/5xy5G2Xw/859-decide-how-we-want-to-message-for-other-countries-when-launching-in-northern-ireland)

PR #2527, opened 2019-10-22
Add max length to standard fields

davidrapson:
Add missing max length to standard fields. All custom types like addresses and names already have max lengths, this covers the generic text fields.


davidrapson:
[Add max-length limits to field validations](https://trello.com/c/pgDP33Rf/888-add-max-length-limits-to-field-validations)

PR #2538, opened 2019-10-23
Switch to postcss-foft-classes

davidrapson:
`postcss-fout-with-a-class` came with a bunch of unnecessary babel code as a dependency which was flagging an unfixable audit vulnerability warning. Switches to an alternative via https://www.zachleat.com/web/font-loading-classes/ which has no dependencies and is a little more flexible . Has the same effect on font loading


davidrapson:
Welp

image

Yeah a lot of unneeded dependencies

davidrapson:
Also means this

image

PR #2540, opened 2019-10-23
Node v12

davidrapson:
Node v12 is now in LTS so it makes sense to switch to it.


.travis.yml
```diff
@@ -11,7 +11,7 @@ dist: xenial
 
 # Defaults
 language: node_js
-node_js: 'lts/*'
```
davidrapson: Due to the wildcard in `lts/*` we've actually been running node 12 in CI for the last couple of days (node 12 entered LTS on 21st). Switching explicitly to a numbered version

PR #2544, opened 2019-10-23
Standard name change

davidrapson:
Updates name of standard form to "Your funding proposal". Covers copy changes and any code changes to better reflect new name / remove references to get advice. Ended up be more straightforward than I thought it'd be.


davidrapson:
[Name change from Get Advice on Your Idea to Your Funding Proposal](https://trello.com/c/E6NmU7xC/884-name-change-from-get-advice-on-your-idea-to-your-funding-proposal)

PR #2558, opened 2019-10-24
Restore CMS preview toolbar

mattandrews:
Missed this when updating the preview functionality. The CMS doesn't reliably return preview/draft status anymore (eg. content can be "live" when previewing a draft) so we rely on the Craft query params to determine if something is a preview link.


mattandrews:
Ended up with commits from another (merged) branch here because I branched before pulling those changes... ugh.

PR #2561, opened 2019-10-25
Retained token for subsequent pwd reset requests

NikhilNanjappa:
After some long testing, found an issue - any subsequent requests after the first will not have token within the `req.body` and hence will be treated as an unauthorized request. This update will fix that.

https://trello.com/c/BhLy1R1r/903-check-for-continued-password-reset-bugs

### DIFF
--- a/controllers/user/password.js
+++ b/controllers/user/password.js
@@ -243,13 +243,11 @@ router
             }
         } else {
             // Handle an unauthorised user who submitted the form with a token
-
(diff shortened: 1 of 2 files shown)


davidrapson:
Might be nice if we could cover this case with a cypress test, but good to go

NikhilNanjappa:
@davidrapson - Just added some tests to cover the multiple weak passwords instance.

PR #2564, opened 2019-10-25
Over 10k validation

davidrapson:
Updates validation rule for standard form to require costs to be _over_ £10,000


davidrapson:
[Prevent users from requesting exactly £10,000 through the Funding Proposal form](https://trello.com/c/FMZmRm9g/910-prevent-users-from-requesting-exactly-10000-through-the-funding-proposal-form)

PR #2572, opened 2019-10-28
Add standard dashboard

davidrapson:
Thanks to @mattandrews for making this so easy. The changes here as mostly just cleanup and a few tweaks but this pretty much worked out of the box.

### DIFF
--- a/controllers/tools/applications.js
+++ b/controllers/tools/applications.js
@@ -62,65 +62,6 @@ function applicationsByDay(responses) {
     });
 }
 
-function minMaxAvg(arr) {
-    const sorted = arr.slice().sort((a, b) => a - b);
-    return {
-        lowest: sorted[0] || 0,
-        highest: sorted[sorted.length - 1] || 0,
-        average: mean(sorted) || 0
-    };
-}
-
-function measureTimeTaken(data) {
-    const appDurations = data.map(row => {
-        const created = moment(row.startedAt);
-        const submitted = moment(row.createdAt);
-        return submitted.diff(created, 'minutes');
-    });
-    let results = minMaxAvg(appDurations);
-
-    // convert the larger amounts to days
-    const minutesToDays = input => input / 60 / 24;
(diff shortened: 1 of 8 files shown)


controllers/tools/applications.js
```diff
@@ -202,14 +195,22 @@ function addCountry(row) {
...
 function getFeedbackDescriptionByAppId(appId) {
     let description;
-    switch (appId) {
-        case 'awards-for-all':
-            description = 'National Lottery Awards for All';
-            break;
-        default:
-            break;
+    if (appId === 'awards-for-all') {
+        description = 'National Lottery Awards for All';
```
mattandrews: Is this just the same mapping as `getApplicationTitle()` above?

davidrapson: It is, but accidentally I think, so made them separate functions.

davidrapson: Oh, on second thoughts it looks like the feedback description _is_ the form title. I can probably merge these then

davidrapson:
@mattandrews Mind giving this another review. Looked at this a bit closer and there was a bug in the query that meant it wasn't correctly separating applications out and did some work to normalise the countries to use the actual `formBuilder` for pending forms. Applications with multiple locations are grouped by the same value we use in the summary (i.e. `uk-wide`)

davidrapson:
[Create Your funding proposal dashboard](https://trello.com/c/cG6fQSlE/912-create-your-funding-proposal-dashboard)

db/models/applications.js
```diff
@@ -77,17 +77,12 @@ class PendingApplication extends Model {
         });
     }
 
-    static findAllByForm(formId, dateRange = {}) {
-        let whereClause = {
-            formId: { [Op.eq]: formId }
-        };
-        if (dateRange.start && dateRange.end) {
-            whereClause = {
```
mattandrews: Not sure how I missed this 🙈

PR #2573, opened 2019-10-28
Fix ampersands issue

mattandrews:
This is to do with how we escape/sanitise strings and then display them again:

image

This change makes sure we mark text inputs as `safe` when displaying/editing them. Probably doesn't cover all cases (the Vue components, eg. budgets and large textareas) might do some weird things, but this fixes a user-reported error of their project name constantly replacing `&` with `&amp;`.


davidrapson:
[Bug: Entering special characters e.g. "&" gets escaped](https://trello.com/c/A6ZYCZ0Y/906-bug-entering-special-characters-eg-gets-escaped)

PR #2574, opened 2019-10-28
Separate email and application css processing

davidrapson:
Spotted an issue with the email css where the `fonts-loaded` processing was conflicting and stripping the fallback font from declarations so that emails were coming through with serif fonts in some email clients (at least Outlook)

image

Rather than fighting with the processing rules needed for the main website styles I've split things out so there are separate processing steps for `style.scss` and `emails.scss`.

For emails we don't need source-maps, and the styles are all inlined so we don't need to minify either. This means that the rules are as follows:

- **style.scss**: Sass processing, Autoprefixer, font loading classes, minification, source-maps
- **emails.css**: Sass processing and autoprefixer only


package.json
```diff
@@ -20,8 +20,9 @@
     "test": "npm run lint && npm run test-unit",
     "test-integration": "npm run start-test & wait-on http://localhost:8090 && npx cypress run",
     "clean": "del-cli $npm_package_config_dist_root/* ./config/assets.json",
-    "build-sass": "node-sass assets/sass --output $npm_package_config_dist_css --source-map $npm_package_config_dist_css",
-    "build-css": "npm run build-sass && postcss $npm_package_config_dist_css/*.css --replace --map",
+    "build-css-style": "node-sass assets/sass/style.scss --output $npm_package_config_dist_css --source-map $npm_package_config_dist_css && postcss $npm_package_config_dist_css/style.css --replace --map",
+    "build-css-emails": "node-sass assets/sass/emails.scss | postcss --no-map -u autoprefixer > $npm_package_config_dist_css/emails.css",
```
davidrapson: The `-u` overrides the config and tells postcss to just use autoprefixer, `--no-map` is needed to disable the default inline sourcemap behaviour. Because we are processing a single file we can pipe the output rather than needing to write files temporarily and reprocess them like we do with the main stylesheet.

PR #2577, opened 2019-10-28
Account cleanup

davidrapson:
1. Remove a bunch of unused breadcrumb code from accounts section (we don't show them here)
2. Add an explicit method for requiring an unactivated account
3. Show generic error using common error style, not alert message

image

### DIFF
--- a/common/authed.js
+++ b/common/authed.js
@@ -75,7 +75,7 @@ function requireActiveUserWithCallback(cb) {
     return (req, res, next) => requireActiveUser(req, res, next, cb);
 }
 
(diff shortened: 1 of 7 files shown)


common/authed.js
```diff
@@ -114,6 +130,7 @@ module.exports = {
     requireNoAuth,
     requireUserAuth,
     requireActiveUser,
+    requiredUnactivatedUser,
```
mattandrews: Nitpicking but should this be called `requireUnactivatedUser`? (eg. no `d`)?

davidrapson: Oh yes! Nitpick away

PR #2579, opened 2019-10-29
Fixup logic for requireUnactivatedUser

davidrapson:
Check for `requireUnactivatedUser` was the wrong way around. Introduced in https://github.com/biglotteryfund/blf-alpha/pull/2577


davidrapson:
Yeah I confused myself, as it's a negative check to require an _un_activated user.

PR #2590, opened 2019-10-30
Increase word count of first question

davidrapson:
https://trello.com/c/m0Aq9zuG/928-update-recommended-word-count

After reviewing the recommendation again the only change we need to make is to increase the word-count of the first question.


davidrapson:
[Update recommended word count](https://trello.com/c/m0Aq9zuG/928-update-recommended-word-count)

PR #2597, opened 2019-10-31
Refactor CMS pages to use new Flexible Content Next 

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/197

Quite pleased with this after @davidrapson's suggestion to split the existing Flexible Content template rather than risk breaking it for existing/older pages.

This change allows the use of two new block types:

### Table of Contents
image

(eg. a block of content which will appear above an auto-generated TOC)

### Child Page List

This will try to respect the CMS user's choice for layout style (eg. photo grid or text link) but will only render the photo grid if possible (eg. if all child pages have images).  

image

This change moves the logic for deciding which layout mode to use into the view macros rather than controller code, which feels more of a sensible home for it. 

image

There are a few slightly thorny things in here like rendering breadcrumbs *inside* the flex content parts, or not wrapping a part in a `<section>` if it's a photo grid (eg. to preserve the above layout), but overall I think this simplifies our templates and makes pagebuilding easier and more explicit for the CMS user.


views/components/child-pages/macro.njk
```diff
@@ -0,0 +1,38 @@
...
+            <li class="section-links__item">
+                <a class="section-links__link" href="{{ page.linkUrl }}">
+                    {{ page.trailText or page.title }}
+                </a>
+            </li>
+        {% endfor %}
+    </ul>
+{% endmacro %}
+
+{% macro showChildPages(children, desiredMode) %}
```
davidrapson: This reads like it returns a boolean, so perhaps call this `childPages` (no `show`)

mattandrews:
Finished refactoring this – could you give it (and [the corresponding API one](https://github.com/biglotteryfund/blf-alpha/pull/2598)) another look, @davidrapson?

PR #2602, opened 2019-10-31
Breadcrumb cleanup

davidrapson:
- Removes `ancestors` from content check (no longer included by CMS)
- Remove single use of `customAncestors`
- Sets section breadcrumbs up front rather than at a distance in `injectBreadcrumbs`

We only show breadcrumbs if there is more than one item in the list so this is pretty safe. The only renaming code in `injectBreadcrumbs` is adding the title which can be—and often is—better handled in individual routers. So eventually we can remove `injectBreadcrumbs` in favour but leaving that for a future PR.

### DIFF
--- a/server.js
+++ b/server.js
@@ -230,8 +230,17 @@ forEach(routes, function(section, sectionId) {
      * Used for determining top-level section for navigation and breadcrumbs
      */
     router.use(function(req, res, next) {
-        res.locals.sectionTitle = req.i18n.__(`global.nav.${sectionId}`);
-        res.locals.sectionUrl = localify(req.i18n.getLocale())(req.baseUrl);
+        const sectionTitle = req.i18n.__(`global.nav.${sectionId}`);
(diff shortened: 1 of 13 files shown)


davidrapson:
Edit: removed all the uses of `injectBreadcrumbs` apart from the ones in `common/index.js` as they need a bit more testing, the rest were easy to remove safely.

server.js
```diff
@@ -230,8 +230,17 @@ forEach(routes, function(section, sectionId) {
...
+        const sectionTitle = req.i18n.__(`global.nav.${sectionId}`);
+        const sectionUrl = localify(req.i18n.getLocale())(req.baseUrl);
+
+        res.locals.sectionTitle = sectionTitle;
+        res.locals.sectionUrl = sectionUrl;
+
+        /**
+         * Set top-level breadcrumb for current section
+         */
+        res.locals.breadcrumbs = [{ label: sectionTitle, url: sectionUrl }];
```
mattandrews: 💯

PR #2611, opened 2019-11-01
Translate bilingual expiry emails

mattandrews:
This was a bit fiddlier than usual as there's no `locale` for this endpoint (just gets called by a robot... an English-speaking one) so we have to default to bilingual for applicants in Wales.

### DIFF
--- a/controllers/apply/emails/expiry-email.njk
+++ b/controllers/apply/emails/expiry-email.njk
@@ -1,6 +1,16 @@
 {% extends "views/layouts/emails.njk" %}
 
 {% block content %}
+
+    {% if isBilingual %}
+        <p>
+            <strong>
+                Bilingual message – please scroll down for Welsh.<br />
+                Neges ddwyieithog – y Gymraeg i’w weld isod.
+            </strong>
+        </p>
+    {% endif %}
+
     <p>Hello,</p>
 
     <p>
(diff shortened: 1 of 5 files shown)


controllers/apply/emails/expiry-email.njk
```diff
@@ -29,7 +39,53 @@
...
+            Best wishes,<br/>
+            The National Lottery Awards for All Team
+        </strong>
     </p>
+
+
+    {% if isBilingual %}
+        <p>
+            <br />
+            --
```
davidrapson: `<hr>`?

controllers/apply/expiry.js
```diff
@@ -78,29 +78,41 @@ async function sendExpiryEmails(req, emailQueue, locale) {
 
     return await Promise.all(
         emailQueue.map(async emailToSend => {
+            const getAppData = field =>
+                get(emailToSend.PendingApplication, `applicationData.${field}`);
+
+            const projectCountry = getAppData('projectCountry');
```
davidrapson: Maybe inline this as it's only used once

mattandrews: It's used again further down for `projectName` (although I'll replace the duplicated calls for `projectCountry`)

PR #2614, opened 2019-11-01
Standard dashboard processing

davidrapson:
In retrospect calling the `formBuilder` on each pending application was a pretty bad idea. It computes the validation response in the constructor and doing this x 2,500 applications is slllooow. Need to do some stress testing on this to confirm that doing the current calculation is still fine, but likely to be significantly less processing.

### DIFF
--- a/controllers/tools/applications.js
+++ b/controllers/tools/applications.js
@@ -19,9 +19,18 @@ const {
 const { getDateRange } = require('./helpers');
 const { DATA_STUDIO_AFA_URL } = require('../../common/secrets');
 
+const awardsForAllFormBuilder = require('../apply/awards-for-all/form');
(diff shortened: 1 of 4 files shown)


davidrapson:
Confirmed this locally by seeding ~2,5000 pending applications (thanks to @NikhilNanjappa work on seeder scripts for making that easy) and saw CPU pegged to 100% and page fail to load even locally. Reworking to _not_ use the form builder loaded instantly.

Also added a small tweak for higher number of applications, to make the ticks use an auto scale, rather than a fixed 10 unit scale

Screenshot 2019-11-04 at 12 09 09

Screenshot 2019-11-04 at 12 10 07

davidrapson:
[Create Your funding proposal internal dashboard](https://trello.com/c/cG6fQSlE/912-create-your-funding-proposal-internal-dashboard)

PR #2615, opened 2019-11-04
Fixup findAllByForm query to include formId

davidrapson:
Extracts query fix from https://github.com/biglotteryfund/blf-alpha/pull/2614 so we can get this out now and give ourselves more time to get the standard dashboard working. Fixes the issue where the `formId` was not included in the query which will cause the awards for all dashboard to report incorrectly once the standard form goes live.

### DIFF
--- a/db/models/applications.js
+++ b/db/models/applications.js
@@ -77,17 +77,12 @@ class PendingApplication extends Model {
         });
     }
 
-    static findAllByForm(formId, dateRange = {}) {
-        let whereClause = {
-            formId: { [Op.eq]: formId }
-        };
-        if (dateRange.start && dateRange.end) {
-            whereClause = {
-                createdAt: { [Op.between]: [dateRange.start, dateRange.end] }
-            };
-        }
+    static findAllByForm(formId, dateRange) {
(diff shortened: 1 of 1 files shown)


db/models/applications.js
```diff
@@ -77,17 +77,12 @@ class PendingApplication extends Model {
...
-    static findAllByForm(formId, dateRange = {}) {
-        let whereClause = {
-            formId: { [Op.eq]: formId }
-        };
-        if (dateRange.start && dateRange.end) {
-            whereClause = {
-                createdAt: { [Op.between]: [dateRange.start, dateRange.end] }
-            };
-        }
+    static findAllByForm(formId, dateRange) {
```
NikhilNanjappa: Is there any chance of `req.query` not containing `start` and `end` properties resulting in undefined value for `dateRange` ? (since the default value for dateRange was removed)

davidrapson: Default value is always passed in here (only called in one place, and has a default value passed if req.query aren't set), but good shout that maybe the default should be in the method. Will come back to this, but OK to leave it as is for now.

PR #2616, opened 2019-11-04
Allow special Funding pages (logo/merchandise) to render sub-templates

mattandrews:
This adds the eventual URL paths for the Logo/Merch pages to the routes so they correctly render their additional logic (eg. logo selection / material order form). 

The dependent change for this is to add the "primary content" for these pages (eg. the paragraphs appearing before the custom code) as Flexible Content to the *old* versions of the pages once this PR is live, as otherwise this content will disappear when these pages start using the new "CMS page" template (which only supports flex content, not segments etc). I can coordinate this when we deploy.

### DIFF
--- a/controllers/routes.js
+++ b/controllers/routes.js
@@ -133,12 +133,25 @@ let funding = {
 // Add Funding subpages in non-prod environments
 // @TODO progress to production when ready
 if (isNotProduction) {
-    funding.pages.push({
-        path: '/*',
-        router: basicContent({
-            cmsPage: true
-        })
-    });
+    funding.pages.push(
+        {
+            path: `/managing-your-grant/promoting-your-project/order-free-plaques-stickers-bunting-and-more`,
+            router: require('./materials')
+        },
+        {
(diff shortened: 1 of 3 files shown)


controllers/routes.js
```diff
@@ -133,12 +133,25 @@ let funding = {
...
-            cmsPage: true
-        })
-    });
+    funding.pages.push(
+        {
+            path: `/managing-your-grant/promoting-your-project/order-free-plaques-stickers-bunting-and-more`,
+            router: require('./materials')
+        },
+        {
+            path: `/managing-your-grant/promoting-your-project/download-our-logo-to-tell-people-about-your-national-lottery-funding`,
```
davidrapson: Do we need to add some redirects for the old locations? Also, worth using this as an excuse to simplify the urls e.g. `managing-your-grant/promoting-your-project/logos`

mattandrews: I was planning to do this when we put these new pages live (soon, this week perhaps) rather than add more conditional checks to the code. Could change the URLs though, will ask Scott!

PR #2618, opened 2019-11-04
Add an animated save button (with delay) to form pages

mattandrews:
Adds some animation to the save button:

save

Tried to avoid hard-coding durations etc where I could but quite fiddly to do between the CSS/JS – looks like using HTML `data-` attributes in CSS is a bit inconsistent and works best with strings, not numbers. Happy enough with this approach though with comments to reflect which bits are interdependent.

If JS fails this will work as normally, and it's off in production (well, the JavaScript which adds the delay/animation classes is off) so we can get a feel for how it works on TEST.

### DIFF
--- a/assets/js/common/forms.js
+++ b/assets/js/common/forms.js
@@ -306,8 +306,11 @@ function init() {
     updateSecondaryNav();
     showLocalSaveWarning();
 
-    // Launch this feature in non-prod envs (for now)
-    if (window.AppConfig.environment !== 'production') {
+    // Launch this feature in non-prod envs (for now) and don't add extra delays for automated tests
+    if (
+        window.AppConfig.environment !== 'production' &&
+        !window.AppConfig.isTestServer
+    ) {
         animateSaveButtons();
     }
 
--- a/assets/js/common/forms.js
+++ b/assets/js/common/forms.js
@@ -239,6 +239,48 @@ function showLocalSaveWarning() {
     }
 }
 
+function animateSaveButtons() {
+    $('.js-save-btn-form').on('submit', function(event) {
+        event.preventDefault();
+
+        const $form = $(this);
+        const $btn = $form.find('.js-save-btn');
+        const $label = $btn.find('.js-save-btn-label');
+
+        // animation is 0.3s long x 3 (with a 0.3s delay, eg. 1.2 total)
+        const cssDotAnimationDuration = 300 * 4;
+        const cssIconAnimationDuration = 300;
+
+        const text = {
+            interstitial: $btn.data('interstitial'),
+            complete: $btn.data('complete')
+        };
+
+        const classes = {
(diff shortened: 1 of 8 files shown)


mattandrews:
[Implement save UI animation on the continue button](https://trello.com/c/3R3WUrvj/934-implement-save-ui-animation-on-the-continue-button)

mattandrews:
And in Welsh:

cy

assets/js/common/forms.js
```diff
@@ -239,6 +239,48 @@ function showLocalSaveWarning() {
     }
 }
 
+function animateSaveButtons() {
+    $('.js-save-btn-form').on('submit', function(event) {
```
davidrapson: Having this on the submit event causes this to override the next/previous links so that they a) start the animation and b) trigger the default form action rather than submitting. Submit even is probably the right one to anchor too but need to add a check to guard against this (probably based on the `nextBtn` / `previousBtn` name)

assets/js/common/forms.js
```diff
@@ -239,6 +239,48 @@ function showLocalSaveWarning() {
...
+
+        window.setTimeout(function() {
+            // Remove this event binding and re-trigger the submit request
+            $form.off(event);
+            $form.submit();
+        }, cssDotAnimationDuration + cssIconAnimationDuration);
+    });
+}
+
 function init() {
```
davidrapson: General comment, at some point we should probably restructure things so that the Vue form code and "non-Vue" form code lives closer together, not a problem for now.

views/components/form-fields-next/macros.njk
```diff
@@ -656,3 +656,22 @@
             {% endif %}
     </div>
 {% endmacro %}
+
+{% macro continueButton() %}
```
davidrapson: Probably prefer to make this a standalone macro, rather than in `form-fields` as it's not related to the fields. Either dedicated macro, or maybe inline it into the step?

mattandrews: Done!

assets/js/common/forms.js
mattandrews: Took a bit of a tweak but sorted this now.

assets/js/common/forms.js
```diff
@@ -306,8 +306,11 @@ function init() {
     updateSecondaryNav();
     showLocalSaveWarning();
 
-    // Launch this feature in non-prod envs (for now)
-    if (window.AppConfig.environment !== 'production') {
+    // Launch this feature in non-prod envs (for now) and don't add extra delays for automated tests
+    if (
+        window.AppConfig.environment !== 'production' &&
+        !window.AppConfig.isTestServer
```
davidrapson: Not sure about this check, Cypress test should probably have the same experience as real users even if that's a touch slower. Does it add a massive delay?

PR #2624, opened 2019-11-04
Force integer input for budget (eg. disallow commas/periods)

mattandrews:
This blocks the input of periods/commas (which is already happening in Chrome but not in Firefox or IE) for the budget tool. Users were confused by this as the current behaviour turns `1,000` into `1` (while leaving `1,000` onscreen).


mattandrews:
[Bug: If able to enter commas or decimals in budget costs field the values are parsed incorrectly.](https://trello.com/c/p7rQuPcM/938-bug-if-able-to-enter-commas-or-decimals-in-budget-costs-field-the-values-are-parsed-incorrectly)

PR #2638, opened 2019-11-06
Load sentry from CDN, inline code-splits

davidrapson:
Takes the first step towards being able to better structure our JS by _feature_ rather than by technology. We already inline `Vue` into the main bundle so the amount of code that was code-split was very small. Switching instead to loading Sentry from the CDN (which is already a more optimised bundle) allows us to inline all code-splits into the main app whilst still retaining a reasonable small bundle. Ran `source-map-explorer` against the bundle to confirm. Our bundle is 90% Vue + jQuery.


davidrapson:
Update: realised I'd missed one of the imports and the combined bundle is still a little bit, going to scratch this for now.

PR #2639, opened 2019-11-06
Add a media object variant for larger screens

mattandrews:
Fixes this layout issue by adding a media object variant which only applies to larger screens.

Before:
image

After:
image


mattandrews:
[Fix positioning of example bank statement for mobile devices](https://trello.com/c/JvLzOHhr/945-fix-positioning-of-example-bank-statement-for-mobile-devices)

PR #2640, opened 2019-11-06
Remove redirects/references to BLF domain

mattandrews:
```
$ dig -type CNAME www.biglotteryfund.org.uk

;; ANSWER SECTION:
www.biglotteryfund.org.uk. 0	IN	CNAME	d1dat5w75fqn9b.cloudfront.net.
```

The above domain now points to its own Cloudfront distribution, which redirects it to the Redirectotron server, which in turn redirects it to https://www.tnlcommunityfund.org.uk meaning the site will never be served directly on www.biglotteryfund.org.uk (and it hasn't been for a while). 

This change removes checks for `www.biglotteryfund.org.uk` in code (eg. to ensure we don't block robots/analytics for that domain – this has been redundant for almost a year now) but it also redirects `apply.tnlcommunityfund.org.uk` to the production domain as we now have DNS changes  to serve that via the main app.

### DIFF
--- a/controllers/domain-redirect.js
+++ b/controllers/domain-redirect.js
@@ -1,8 +1,8 @@
 'use strict';
 
 module.exports = function(req, res, next) {
-    if (req.get('host') === 'www.biglotteryfund.org.uk') {
-        res.redirect(301, `https://www.tnlcommunityfund.org.uk${req.originalUrl}`);
+    if (req.get('host') === 'apply.tnlcommunityfund.org.uk') {
+        res.redirect(301, `https://www.tnlcommunityfund.org.uk/apply`);
     } else {
         next();
     }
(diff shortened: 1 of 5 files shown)


controllers/domain-redirect.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 module.exports = function(req, res, next) {
-    if (req.get('host') === 'www.biglotteryfund.org.uk') {
-        res.redirect(301, `https://www.tnlcommunityfund.org.uk${req.originalUrl}`);
+    if (req.get('host') === 'apply.tnlcommunityfund.org.uk') {
+        res.redirect(301, `https://www.tnlcommunityfund.org.uk/apply`);
```
mattandrews: Thought this made more sense than trying to preserve the original path which would always 404.

PR #2653, opened 2019-11-08
Activation email

davidrapson:
Updates activation email template:

image

We previously had this text as coming from the locale yaml files. The downside of this is that this required sprintf style substitution which becomes unwieldy when needing to reuse substitutions and pass larger numbers of variables. For the time being I've inlined the text into the template directly.


davidrapson:
Screenshot 2019-11-08 at 12 07 50

PR #2656, opened 2019-11-11
User account updates

davidrapson:
- Adds `novalidate` consistently to all user forms
- Adds rate limited error text to locale file (missing translation)
- Introduces a common `backLink` macro


davidrapson:
image

PR #2667, opened 2019-11-11
A11y fixes

mattandrews:
Various fixes from the accessibility report – I've kept the commits granular to demonstrate what each one is doing. 

The only thing that's possibly controversial in here is changing all of the hover/active states for `.btn` to a darker shade. Should be fine now the cookie banner has changed but might be worth having a bit of a click around the site when this makes it onto TEST to verify there aren't any weird uses of that button (eg. on a dark background).

### DIFF
--- a/controllers/apply/standard-proposal/form.js
+++ b/controllers/apply/standard-proposal/form.js
@@ -216,8 +216,8 @@ module.exports = function({ locale = 'en', data = {} } = {}) {
             cy: ''
         }),
         startLabel: localise({
-            en: 'Start',
-            cy: 'Dechrau'
+            en: 'Start your application',
+            cy: 'Dechrau ar eich cais'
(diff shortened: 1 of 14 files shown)


controllers/apply/standard-proposal/form.js
```diff
@@ -216,8 +216,8 @@ module.exports = function({ locale = 'en', data = {} } = {}) {
             cy: ''
         }),
         startLabel: localise({
-            en: 'Start',
-            cy: 'Dechrau'
+            en: 'Start your application',
```
mattandrews: I'm not sure if we're using the word "application" here, but it can't just be this:
image

davidrapson: Should this be "Start your proposal?"

PR #2668, opened 2019-11-13
File upload trim

davidrapson:
Wrote a failing test and then updated file upload code to consistently trim the filename.


davidrapson:
[Bug: Applicant submitted file with a space in it causing a failed submission](https://trello.com/c/TOc0nHOj/968-bug-applicant-submitted-file-with-a-space-in-it-causing-a-failed-submission)

PR #2669, opened 2019-11-13
Consider ContentVersion step optional

davidrapson:
Follow on from https://github.com/biglotteryfund/blf-alpha/pull/2668

We can consider an application "submitted" once the main FormData record has been created. Consider ContentVersion step optional but log an error for investigation if upload fails

💭 welcome if this is preferable to considering a failure attaching the bank statement a hard failure.

### DIFF
--- a/controllers/apply/form-router-next/submission.js
+++ b/controllers/apply/form-router-next/submission.js
@@ -117,7 +117,7 @@ module.exports = function(
                                 applicationId: currentApplication.id,
                                 filename: field.value.filename
                             }).then(versionData => {
-                                return salesforce.contentsVersion({
+                                return salesforce.contentVersion({
                                     recordId: salesforceRecordId,
                                     attachmentName: `${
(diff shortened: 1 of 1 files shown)


controllers/apply/form-router-next/submission.js
```diff
@@ -103,26 +103,39 @@ module.exports = function(
...
+                                });
                             });
                         });
-                    });
 
-                await Promise.all(contentVersionPromises);
+                    await Promise.all(contentVersionPromises);
+                } catch (error) {
+                    logger.error('Error creating ContentVersion', error, {
+                        formId: formId
```
mattandrews: Worth logging the application ID here too to help us investigate? With the recent one I had to just look for submission timestamps which matched the log event date.

davidrapson:
@mattandrews Mind giving this another scan? Tidied up some of the logger code

PR #2670, opened 2019-11-13
Make footer layout match DOM order

davidrapson:
Fixes an accessibility issue with flexbox `order`. Switches to a simple float layout which matches DOM order.

image

More context here https://tink.uk/flexbox-the-keyboard-navigation-disconnect/

### DIFF
--- a/assets/sass/components/_global-footer.scss
+++ b/assets/sass/components/_global-footer.scss
@@ -23,13 +23,13 @@
 
     .global-footer__inner {
         padding: 0;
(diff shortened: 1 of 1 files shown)


davidrapson:
[#a11y Fix tab order - DAC_Illogical_tab_order_issue1 DAC_Illogical_tab_order_issue1](https://trello.com/c/oldUeLck/961-a11y-fix-tab-order-dacillogicaltaborderissue1-dacillogicaltaborderissue1)

PR #2673, opened 2019-11-13
Disambiguate untitled applications

davidrapson:
Disambiguate untitled applications using a suffix

image

### DIFF
--- a/views/components/application-card/macro.njk
+++ b/views/components/application-card/macro.njk
@@ -70,7 +70,14 @@
                    class="btn btn--small">
                     {{ cardCopy.continue }}
                     <span class="u-visually-hidden">
-                        ({{ props.projectName or props.untitledName }})
+                        — {{ props.projectName or props.untitledName }}
+                        {%- if not props.projectName %}
+                            {{ __(
+                                cardCopy.createdAt,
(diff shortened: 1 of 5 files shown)


views/components/application-card/macro.njk
```diff
@@ -58,7 +69,9 @@
                 <a href="{{ localify(props.editUrl) }}"
                    class="btn btn--small">
                     {{ cardCopy.continue }}
-                    <span class="u-visually-hidden">({{ projectName | safe }})</span>
+                    <span class="u-visually-hidden">
+                        ({{ props.projectName or props.untitledName }})
```
mattandrews: I think the initial a11y report was pointing out that a list of untitled applications will always read out "Untitled proposal" at this point, so could we add the datestring into this label?

davidrapson: Good shout, I did wonder that as I was updating this.

davidrapson:
[Add a unique identifier for untitled applications on the dashboard](https://trello.com/c/Lq2afSAj/969-add-a-unique-identifier-for-untitled-applications-on-the-dashboard)

davidrapson:
Added date to assistive label and just waiting on a small bit of welsh copy before I merge this.

PR #2675, opened 2019-11-14
Field types

davidrapson:
We wrote new field types for the Your Funding Proposal form. It makes sense to start using them on the Awards for All form too. The most obvious candidates to try first are the email, phone, and name fields as they are the most direct swap.


controllers/apply/lib/validate-model.js
```diff
@@ -99,7 +99,9 @@ module.exports = function validateModel(formModel) {
             .required()
     });
 
-    const validationResult = Joi.validate(formModel, formSchema);
+    const validationResult = Joi.validate(formModel, formSchema, {
+        allowUnknown: true
```
davidrapson: I've added this because the new field classes have methods and class properties on them that it doesn't make sense to include in this validation. The class field types have their own validation so we can slowly move away from needing this rough schema validation, we don't use it at all on the funding proposal form.

PR #2678, opened 2019-11-14
Remove old questions list

davidrapson:
We don't link to this screen and we're currently reviewing how we present this information. We don't know exactly what form it will take, but it's unlikely to be the same format as this page. We can always reinstate parts of the code if we need it, but it currently adds a fair bit of complexity to the form for a feature we're not using.


deploy/bootstrap.sh
```diff
@@ -42,11 +42,6 @@ server_config=/var/www/biglotteryfund/deploy/server.conf
 APP_ENV_PLACEHOLDER="APP_ENV"
 sed -i "s|$APP_ENV_PLACEHOLDER|$APP_ENV|g" $server_config
 
-# Fix phantomjs install for Ubuntu
```
davidrapson: Noticed this was stilling being run in `bootstrap` despite the fact we disabled PDF generation a while a go and no longer even install phantomjs

mattandrews:
I'm slightly wary of deleting this code just because of the complexity (eg. the stuff filtering the possibly-optional fields) that it seems like we'll still want in whatever we replace it with? Would prefer to wait until we know what form the new questions page will take before we delete this stuff, rather than pre-emptively doing it then having to re-add it, I think – surely we'll still want a way to mark out possibly-optional fields, even if we change the visual formatting?

mattandrews:
(good call on deleting the phantomJS code though!)

davidrapson:
Based on conversations yesterday definitely erring on the side of any replacement page taking a different direction where this code would be redundant, but happy to hold for now 👍

PR #2681, opened 2019-11-14
Add temporary warning text to RC form in advance of closure

mattandrews:
Adds a warning to RC form pages to clarify the form will change.

This is hard-coded which I think is acceptable given we'll shortly delete all of this code, and that it's not used for any other forms.

image
image


mattandrews:
[Add a message to existing Reaching Communities form to explain the form will soon change](https://trello.com/c/JlhIyaDL/971-add-a-message-to-existing-reaching-communities-form-to-explain-the-form-will-soon-change)

controllers/apply/reaching-communities/startpage.njk
```diff
@@ -12,13 +12,10 @@
                 suffix = copy.description | safe
             ) }}
 
-            <h2>Find out how we can help you</h2>
```
davidrapson: Did you mean to remove this part?

mattandrews: Oops, no! Thanks.

PR #2683, opened 2019-11-14
Boost radio buttons

mattandrews:
Use GDS-style radio buttons ([as defined here](https://design-system.service.gov.uk/components/radios/)) to make ours easier to click:

### Before:
image

### After:
image

CSS is taken directly from GDS, then modified a bit (eg. uses variables to allow changing the button sizes, uses our default highlight colour etc).

Main change is the HTML structure for the radios themselves as much of the (CSS) logic here depends upon the input/radio being siblings.

### DIFF
--- a/assets/sass/components/_forms.scss
+++ b/assets/sass/components/_forms.scss
@@ -179,7 +179,9 @@
    ========================================================================= */
 
 $inputPadding: 6px 12px;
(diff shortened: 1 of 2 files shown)


mattandrews:
Good call – checking the same GDS docs reveals they have a similar thing for checkboxes so I've just added that in:

image

PR #2684, opened 2019-11-14
Standard single choice

davidrapson:
image

Updates project countries to be a single value radio option whilst supporting pending proposals and existing Salesforce schema that expects an array for the selection.


davidrapson:
[Only allow single country selection for England launch](https://trello.com/c/Vr7Ze2eY/985-only-allow-single-country-selection-for-england-launch)

PR #2687, opened 2019-11-15
Fix tools (part two)

mattandrews:
Add back the country picker to AFA dashboard:

image

### DIFF
--- a/controllers/tools/views/applications.njk
+++ b/controllers/tools/views/applications.njk
@@ -8,6 +8,29 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.8.0/Chart.bundle.min.js" crossorigin="anonymous"></script>
 {% endblock %}
 
+{% set countries = [
+    {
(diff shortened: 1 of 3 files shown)


controllers/tools/views/applications.njk
```diff
@@ -28,7 +51,17 @@
             </p>
 
             <form action="" method="get" class="tools-filter-form">
+                <div class="tools-filter-form__item">
+                    <label for="country" class="ff-label">Country</label>
 
+                    <select name="country" class="ff-text">
```
davidrapson: Should this be `ff-select` rather than `ff-text`?

mattandrews: Probably, but that class doesn't exist (though it's used in the templates a bit)!

Looks like this without `.ff-text`
image

davidrapson: Hah, I guess it's just there by convention then. Go without the class, preferable to using `ff-text` on something unexpected.

PR #2688, opened 2019-11-15
Inject copy cleanup

davidrapson:
Been finding that the way we use `injectCopy` leads to a bunch of indirection and setting locals that are either unused or overridden later. We can achieve the same thing with direct `req.i18n.__` calls. Starting with the apply section.


controllers/apply/form-router-next/index.js
```diff
@@ -152,9 +78,76 @@ function initFormRouter({
                 });
             }
         }),
-        handleMultipartFormData,
-        injectCopy('applyNext'),
-        setCommonLocals,
+        function(req, res, next) {
+            function needsMultipart() {
```
mattandrews: Could you reinstate the comment that went with this? It's not always obvious why we're doing this stuff so would be worth keeping it in to explain the *why*

davidrapson: Good shout, didn't mean to loose that

PR #2679, opened 2019-11-14
Launch Your Funding Proposal form in England

davidrapson:
Launch Your Funding Proposal form in England. There is a _tonne_ of code we get to delete once this happens, but for now this PR:

- Enables England as an option in production
- Redirects the old enquiry form to the apply section
- Adds new country message

image

image

PR #2690, opened 2019-11-15
Update country message for England launch

davidrapson:
image


davidrapson:
[Only allow single country selection for England launch](https://trello.com/c/Vr7Ze2eY/985-only-allow-single-country-selection-for-england-launch)

davidrapson:
Just clocked this needs to go live at launch so have merged this commit into https://github.com/biglotteryfund/blf-alpha/pull/2679

PR #2699, opened 2019-11-18
Split out facet field styles

davidrapson:
This is a bit of yak shaving that attempts to solve a couple of things:

a) Removes coupling to old `form-field` macros for facet filters, making deleting the old macros one we launch the new standard form in England simpler
b) Resolve some dual-purpose style issues between forms and facet filters. Most notable since https://github.com/biglotteryfund/blf-alpha/pull/2683 where both facet radios and form radios used the base `ff-choice` class but displayed radically differently based on their context

As a result of this reworking it also fixes an accessibility issue on the insights documents page where fields were missing labels. Fixed by combining fields into a single facet group.

Summary: Splits out facet field styles into their own set of classes to isolate them from core form field styles.


controllers/digital-fund/views/assistance.njk
```diff
@@ -2,7 +2,7 @@
 {% from "components/hero.njk" import hero with context %}
 {% from "components/icons.njk" import iconTick %}
 {% from "components/breadcrumb-trail/macro.njk" import breadcrumbTrail %}
-{% from "components/form-fields/macros.njk" import formErrors with context %}
+{% from "components/form-fields-next/macros.njk" import formErrors %}
```
mattandrews: Think this will need updating after the other branch is merged

PR #2700, opened 2019-11-19
Add modal dialog with session expiry warnings

mattandrews:
This uses the GDS in-progress modal component ([sourced here](https://github.com/alphagov/govuk-design-system-backlog/issues/30) - demo [here](https://components.publishing.service.gov.uk/component-guide/modal_dialogue)) to add an accessible modal to the site. 

We can use these anywhere on the site by adding the following to a page:

```
{% block modals %}
    {% call modal('your-id-here, title = 'Some title here') %}
        <p>Your content goes here</p>
    {% endcall %}
{% endblock %}
```

... and then a trigger can open the modal like so:

    <button data-toggle="modal" data-target="your-id-here">Show modal</button>

Or directly in JavaScript via:

    import modal from './modal';
    modal.triggerModal('your-id-here');

This change adds the session expiry warnings like so – you'll see this message when you have 10 minutes remaining:

image

Followed by this unclosable dialog when the total session (two hours) has expired:

image

For now I'm only enabling this on non-prod environments so we can test it and confirm it works as expected.

### DIFF
--- a/assets/sass/components/_modal.scss
+++ b/assets/sass/components/_modal.scss
@@ -1,0 +1,121 @@
+$govuk-modal-margin: 20px;
+$govuk-modal-close-button-size: 44px;
+$govuk-modal-z-index: 1000;
+
+.gem-c-modal-dialogue {
+    display: none;
+    position: fixed;
+    z-index: $govuk-modal-z-index;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    overflow-x: hidden;
+    overflow-y: scroll;
+    outline: 0;
+}
+
+.gem-c-modal-dialogue__box {
+    display: block;
+    position: fixed;
+    background: white;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    margin: 0;
+    padding: 0;
+    overflow-y: auto;
+    border: 0;
+
+    @include mq('medium') {
+        position: relative;
+        top: inherit;
+        right: inherit;
+        bottom: inherit;
+        left: inherit;
+        width: auto;
+        max-width: 700px;
+        height: auto;
+        margin: 20px;
+        border: 1px solid black;
+    }
+}
+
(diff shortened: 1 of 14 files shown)


assets/sass/components/_modal.scss
```diff
@@ -0,0 +1,99 @@
...
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    margin: 0;
+    padding: 0;
+    overflow-y: auto;
+    border: 0;
```
mattandrews: Looks like this (eg. fullscreen) on mobile:

image

config/default.json
```diff
@@ -29,7 +29,8 @@
     "enableSeeders": false
   },
   "awardsForAll": {
-    "enableExpiration": true
+    "enableExpiration": true,
+    "enableSessionExpiryWarning": true
```
davidrapson: This applies to the whole website / all forms right? If so, I'd put this flag under the common `features` config rather than scoping to awards for all.

config/locales/en.yml
```diff
@@ -1692,6 +1692,23 @@ applyNext:
...
+      body: >
+        <p>We’ll log you out if you don’t respond in ten minutes.</p>
+        <p>We do this to keep your information secure.</p>
+      actions:
+        continue: Continue working
+        logOut: Log out
+    expired:
+      title: Sorry, we’ve had to log you out
+      body: >
+        <p>You’ve not been filling in this form for over two hours. So we’ve had to log you out to keep your personal information safe.</p>
```
davidrapson: Should the "two hours" part be dynamic based on the session expiry setting? Or is it easier to just keep this in sync if that changes.

mattandrews: I thought about this but it'd involve passing around config items into templates, JS files, etc – thought it was more pragmatic to just hardcode it and change it as/when.

mattandrews:
Updated – can you give it another quick check @davidrapson?

PR #2707, opened 2019-11-20
Pass a dynamic form short ID for Hotjar tagging 

mattandrews:
Allows us to distinguish between each application form.

I tested each tag event and confirmed they all work as expected.


mattandrews:
Will fix the merge conflicts now (renamed files).

PR #2708, opened 2019-11-21
Fix ampersands

mattandrews:
The changes [here](https://github.com/biglotteryfund/blf-alpha/pull/2573) have introduced issues where `&amp;` is being stored in the database and therefore persisted into Salesforce.

This is because the `DOMPurify.sanitize()` method turns `&` into `&amp;` – I tried using the HTML config (`{USE_PROFILES: {html: true}}`) to no effect so have resorted to manually replacing `&amp;` with `&` again.

I also spotted that the use of ` | safe` in form inputs broke things when quote marks are used (not uncommon in project names) – I've swapped this for ` | escape` which does something similar but means quotes won't break inputs.


davidrapson:
[Review issues with ampersand characters not being escaped properly when sending to GMS](https://trello.com/c/uBVugxJi/990-review-issues-with-ampersand-characters-not-being-escaped-properly-when-sending-to-gms)

PR #2709, opened 2019-11-21
Show a native notification on app reload

mattandrews:
Scratching an itch really, but it might be handy to notify us when the app has (re)loaded, to save us all getting RSI clicking refresh when waiting for compilation to finish.

image

### DIFF
--- a/bin/www
+++ b/bin/www
@@ -3,6 +3,10 @@
 const http = require('http');
 const config = require('config');
 
+const NODE_ENV = config.util.getEnv('NODE_ENV');
+
+const notifier = NODE_ENV === 'development' ? require('node-notifier') : false;
+
 const logger = require('../common/logger').child({
     service: 'www'
 });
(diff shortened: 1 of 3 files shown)


bin/www
```diff
@@ -68,9 +72,18 @@ function startServer() {
...
     });
     server.on('listening', function() {
-        logger.debug(
-            `App now running on ${describeAddressBinding(server.address())}`
-        );
+        const msg = `App now running on ${describeAddressBinding(
+            server.address()
+        )}`;
+        logger.debug(msg);
+        if (notifier) {
```
davidrapson: Could maybe write this as:

```js

if (NODE_ENV === 'development') {
  require('node-notifier').notify({…})
}
```

bin/www
```diff
@@ -68,9 +72,18 @@ function startServer() {
...
+            notifier.notify({
+                title: 'blf-alpha',
```
davidrapson: Wonder if we should renamed references to `blf-alpha` over to `tnlcommunityfund` at some point?

davidrapson:
(as discussed could also do this with a even listener in `nodemon.json`

PR #2712, opened 2019-11-21
Session expiry tweaks

mattandrews:
A few things in here:

- Adds session expiry message to the summary screen (not the warning, just the "You've been logged out" version)
- Disables the "Are you sure you wish to leave this page?" warning when clicking on a link within the session expired modal (eg. because this warning is redundant)
- Updates the expiry message to be more generic (rather than "You’ve not been filling in this form for over two hours" which won't always be true)
- Shows a dynamic timestamp for when your session will end
- Adds `req.session.touch()` before extending session, which seems to render the expiry/maxage more reliably, and is what the method is meant to be used for


mattandrews:
The one thing I wanted to do here but couldn't is to show a status message when you've been logged out. I can't think of a way to do this though – the `requireAuthed` methods all kick you to `/user/login` if you're not authenticated... but how would we know if the user was previously authed when they clicked that link?

We only know this on the form submission step because we check for a POST of data by an unauthed user, but this will now become impossible as the modal prevents submission of the form by unauthed users.

assets/js/common/forms.js
```diff
@@ -17,30 +17,8 @@ function handleBeforeUnload(e) {
     return confirmationMessage; // Gecko, WebKit, Chrome <34
 }
 
-function handleAbandonmentMessage(formEl) {
```
davidrapson: 💯

assets/js/common/session.js
```diff
@@ -36,6 +37,8 @@ function handleSessionExpiration() {
             isAuthenticated = false;
             clearSessionExpiryWarningTimer();
             trackEvent('Session', 'Warning', 'Timeout reached');
+            // Prevent the page abandonment warning from showing
+            forms.removeBeforeUnload();
```
davidrapson: This is a touch awkward, suggests this session code should probably live in `common/forms.js` instead.

assets/js/common/session.js
```diff
@@ -69,15 +72,23 @@ function handleSessionExpiration() {
 }
 
 function getUserSession() {
-    return $.ajax({
+    const sessionCheck = $.ajax({
```
davidrapson: Probably don't need the intermediary variable. Could be `return $.ajax({}).then(response => {…})`

mattandrews: I tried this first and the other uses of `getUserSession()` stopped working... not sure jQuery returns the right thing if passed like that?

davidrapson: That's surprising, but 👌

mattandrews: Ah wait it's because I didn't `return response` inside the first `then()` call. Will add!

assets/js/common/session.js
mattandrews: Done!

PR #2713, opened 2019-11-21
Separate project start and end date fields

davidrapson:
image

This PR explores splitting the project start and end date into two fields. In broad terms this is much simpler: results in a nicer user experience and can be modelled using the regular date field and will allow us to remove the custom extension and macros for the combined date range field. However it needs a little thought around how we handle transforming application date into the new format.

Things still to do or consider:

- [x] Support new fields behind a feature flag on awards for all form
- [x] Update salesforce transform function to account for new schema
- [x] Determine approach to migrate or transform pending applications to new format


controllers/apply/lib/joi-extensions/date-parts.js
```diff
@@ -143,39 +142,6 @@ module.exports = function dateParts(joi) {
                         );
                     }
                 }
-            },
-            {
```
davidrapson: Worked out we can do this validation with less specific `minDate` and `maxDate` rules, so if this seems sensible might extract parts of this out into a separate PR that can be merged earlier.

davidrapson:
[Implement updates to project date question to reduce number of warnings](https://trello.com/c/wcEVgoi2/977-implement-updates-to-project-date-question-to-reduce-number-of-warnings)

controllers/apply/awards-for-all/transforms.test.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-env jest */
+'use strict';
+
+const { mockResponse } = require('./mocks');
+const transform = require('./transform');
+
+test('should transform previous application schemas', function() {
```
davidrapson: Added these tests for a transform step which at least confirms the result we'd need. Not yet _calling_ this from anywhere, still need to work out the best place to put things.

controllers/apply/awards-for-all/transforms.js
```diff
@@ -0,0 +1,42 @@
...
+const get = require('lodash/fp/get');
+const Sentry = require('@sentry/node');
+
+const logger = require('../../../common/logger').child({
+    service: 'transform'
+});
+
+function transformProjectDateRange(applicationData) {
+    if (has('projectDateRange')(applicationData)) {
+        logger.info('Transforming projectDateRange');
```
mattandrews: 💯

controllers/apply/awards-for-all/transforms.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    return applicationData;
+}
+
+function transform(applicationData) {
+    // Add additional transforms here
+    if (config.get('awardsForAll.enableNewDateRange') === true) {
+        return transformProjectDateRange(applicationData);
```
mattandrews: Just to check, we're calling this like so:
```
    const originalData = application.applicationData || {};
    const data = transform(originalData);
```

This means it'll modify the original `application.applicationData` too, right? Probably okay in this case but is it worth cloning the source data here before the transforms?

davidrapson: It shouldn't do as I'm passing it in separately, but will double check.

davidrapson:
@mattandrews Yeah a few missing translations still. This PR is also now too big for my liking, made a fair few refactorings along the way so I'm going to take the time to extract out a few smaller PRs.

davidrapson:
I've split out two pull requests:

- https://github.com/biglotteryfund/blf-alpha/pull/2746
- https://github.com/biglotteryfund/blf-alpha/pull/2745

Both of which can me merged ahead of time. Once that's done I'll clean up this PR, or raise a new pull request for a fresh review with this more focussed set of changes.

davidrapson:
Closed in favour of https://github.com/biglotteryfund/blf-alpha/pull/2761

Issue #1856, opened 2019-04-09
CMS enhancements


mattandrews:
## Needs review
- [x] Update Craft to newest version (and Redactor plugin) - https://github.com/biglotteryfund/craft-dev/pull/204
- [ ] Massive images being uploaded for segments or [within content areas](https://www.tnlcommunityfund.org.uk/about/the-national-lotterys-25th-birthday) and not resized - https://github.com/biglotteryfund/craft-dev/pull/200

## Stuff we should do / fix
- [x] Make hero images optional on research pages (don't attempt render the page with an empty hero image)
- [x] When rendering in preview mode, avoid showing 500 errors (eg. for missing fields) and instead show an explanation about what's missing (or try to render a page regardless)
- [ ] Move all URL localising to the CMS rather than the webapp
- [ ] Revise the use of expiry dates to indicate funding programme closure (see https://github.com/biglotteryfund/craft-dev/pull/199)

## Not sure
- [ ] **Make it easier when choosing hero images to preview the image in question**
Would be nice, but doesn't look like Craft offers any way to do this

- [ ] **Image editor doesn't allow resizing (just crop/rotate)**
I think this is because we don't define any image crop sizes (we just let Imgix handle this). Not sure it's really needed?

- [ ] **Is it possible to make content "invisible" (eg. not linked automatically but still linkable) like press releases? Should it be?**
This has come up once and probably more trouble than it's worth for a one-off use case.

- [ ] **Make it easier to view/revert to entry versions > 10 entries ago (https://github.com/workingconcept/versioneer-craft-plugin)**
This has only been necessary once (and was a bit of a faff) so not sure we need it? This would also require adding this new field to every section.

## Done
- [x] Allow previewing the homepage for changes to tiles - https://github.com/biglotteryfund/blf-alpha/pull/2913
- [x] Allow blank slugs for custom open graph metadata to be anywhere  (not just first) - https://github.com/biglotteryfund/craft-dev/pull/202
- [x] Make it possible to make child pages of insight pages - https://github.com/biglotteryfund/craft-dev/pull/203

PR #2715, opened 2019-11-22
Fix absolute URLs when localised

mattandrews:
I need to add an absolute URL (eg. to CNL25 site) to the homepage, but the `localise()` function gets called on all the paths provided here:

image

This change means that absolute URLs won't be localised, otherwise you end up with something weird where we try to add `/welsh` to a full URL.

### DIFF
--- a/common/urls.js
+++ b/common/urls.js
@@ -4,6 +4,8 @@ const querystring = require('querystring');
 
 const WELSH_REGEX = /^\/welsh(\/|$)/;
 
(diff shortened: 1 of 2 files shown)


mattandrews:
True, we have `uriForLocale` at the CMS end, though we'd need the same fix there as it doesn't check for absolute URLs either. Maybe let's merge this now (I need it for tomorrow morning) and add this to our list of [CMS enhancements](https://github.com/biglotteryfund/blf-alpha/issues/1856
) to make?

PR #2725, opened 2019-11-26
Add some redirects for constantly-requested logo images

mattandrews:
This has bothered me for ages:

image

Autodiscover has been solved by switching off the old domain redirect service (which was forwarding Outlook clients requesting `https://tnlcommunityfund.org.uk/autodiscover/autodiscover.xml` onto `https://www.tnlcommunityfund.org.uk/autodiscover/autodiscover.xml`) but those two logo requests are a huge chunk of our 404 stats and get tons of requests every day:

image

I can't prove it as the logs don't include referrers, and searching the web for references to these file paths only turns up a few small local blogs/sites which are unlikely to be sending 1000 reqs/day, so my theory is that these images are hotlinked somewhere internally (eg. the intranet or some other tool), which I've asked IT about.

In the meantime, this change redirects them to the current logo so we'll stop seeing so many 404s, and so whatever service is hotlinking them stands a chance of getting an expected response back.

### DIFF
--- a/controllers/__snapshots__/aliases.test.js.snap
+++ b/controllers/__snapshots__/aliases.test.js.snap
@@ -6402,5 +6402,85 @@ Array [
     "from": "/welsh/wales/about/customer-service/supplier-zone/contracts-finder",
     "to": "/welsh/about/customer-service/supplier-zone",
   },
+  Object {
+    "from": "/-/media/Images/Logos/JPEGs/hi_big_e_min_blue.jpg",
+    "to": "/assets/images/logos/tnlcf/monolingual/colour.png",
+  },
+  Object {
+    "from": "/welsh/-/media/Images/Logos/JPEGs/hi_big_e_min_blue.jpg",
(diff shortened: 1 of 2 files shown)


controllers/__snapshots__/aliases.test.js.snap
```diff
@@ -6402,5 +6402,85 @@ Array [
...
+  Object {
+    "from": "/welsh/-/media/Images/Logos/JPEGs/hi_big_e_min_blue.jpg",
+    "to": "/welsh/assets/images/logos/tnlcf/monolingual/colour.png",
+  },
+  Object {
+    "from": "/england/-/media/Images/Logos/JPEGs/hi_big_e_min_blue.jpg",
+    "to": "/assets/images/logos/tnlcf/monolingual/colour.png",
+  },
+  Object {
+    "from": "/welsh/england/-/media/Images/Logos/JPEGs/hi_big_e_min_blue.jpg",
```
davidrapson: We probably don't need all the regional redirects here. One for a bit of future clean up I reckon to make the region stuff opt-in. Fine for now though.

PR #2726, opened 2019-11-28
User dashboard tool

mattandrews:
This change creates a searchable dashboard of registered users, restricted to Fund staff, and further restricted to users who are members of two internal Active Directory groups (Advice team and Digital team).

This last bit was achieved by modifying the manifest config for the Azure AD auth app and specifying `"groupMembershipClaims": "All"`. This gives back an array of the user's group IDs, which I've reverse engineered (eg. searched in Azure) and found the two we're interested in, and verified it works as expected (eg. non-members can't access this new endpoint). I think this is a neater way of restricting tool access as it doesn't involve complex Azure config (eg. we'd otherwise have to initialise a second Azure auth strategy with a different config to look up members of a group) or having to do maintenance of adding/removing users – it's already tied to their existing membership.

The dashboard itself looks like this:

image

Activations or password resets are confirmation-limited:

image

We show pending applications with a note to look in GMS for submitted:

image


mattandrews:
I've also completed a Data Protection update on this tool in our Information Asset Register – may need final confirmation that the document itself is valid but in the meantime this code can be reviewed.

controllers/tools/views/user-dashboard.njk
```diff
@@ -0,0 +1,202 @@
+{% extends "layouts/main.njk" %}
+{% from "components/breadcrumb-trail/macro.njk" import breadcrumbTrail %}
+{% from "components/staff-status/macro.njk" import staffStatus with context %}
+
+{% block extraHead %}
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
```
davidrapson: Any way we can avoid using bootstrap in _addition_ to our own styles? Cool with coming back to this to allow us to get something workable out, but good to avoid it if we can.

controllers/tools/views/user-dashboard.njk
```diff
@@ -0,0 +1,202 @@
...
+            </thead>
+            <tbody>
+            {% for user in users %}
+                <tr>
+                    <td>
+                        {{ user.username }}<br />
+                        <small>
+                            <strong>User ID</strong>: {{ user.id }} <br />
+                            <strong>Registered</strong>: {{ dateItem(user.createdAt) }} <br />
+                            <strong>Last login</strong>: {{ dateItem(user.updatedAt) }}
```
davidrapson: Don't think `updatedAt` is the same as last login. Most likely when the record was last updated via  password reset, change of email etc.

controllers/tools/views/user-dashboard.njk
mattandrews: Think for now it'd be more effort than it's worth (until we know how this tool gets used) but agree it'd be nice to avoid it if we can.

PR #2581, opened 2019-10-29
Acccount activation without authentication

NikhilNanjappa:
https://trello.com/c/BHW8T4OO/913-dont-require-users-to-be-logged-in-when-activating-their-accounts.

PR #2727, opened 2019-11-28
Redirect already-activated users from activation page

mattandrews:
Found this kept happening to me when testing deactivation. You shouldn't be able to get to the activation page once you're active!

### DIFF
--- a/controllers/user/activate.js
+++ b/controllers/user/activate.js
@@ -5,7 +5,8 @@ const express = require('express');
 const { Users } = require('../../db/models');
 const {
     redirectUrlWithFallback,
-    requireUnactivatedUser
+    requireUnactivatedUser,
+    isActivated
 } = require('../../common/authed');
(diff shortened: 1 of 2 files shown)


controllers/user/activate.js
```diff
@@ -24,6 +25,10 @@ router
     .route('/')
     .all(injectCopy('user.activate'))
     .get(async function(req, res) {
+        // Prevent activated users from persisting on this page
+        if (req.user && isActivated(req.user)) {
```
davidrapson: Is this not the same as what `requireUnactivatedUser` does? Which is currently only on the POST endpoint, could we add that to `.all` instead?

mattandrews: No – we used to have that here but removed it so that unauthed users could still activate (eg. when using another device): https://github.com/biglotteryfund/blf-alpha/pull/2581

PR #2736, opened 2019-12-02
Call gov.uk Notify API for user emails in TEST

mattandrews:
Kind of an experiment - swapping out SES for [gov.uk Notify](https://www.notifications.service.gov.uk).

The goal here is to improve deliverability of emails generally, but particularly for account activations. We get some nice delivery stats baked in:

image

image

... and we can even send SMS messages very easily if we wanted to. Quite nice to have this future option!

## Tradeoffs of this approach: 

- We can't do a huge amount of customisation of the templates (even [bold/italics](https://govuk.zendesk.com/hc/en-us/requests/3864744), currently)

- Translation is a bit fiddly – we can use variables in templates like this:

    `((under18??Please get your application signed by a parent or guardian.)) `

    ... but they can only be `yes` or `no`. I guess we could pass `isWelsh` and co-locate the copy, but right now I've just made language-specific templates for each email, eg:

    image

- The address the emails are sent from appears to clients as `The National Lottery Community Fund <the.national.lottery.community.fund@notifications.service.gov.uk>`, though we can customise the reply-to address.

- We have to configure emails and their variables using a web-based dashboard, although the UI for this is really nice so this is probably a decent trade-off too.

- We can only have one "branding" for all our emails – at the moment it's our monolingual logo, which we can't directly control (we have to submit a request for it to be changed, which to be fair took less than 20 minutes last time). We probably have to use a bilingual logo as we can't embed images or override the brand otherwise.

I'd like to merge this onto `TEST` and see how it feels using it. If we're happy with it we could set up a test endpoint for users we know have had issues receiving our emails in the past, and get them to visit it and send an email so we can confirm they receive it.


mattandrews:
(tests are failing currently due to missing API credentials – will fix!)

mattandrews:
Can't merge this until we agree to the Gov.uk T&Cs which requires someone within the Fund with "the authority to sign contracts for The National Lottery Community Fund" to approve it – I'm chasing folks now for this.

mattandrews:
Closing this for now as we're hovering around 96% activated users pretty consistently now, and completing the DPIA documentation for this may end up being more complex than it's worth doing for a test.

PR #2740, opened 2019-12-03
Make location description mandatory

davidrapson:
Makes project location description mandatory on the your funding proposal form.

image

In order to make it clear to anyone with a pending form that there is a new required field we're including the field in the featured errors list.

image

Needed to make a change to the featured logic to support an `includeBase` flag which allows us to specify if we want include the `base` message type when determining if an error is featured.


davidrapson:
[Explore gaining more specific location information on the 'Your Funding Proposal' form to support funding officer allocation](https://trello.com/c/fEIlB2Ub/1002-explore-gaining-more-specific-location-information-on-the-your-funding-proposal-form-to-support-funding-officer-allocation)

PR #2743, opened 2019-12-04
Prefer noindex over robots.txt

davidrapson:
Started doing this in order to merge some archival code that was spread across multiple files, in the process I was reminded of this from Google's docs.

> **You should not use robots.txt as a means to hide your web pages from Google Search results.** This is because, if other pages point to your page with descriptive text, your page could still be indexed without visiting the page. If you want to block your page from search results, use another method such as password protection or a noindex directive.

https://support.google.com/webmasters/answer/6062608

So basically this switches to using `noindex` for pages we genuinely want to be excluded and `robots.txt` for the handful where we just want traffic management.


davidrapson:
Removed grants from the list too, let's see how we go.

PR #2744, opened 2019-12-05
Process map

mattandrews:
Adds what I'm calling a "process map" to the Standard form.

The content itself is pending so I'm after a tyre-kicking of the CSS really as that's the bulk of the work here.

Depending on usage there are a few scenarios where this can look weird (eg. packing too much text into the circles/arrows, or adding too many items), but for the most part I'm fairly happy with this.

## Wider screens
image

## Smaller screens
image

### DIFF
--- a/assets/sass/components/_programmes.scss
+++ b/assets/sass/components/_programmes.scss
@@ -118,8 +118,10 @@
    ========================================================================= */
 
 .process-map {
-    $circleWidth: 75%;
-    $circleWidthSmaller: 60%;
+    $circleWidthLarge: 75%;
+    $circleWidthMedium: 60%;
+    $circleWidthSmall: 40%;
+    $midSizeBreakpoint: 500px;
     $arrowSize: 7px;
     $arrowLineHeight: 10px;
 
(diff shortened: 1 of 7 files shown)


views/components/content-box/macro.njk
```diff
@@ -1,6 +1,6 @@
-{% macro contentBox(id = false, isConstrained = true) %}
+{% macro contentBox(id = false, isConstrained = true, skipProse = false) %}
     <section class="content-box{% if isConstrained %} u-inner-wide-only{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
-        <div class="s-prose u-constrained-content-wide">
+        <div class="{% if not skipProse %}s-prose {% endif %}u-constrained-content-wide">
```
mattandrews: This was to allow us to put things in content boxes without inheriting all the `s-prose` styles (eg. all the list resets, which otherwise break this new component).

assets/sass/components/_programmes.scss
```diff
@@ -113,6 +113,237 @@
     }
 }
 
+/* =========================================================================
+   Process Map
+   ========================================================================= */
+
+.process-map {
```
davidrapson: I'd probably put this in `_applications.scss` as it's more related to that, but pretty subjective so go with your gut.

views/components/content-box/macro.njk
davidrapson: Good call. I'd be just hard-coding the content box classes if I needed to skip this, but a flag makes sense 👍

views/components/process-map/examples.njk
```diff
@@ -0,0 +1,105 @@
+{% from "components/styleguide.njk" import sgExample %}
```
davidrapson: 😍 😍 😍

mattandrews:
Good shout on that breakpoint, have tweaked it now:

image

Could you give it one more look @davidrapson?

PR #2747, opened 2019-12-06
Extract more awards for all fields

davidrapson:
Continues the process of extracting more field definitions into single files. Does everything aside from beneficiaries, terms, and date of birth fields (will update after https://github.com/biglotteryfund/blf-alpha/pull/2746 is merged)


controllers/apply/awards-for-all/fields/main-contact-phone.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+const PhoneField = require('../../lib/field-types/phone');
+
+module.exports = function(locale) {
```
davidrapson: This looks a little silly, but felt worth being consistent even if it means the definition is small.

controllers/apply/lib/field-types/field.js
```diff
@@ -30,7 +29,11 @@ class Field {
 
         this.type = props.type ? props.type : this.getType();
 
-        this.attributes = defaults(this.defaultAttributes(), props.attributes);
+        this.attributes = Object.assign(
```
davidrapson: `defaults` has this behaviour: "Once a property is set, additional values of the same property are ignored". This isn't what we want so switch to `Object.assign`

controllers/apply/awards-for-all/fields/project-budget.js
```diff
@@ -0,0 +1,127 @@
...
+            cy: 'Rhestrwch y costau hoffech i ni eu hariannu'
+        }),
+        explanation: localise({
+            en: `<p>
+                You should use budget headings, rather than a detailed list
+                of items. For example, if you're applying for pens, pencils,
+                paper and envelopes, using 'office supplies' is fine.
+                Please enter whole numbers only.
+            </p>
+            <p>Please note you can only have a maximum of 10 rows.</p>`,
```
mattandrews: This isn't part of this PR really but should the "10" here use the `ROW_LIMIT` above?

controllers/apply/awards-for-all/form.test.js
```diff
@@ -909,7 +908,7 @@ describe('Contacts', () => {
         expect(validationResult.isValid).toBeFalsy();
     });
 
-    test.each(CONTACT_EXCLUDED_TYPES)(
+    test.each(['school', 'college-or-university', 'statutory-body'])(
```
mattandrews: 💯

PR #2761, opened 2019-12-09
New date range clean

davidrapson:
Clean version of https://github.com/biglotteryfund/blf-alpha/pull/2713

Enabled behind a flag in development only.

image

Splits the project start and end date into two fields. Results in a nicer user experience and can be modelled using the regular date field and will allow us to remove the custom extension and macros for the combined date range field. However, it needs a step where data in the old schema format is translated into the new one.

### DIFF
--- a/controllers/apply/awards-for-all/form.js
+++ b/controllers/apply/awards-for-all/form.js
@@ -1361,6 +1361,8 @@ module.exports = function({
         allFields: fields,
         featuredErrorsAllowList: [
             { fieldName: 'projectDateRange', includeBase: false },
+            { fieldName: 'projectStartDate', includeBase: false },
+            { fieldName: 'projectEndDate', includeBase: false },
             { fieldName: 'seniorContactRole', includeBase: false },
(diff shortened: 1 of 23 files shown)


davidrapson:
Copy sent for translation before I merge this

davidrapson:
@mattandrews Mind giving this one final review. Added missing translations and rejigged flags to a) enable on the test site and b) allow unit tests to cover both flag states.

controllers/apply/awards-for-all/form.js
```diff
@@ -31,7 +32,11 @@ const { checkBankAccountDetails } = require('./lib/bank-api');
 module.exports = function({
     locale = 'en',
     data = {},
-    showAllFields = false
+    showAllFields = false,
+    flags = {
+        // Set default flags based on config, but allow overriding for tests
+        enableNewDateRange: config.get('awardsForAll.enableNewDateRange')
```
mattandrews: Really nice 👍

PR #2762, opened 2019-12-09
Only show logout/applications links for non-staff users

mattandrews:
Disables the universal user nav for staff to avoid confusion (eg. redundant "my applications" link).


mattandrews:
[Disable sitewide login nav for Staff users](https://trello.com/c/fvs8QlyE/1009-disable-sitewide-login-nav-for-staff-users)

PR #2765, opened 2019-12-10
Test cleanup

davidrapson:
Updates the awards for all schema tests to try and follow a similar pattern to the standard proposal ones with the goals of:

a) expressing similar tests more concisely and with fewer abstractions
b) increase the speed of running these test as they are by far the longest unit tests

done through a few key changes:

- Introduced a general "invalid form" test which allows us to test common invalid error messages in one go. Relative complexity of the two forms aside this is the main reason the standard form tests are much leaner
- Re-grouped related tests based on either country or organisation type which are the two conditions most other fields change based on
- Reducing the use of abstracted methods so it's clearer to see what the expectations are in a test

results are that:

- are ~300 hundred lines  shorter for the same kind of coverage
- typically run twice as fast. Although these are unit tests so we're only talking 2-3 seconds. Not much difference in CI but is noticeable using a test watcher.

### DIFF
--- a/controllers/apply/awards-for-all/fields.js
+++ b/controllers/apply/awards-for-all/fields.js
@@ -305,13 +305,9 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
             }),
             type: 'text',
             isRequired: true,
-            schema: Joi.when('projectCountry', {
-                is: Joi.exist(),
-                then: Joi.string()
-                    .max(FREE_TEXT_MAXLENGTH.large)
-                    .required(),
-                otherwise: Joi.any().strip()
-            }),
+            schema: Joi.string()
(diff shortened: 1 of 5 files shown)


controllers/apply/awards-for-all/fields.js
```diff
@@ -347,13 +343,9 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
                 autocomplete: 'postal-code'
             },
             isRequired: true,
-            schema: Joi.when('projectCountry', {
```
davidrapson: As discussed removed these on the basis that `projectCountry` is always required so I don't think there's logically a case where we'd need to strip anything here, or that joi ever even runs it, so should be safe to go. Happy to remove this change if we'd rather just focus this change on test updates.

PR #2767, opened 2019-12-11
Contacts: next

mattandrews:
This looks like tons of new code but most of it is a copy/paste of the `awards-for-all` directory. Will call out the main changes in comments.

This isn't completely finished but would let us merge it so everyone can review it without affecting the existing fields/forms, then we can migrate them over when we're happy.

PR #2768, opened 2019-12-11
Contact check

davidrapson:
Very experimental and needs to be combined with https://github.com/biglotteryfund/blf-alpha/pull/2767 but explores adding a custom render function for steps. Involves some solid yak shaving along the way to support this.

image


controllers/apply/contacts-next/contact-check.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+const path = require('path');
+
+module.exports = function(req, res, props) {
+    const { form } = props;
+
+    const seniorContactFieldMappings = {
+        seniorContactName: 'Name',
```
mattandrews: Worth using `localise()` here just so it's baked in from the start?

controllers/apply/contacts-next/form.js
```diff
@@ -61,6 +62,13 @@ module.exports = function({
         );
     }
 
+    function stepCheckContacts() {
+        return new CustomStep({
+            title: 'Check your contacts',
```
mattandrews: Same here on `localise()`, worth just having the structure in place I reckon.

PR #2779, opened 2019-12-16
File field type

davidrapson:
One of the things preventing moving over to the new field types completely for awards for all is missing a file field type. This adds one.

### DIFF
--- a/controllers/apply/lib/field-types/file.js
+++ b/controllers/apply/lib/field-types/file.js
@@ -1,0 +1,88 @@
+'use strict';
+const mime = require('mime-types');
+const filesize = require('filesize');
+
+const Field = require('./field');
+const Joi = require('../joi-extensions');
+
+class FileField extends Field {
+    constructor(props) {
+        super(props);
+
+        this.labelExisting = props.labelExisting || null;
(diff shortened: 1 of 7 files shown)


controllers/apply/lib/field-types/file.js
```diff
@@ -0,0 +1,88 @@
...
+const Field = require('./field');
+const Joi = require('../joi-extensions');
+
+class FileField extends Field {
+    constructor(props) {
+        super(props);
+
+        this.labelExisting = props.labelExisting || null;
+
+        const maxFileSize = {
```
davidrapson: Ended up making the max size and static as file this is related to the server limits so isn't particularly variable.

controllers/apply/lib/field-types/file.js
```diff
@@ -0,0 +1,88 @@
...
+            value: 12 * 1048576,
+            label: '12MB'
+        };
+
+        const supportedFileTypes = [
```
davidrapson: Likewise, made supported file types static as we only have one type of file upload so not too worried about this being configurable at the moment.

PR #2786, opened 2019-12-18
Use a customised name for <details> clicks in HotJar tags

mattandrews:
We were using the same HotJar tag for all clicks on `<details>` elements. This change now uses the field name (if the details element is part of a field) to distinguish between these clicks and ones on the summary screen.


mattandrews:
[Debug Hotjar not properly tracking clicks on <details> elements (I mean, er, "clicky things")](https://trello.com/c/0fq7piQl/1015-debug-hotjar-not-properly-tracking-clicks-on-details-elements-i-mean-er-clicky-things)

PR #2789, opened 2019-12-18
Block deleted apps

mattandrews:
Dashboards will now show this message for any application whose `expiryDate` is in the past:

image

If they somehow managed to still get into the application itself, they'll always see this:

image

This change also includes the expiry **time** as well as date, eg. the actual date stored in the database. This removes the confusion when someone could still edit their application when it was scheduled to be deleted and might go at any minute.

Now, if we tell someone their application expires at 11am, they can edit it until then, but it will be deleted at the next Lambda run (eg. 4pm currently) and they won't be able to access it from 11am onwards.

### DIFF
--- a/controllers/apply/form-router/index.js
+++ b/controllers/apply/form-router/index.js
@@ -315,6 +315,16 @@ function initFormRouter({
                         'status'
                     );
 
+                    if (currentApplication.isExpired) {
+                        return res.render(
+                            path.resolve(__dirname, './views/expired'),
+                            {
+                                title: res.locals.copy.expired.title,
+                                csrfToken: req.csrfToken()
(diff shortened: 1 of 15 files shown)


controllers/apply/form-router/index.js
```diff
@@ -315,6 +315,16 @@ function initFormRouter({
                         'status'
                     );
 
+                    if (currentApplication.isExpired) {
```
davidrapson: Mind adding this a subsequent middleware function?  Wary of this one getting a bit overloaded. Makes sense to do this early on though.

mattandrews: Good call 👍 done

PR #2790, opened 2019-12-19
Tweak expiry metadata

mattandrews:
Two things here:

1. Log in Datadog the application status (eg. `PENDING`, `COMPLETE` etc) when we delete expired apps. This will give us some (potentially painful) data on what state these apps are in at expiry time.

2. Log the `userId` in the email queue table. This is to aid debugging in the case where a customer gets in touch after their application has expired but we can't prove it was deleted as we don't know its application ID.

### DIFF
--- a/db/models/applications.js
+++ b/db/models/applications.js
@@ -373,6 +373,17 @@ class ApplicationEmailQueue extends Model {
             },
 
             /**
+             * User model reference
+             * (optional as it was added after the model was created)
+             * Intended to allow looking up a user's deleted applications
+             * after the application itself was deleted
+             */
+            userId: {
+                type: DataTypes.INTEGER,
+                allowNull: true
+            },
+
+            /**
              * Email type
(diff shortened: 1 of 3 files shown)


db/models/applications.js
```diff
@@ -372,6 +372,17 @@ class ApplicationEmailQueue extends Model {
...
+            /**
+             * User model reference
+             * (optional as it was added after the model was created)
+             * Intended to allow looking up a user's deleted applications
+             * after the application itself was deleted
+             */
+            userId: {
+                type: DataTypes.INTEGER,
+                allowNull: true
+            },
```
mattandrews: My plan is to just manually add this column to the test/live tables – given it allows NULL it's a safe change (just done it locally) even before we merge this.

mattandrews:
[Store the userID alongside the applicationID in the application expiry emails table (to aid debugging)](https://trello.com/c/GjAHgX3Z/1016-store-the-userid-alongside-the-applicationid-in-the-application-expiry-emails-table-to-aid-debugging)

PR #2795, opened 2019-12-20
Field type test coverage

davidrapson:
A little more field type work to increase the overall test coverage for field types. Results in a few tweaks to the field types but will call those out in comments

### DIFF
--- a/controllers/apply/lib/field-types/address.js
+++ b/controllers/apply/lib/field-types/address.js
@@ -118,13 +118,7 @@ class AddressField extends Field {
 
     get displayValue() {
         if (this.value) {
-            return compact([
-                this.value.line1,
-                this.value.line2,
-                this.value.townCity,
-                this.value.county,
-                this.value.postcode
-            ]).join(',\n');
+            return compact(Object.values(this.value)).join(',\n');
         } else {
             return '';
         }
--- a/controllers/apply/lib/field-types/address.js
+++ b/controllers/apply/lib/field-types/address.js
@@ -10,16 +10,16 @@ class AddressField extends Field {
 
         this.type = 'address';
 
-        if (!props.explanation) {
-            this.explanation = this.localise({
-                en: `<p>Enter the postcode and search for the address, or enter it manually below.`,
-                cy: `Rhowch y cod post a chwiliwch am y cyfeiriad, neu ei deipio isod.`
-            });
-        }
+        this.isRequired = true;
+
+        this.explanation = this.localise({
+            en: `Enter the postcode and search for the address, or enter it manually below.`,
+            cy: `Rhowch y cod post a chwiliwch am y cyfeiriad, neu ei deipio isod.`
+        });
     }
 
     defaultSchema() {
-        const schema = Joi.object({
+        return Joi.object({
             line1: Joi.string()
                 .max(255)
                 .required(),
@@ -37,13 +37,7 @@ class AddressField extends Field {
             postcode: Joi.string()
                 .postcode()
                 .required()
-        });
-
-        if (this.isRequired) {
-            return schema.required();
-        } else {
-            return schema.optional();
-        }
+        }).required();
     }
 
     defaultMessages() {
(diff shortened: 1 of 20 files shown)


controllers/apply/lib/field-types/address.js
```diff
@@ -10,16 +10,16 @@ class AddressField extends Field {
...
 
-        if (!props.explanation) {
-            this.explanation = this.localise({
-                en: `<p>Enter the postcode and search for the address, or enter it manually below.`,
-                cy: `Rhowch y cod post a chwiliwch am y cyfeiriad, neu ei deipio isod.`
-            });
-        }
+        this.isRequired = true;
+
+        this.explanation = this.localise({
```
davidrapson: This explanation is a core part of the address field, so have made it static. We don't override it anywhere.

controllers/apply/lib/field-types/address.js
```diff
@@ -10,16 +10,16 @@ class AddressField extends Field {
...
+            en: `Enter the postcode and search for the address, or enter it manually below.`,
+            cy: `Rhowch y cod post a chwiliwch am y cyfeiriad, neu ei deipio isod.`
+        });
     }
 
     defaultSchema() {
-        const schema = Joi.object({
+        return Joi.object({
```
davidrapson: We never have an optional address lookup (not even sure if that'd work) so have made it always required.

controllers/apply/lib/field-types/address.js
```diff
@@ -124,12 +118,7 @@ class AddressField extends Field {
 
     get displayValue() {
         if (this.value) {
-            return compact([
-                this.value.line1,
-                this.value.townCity,
-                this.value.county,
-                this.value.postcode
-            ]).join(',\n');
+            return compact(Object.values(this.value)).join(',\n');
```
davidrapson: Realised that the fix in https://github.com/biglotteryfund/blf-alpha/pull/2780 needed to apply here too. No need to pick out all the values manually when we can use Object.values.

controllers/apply/lib/field-types/field.js
```diff
@@ -13,19 +17,14 @@ class Field {
...
-        // @TODO: Remove this after fully migrating to field types
-        this._isClass = true;
-
         const label = props.label ? props.label : this.defaultLabel();
         if (label) {
             this.label = label;
         } else {
-            // @TODO: Re-enable once welsh translations have been added
-            // throw new Error('Must provide label');
+            throw new Error('Must provide label');
```
davidrapson: This will throw an error when we enable Welsh for the standard form, but that's a good thing as it will show where we're missing translations for labels.

controllers/apply/lib/field-types/file.js
```diff
@@ -30,23 +31,17 @@ class FileField extends Field {
...
-                    .required()
-            });
+        this.isRequired = true;
 
-        if (this.isRequired) {
-            this.schema = schema.required();
-        } else {
-            this.schema = schema.optional();
-        }
+        this.schema = Joi.object({
```
davidrapson: The file schema shouldn't be overridable as it wouldn't make sense with an alternative schema.

controllers/apply/lib/field-types/file.js
```diff
@@ -74,11 +69,12 @@ class FileField extends Field {
 
     get displayValue() {
         if (this.value) {
-            const mimeType = mime.extension(this.value.type) || 'File';
-            const fileSize = filesize(this.value.size, { round: 0 });
-            return `${
-                this.value.filename
-            } (${mimeType.toUpperCase()}, ${fileSize})`;
+            const mimeType = mime.extension(this.value.type);
```
davidrapson: Reworked this display logic so we don't show "FILE" in english only if we try to format an unknown file type.

controllers/apply/lib/field-types/file.test.js
```diff
@@ -27,4 +28,12 @@ test('FileField', function() {
...
         expect.stringContaining('"type" must be one of')
     );
+
+    // Unknown file type
+    field.withValue({
+        filename: 'example.madeup',
+        size: 13264,
+        type: 'application/not-a-thing'
+    });
+    expect(field.displayValue).toBe('example.madeup (13 KB)');
```
mattandrews: 💯

PR #2808, opened 2020-01-07
Pass around an auth cookie to determine login state

mattandrews:
This aims to cut down on the number of calls to `/user/session` which may be contributing to stability issues (eg. every pageview hits it).

When a regular user logs in, a cookie will be set (called `blf-auth-state` which matches the existing naming convention, natch). This cookie is whitelisted for all Cloudfront routes.

If this cookie is set and matches `logged-in` (it's unencrypted as this is solely for UI enhancements rather than as a sole check for authentication) then we toggle the `/user/logout` links. 

This has no effect for Staff users (eg. they'll still be prompted to log in if they're signed in as Staff), which improves the current behavious where logged-in Staff see no UI links at all (which has confused at least one staffer according to my inbox!).

### DIFF
--- a/assets/js/common/forms.js
+++ b/assets/js/common/forms.js
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import forEach from 'lodash/forEach';
 import { trackEvent, tagHotjarRecording } from '../helpers/metrics';
+import { getCookieValue } from '../helpers/cookies';
 import modal from './modal';
 import debounce from 'lodash/debounce';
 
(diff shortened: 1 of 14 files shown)


mattandrews:
[Check for logged-in cookie instead of AJAX call to /user/session](https://trello.com/c/U8aM399d/1021-check-for-logged-in-cookie-instead-of-ajax-call-to-user-session)

mattandrews:
@davidrapson give this another look now – removed the Cloudfront changes and added expiry time.

davidrapson:
@mattandrews Latest changes make sense. Let's see if the hunch about cloudfront rules is correct, but based on our use client-side specific case we don't need to worry if the cookie is passed through to the backend on cached routes.

PR #2818, opened 2020-01-09
Correct logging for application expiry Lambda function

mattandrews:
Noticed there was an issue with how we log expiries compared to expiry emails. Lambda function currently gets this output:

```
{
    "emailQueue": [],
    "expired": true
}

```
`expired` there should be an array for each expiry processed (like the `emailQueue` is) rather than a success/fail flag.

This change makes it match `emailQueue`, so the output will be:
```

{
  "emailQueue": [],
  "expired": [
    {
      "applicationDeleted": true
    },
    {
      "applicationDeleted": true
    }
  ]
}
```

### DIFF
--- a/controllers/apply/expiry.js
+++ b/controllers/apply/expiry.js
@@ -201,14 +201,17 @@ async function deleteExpiredApplications(expiredApplications) {
 
             const dbStatus = await PendingApplication.deleteBatch(ids);
 
-            expiredApplications.forEach(application => {
+            const status = expiredApplications.map(application => {
                 logger.info(`Deleting expired application`, {
                     formId: application.formId,
                     applicationStatus: application.currentProgressState
                 });
+                return {
+                    applicationDeleted: true
+                };
             });
 
-            return dbStatus === expiredApplications.length;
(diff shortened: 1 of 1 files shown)


controllers/apply/expiry.js
```diff
@@ -201,14 +201,17 @@ async function deleteExpiredApplications(expiredApplications) {
...
                     formId: application.formId,
                     applicationStatus: application.currentProgressState
                 });
+                return {
+                    applicationDeleted: true
+                };
             });
 
-            return dbStatus === expiredApplications.length;
+            return dbStatus === expiredApplications.length ? status : false;
```
davidrapson: Not sure I follow this line. If we have expired applications then when would `dbStatus === status` be truthy? Is this not comparing two different types of value?

mattandrews: It's a bit confusing but I believe `dbStatus` returns the number of rows affected by the operation (in this case the bulk deletion above), so it always returns `true` unless something weird happened (eg. it deleted some but not all of the applications). We do something similar in the `sendExpiryEmails` function below this to confirm it did as expected.

PR #2821, opened 2020-01-09
Include client IP in salesforce submission

davidrapson:
Slightly speculative PR here. In theory this is all that would be needed to pass the client IP through to Salesforce. If we wanted to do this we'd need to confirm the process for mapping this somewhere in salesforce where it could be useful in reports.

### DIFF
--- a/controllers/apply/awards-for-all/docs/schema.md
+++ b/controllers/apply/awards-for-all/docs/schema.md
@@ -4,6 +4,9 @@ The following documents the data schema for Awards for All applications when sub
 
 ## Changelog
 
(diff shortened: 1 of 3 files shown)


davidrapson:
[Review storing IP address on submission to help identify multiple applications from the same device.](https://trello.com/c/ZDR8VAEJ/1032-review-storing-ip-address-on-submission-to-help-identify-multiple-applications-from-the-same-device)

davidrapson:
Confirmed there's no mapping issues with us starting to send this so we can get this out tomorrow.

PR #2823, opened 2020-01-10
Manually download initial clamav database

davidrapson:
After reviewing a few more clamav install scripts, notably Dockerfiles, noticed that a common approach was to manually download the three database files upfront before starting freshclam in the background. Tested these steps on a plain Ubuntu VM successfully so worth giving this a go.


davidrapson:
[Avoid need to wait for an arbitrary period for clamav to be running](https://trello.com/c/C783oJ58/1035-avoid-need-to-wait-for-an-arbitrary-period-for-clamav-to-be-running)

PR #2826, opened 2020-01-10
Skip straight to success screen if a duplicate application submission occurs

mattandrews:
Found this tricky to reproduce as reloading the success screen kicks you into the dashboard (as your currently-editing ID session key is deleted after submission), but it's possible that double clicks on submit might cause this, or weird network dropouts etc. Either way, this detects an existing submission and just renders the success screen instead of trying to re-insert it.

### DIFF
--- a/controllers/apply/form-router/submission.js
+++ b/controllers/apply/form-router/submission.js
@@ -50,7 +50,9 @@ module.exports = function(
             currentApplication.id
         );
         if (submittedApplicationExists) {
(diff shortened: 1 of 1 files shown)


mattandrews:
[Fix double submit error](https://trello.com/c/4wIU4r43/1034-fix-double-submit-error)

mattandrews:
Good call, just added.

davidrapson:
:shipit:

PR #2825, opened 2020-01-10
Cleanup deploy scripts

davidrapson:
Does some work to better document our deploy scripts. Main changes are to:

- Rename scripts to match their associated CodeDeploy phases
- Merge bootstrap.sh and provision.sh as they are part of the same CodeDeploy phase
- Add more detailed comments to provision script to describe _why_ we are doing each step
- Added https://github.com/koalaman/shellcheck  to Travis

PR #2827, opened 2020-01-10
Cleanup nginx config

davidrapson:
Builds on from https://github.com/biglotteryfund/blf-alpha/pull/2825 and cleans up and documents our nginx configs.


deploy/nginx.conf
```diff
@@ -1,10 +1,9 @@
+# Directives in this file are loaded in the global `http` block
+
 # Passenger config
 client_max_body_size 50M;
 passenger_core_file_descriptor_ulimit 50000;
 
-# Switch on gzip for static assets
-gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
```
davidrapson: This is set in `server.conf` so we don't need this twice.

deploy/server.conf
```diff
@@ -1,46 +1,48 @@
...
+    # user_routes is defined in nginx.conf
     location ~ ^/user {
         limit_req zone=user_routes burst=20;
     }
 
     location ~ ^/welsh/user {
         limit_req zone=user_routes burst=20;
     }
-
-    # Media: images, icons, video, audio, HTC
```
davidrapson: Given that everything under `/assets` is sent directly to S3 this block should never get executed.

mattandrews: True, I guess this only gets called when accessing the site directly, but we can live with that not being cached etc (everything else isn't anyway) 👍

davidrapson: Good point, happy to keep this in to account for that use-case?

PR #2842, opened 2020-01-13
Streamline client-side js

davidrapson:
Streamlines our client-side JS by inlining the  analytics initialisation code instead of code-splitting, replacing a babel plugin with a simpler lint rule, and inlining a couple single-use helpers.


assets/js/main.js
```diff
@@ -49,23 +50,26 @@ import(/* webpackChunkName: "vue-apps" */ './vue-apps/index').then(vueApps => {
...
 }
 
-if (!shouldBlockAnalytics() && !isDoNotTrack) {
-    import(/* webpackChunkName: "analytics" */ './analytics').then(
-        analytics => {
-            analytics.init();
-        }
-    );
+if (shouldInitAnalytics() === true) {
+    analytics.init();
```
davidrapson: We were code-splitting this but it was a ~2kb module so really isn't needed. Inlining this to save the extra network request.

assets/js/main.js
```diff
@@ -49,23 +50,26 @@ import(/* webpackChunkName: "vue-apps" */ './vue-apps/index').then(vueApps => {
...
+        window.navigator.doNotTrack === '1' ||
+        window.navigator.msDoNotTrack === '1';
 
-/* Disable analytics outside of the real domain to avoid polluting data. */
-function shouldBlockAnalytics() {
-    const isProduction = window.AppConfig.environment === 'production';
-    const isProdDomain =
-        window.location.hostname === 'www.tnlcommunityfund.org.uk';
-    return !isProduction && !isProdDomain;
+    if (window.AppConfig.environment === 'production') {
```
davidrapson: The logic here was getting a little bit "will you heed my rhyming verse, and take my answer in reverse: when is the hostname not-not the one for which I search" so I've tried to make it more direct.

mattandrews: Great improvement 👍

PR #2851, opened 2020-01-14
Add Standard expiry emails

mattandrews:
This change means that newly-created Standard apps will also create three expiry emails in the queue, with the same config as AFA.

I've updated the email templates to reflect this, which wasn't as tricky as I expected – there weren't too many AFA-specific references.

I also reinstated a temporary seeder endpoint which will backdate the email queue for all pending Standard applications. Testing this locally works as expected – one tradeoff is that it's a one-shot method: running it twice will add duplicate queue items. When we first launched the expiry queue we could just truncate the table and seed it again if something went wrong, but we can't do that now. That said, a simple SQL query could remove all queued Standard emails if we got something wrong here. It looks fine to me though as the logic is pretty standard. Thoughts welcome!


controllers/apply/expiry.js
```diff
@@ -81,33 +91,25 @@ async function sendExpiryEmails(req, emailQueue) {
...
-                    break;
-                case 'AFA_ONE_DAY':
-                    subjectLine = {
-                        en: 'You have one day to finish your application',
-                        cy: 'Mae gennych ddiwrnod i orffen eich cais'
-                    };
-                    break;
+            const formId = emailToSend.PendingApplication.formId;
+
+            let projectCountry, isBilingual, emailNameParam;
```
davidrapson: There's a lot of different assignments going on here. If possible I'd look to break this up into separate local functions that can be called where you need them, even if this involves a little more processing e.g.

```js
function getAppData(field) {
    return get(
        emailToSend.PendingApplication,
        `applicationData.${field}`
    );
}

function getProjectCountry() {
    if (formId === 'awards-for-all') {
        return getAppData('projectCountry');
    } else if (formId === 'standard-enquiry') {
        const countries = getAppData('projectCountries');
        return countries.length === 1 ? countries[0] : 'Multiple';
    }
}

function isBilingual() {
    if (formId === 'awards-for-all') {
        return getAppData('projectCountry') === 'wales';
    } else if (formId === 'standard-enquiry') {
        return getAppData('projectCountries').includes('wales');
    }
}

function getEmailName() {
    if (formId === 'awards-for-all') {
        return 'application_expiry_afa';
    } else if (formId === 'standard-enquiry') {
        return 'application_expiry_standard';
    }
}
```

controllers/apply/standard-proposal/constants.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const EXPIRY_EMAIL_REMINDERS = [
```
davidrapson: Just wondering, given these are the same for awards for all do we need to define them twice? Could we pass an "enableExpiryReminders" flag to a form model which is what determines if we should set these? Definitely don't need to decide in this PR.

davidrapson:
[Configure expiry email sending for Standard applications](https://trello.com/c/N4S89lNZ/1036-configure-expiry-email-sending-for-standard-applications)

controllers/apply/standard-proposal/constants.js
mattandrews: The `emailType` keys are different (`STANDARD_ONE_WEEK` vs `AFA_ONE_WEEK`), although this was because I envisioned there being different email templates for each one of these, whereas we're actually sending the same email for all six types (just with some different variables). I'd like to keep differentiating them (if only for the database/logs so we know what types of email we've sent) but maybe there's a smart way to avoid duplicating the actual intervals/dates etc.

controllers/apply/email.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
```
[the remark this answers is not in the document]
mattandrews: Yeah makes sense... what do you think, a new object on the form model called something like `expiryEmailParameters` with properties like `sender`, `editLink` and `intro`? (the subject line should probably be defined alongside the email config (eg. 1 month, 1 day, 1 week) as they're so linked to the definition of those intervals.

controllers/apply/standard-proposal/constants.js
davidrapson: Makes sense. This is probably one of those cases where duplicating until we have a clearer model for the data makes sense. Go with it as is 👍

controllers/apply/email.js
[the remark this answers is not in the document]
davidrapson: Yeah something like that would make sense, don't think we need to do that in this PR though

PR #2852, opened 2020-01-14
Upgrade globby and clean up patterns router

davidrapson:
Upgrade `globby` and clean up patterns router. Mostly so that we set breadcrumbs in the same way as the rest of the app.

### DIFF
--- a/controllers/pattern-library/index.js
+++ b/controllers/pattern-library/index.js
@@ -1,103 +1,90 @@
 'use strict';
 const fs = require('fs');
 const path = require('path');
-const globby = require('globby');
 const express = require('express');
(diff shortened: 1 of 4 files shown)


controllers/pattern-library/index.js
```diff
@@ -1,103 +1,90 @@
...
                     .sort();
+
                 res.render(path.resolve(__dirname, './views/components'), {
                     title,
                     breadcrumbs,
                     componentSlugs: componentSlugs
                 });
             } else {
-                next(new Error('No components fount'));
+                next(new Error('No components found'));
```
mattandrews: :trollface:

PR #2853, opened 2020-01-14
Apply lodash/import-scope project-wide

davidrapson:
Added in https://github.com/biglotteryfund/blf-alpha/pull/2842 but realised it would be useful app wide to help us import lodash methods in a consistent way.


davidrapson:
This will introduce a few lint warnings for now. Included a couple of sample changes. Not a huge number that'd need updating. Could be done in a couple of follow on PRs if this change looks OK.

PR #2854, opened 2020-01-15
Add trading name step

mattandrews:
Needs translation, but worth a review in its current form:

Adds this new question:

image

... which leads to:

image

### DIFF
--- a/controllers/apply/awards-for-all/fields.js
+++ b/controllers/apply/awards-for-all/fields.js
@@ -6,10 +6,12 @@ const { oneLine } = require('common-tags');
 
 const Joi = require('../lib/joi-extensions');
 
+const Field = require('../lib/field-types/field');
 const EmailField = require('../lib/field-types/email');
 const DateField = require('../lib/field-types/date');
 const PhoneField = require('../lib/field-types/phone');
 const NameField = require('../lib/field-types/name');
+const RadioField = require('../lib/field-types/radio');
 
 const fieldContactLanguagePreference = require('./fields/contact-language-preference');
 const fieldOrganisationStartDate = require('./fields/organisation-start-date');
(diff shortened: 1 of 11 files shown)


controllers/apply/awards-for-all/fields.js
```diff
@@ -1173,6 +1173,43 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
                 }
             ]
         },
+        organisationHasDifferentTradingName: {
```
davidrapson: If added a new field I'd suggest using `RadioField` from the new field types. Something like:

```js
new RadioField({
    locale: locale,
    name: 'organisationHasDifferentTradingName',
    label: localise({
        en: `Does your organisation use a different name in your day-to-day work?`,
        cy: `@TODO i18n`
    }),
    options: [
        {
            value: 'yes',
            label: localise({ en: `Yes`, cy: `@TODO i18n` })
        },
        {
            value: 'no',
            label: localise({ en: `No`, cy: `@TODO i18n` })
        }
    ],
    messages: [
        {
            type: 'base',
            message: localise({
                en: 'Select an option',
                cy: 'Dewis opsiwn'
            })
        }
    ]
})
```

controllers/apply/awards-for-all/fields.js
```diff
@@ -1221,15 +1258,33 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
...
-            isRequired: false,
-            schema: Joi.string()
-                .allow('')
-                .max(FREE_TEXT_MAXLENGTH.large)
-                .optional(),
+            isRequired: true,
+            get schema() {
+                return Joi.when('organisationHasDifferentTradingName', {
+                    is: 'yes',
+                    then: Joi.string()
```
davidrapson: Should this be required at this point? Rather than optional?

controllers/apply/awards-for-all/fields.js
```diff
@@ -1221,15 +1258,33 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
         organisationTradingName: {
```
davidrapson: Like above worth converting this to use a field type if we're updating this field.

mattandrews:
@davidrapson all the `Field` (and subtypes) tests are [failing](https://travis-ci.org/biglotteryfund/blf-alpha/builds/638423352#L658) after that error I added. Do you think we should remove it for now? Making each test work again will be quite painful (eg. adding `base` messages to every mock, or modifying the error check logic to look only for `Field` instances, as opposed to `Radio` etc).

davidrapson:
@mattandrews Yeah, remove it for now. Happy to pick up a follow on PR to re-add it, or maybe pair on it.

controllers/apply/awards-for-all/transforms.js
```diff
@@ -27,7 +27,7 @@ function transformProjectDateRange(applicationData) {
 }
 
 function transformOrgHasDifferentTradingName(applicationData) {
-    if (!!get('organisationTradingName')(applicationData)) {
+    if (get('organisationTradingName')(applicationData)) {
```
davidrapson: As this is a boolean check could use `has` rather than `get`.

mattandrews: Will have a look at this, empty strings might break...

mattandrews: ... just checked and I think this will be fine. If the field's value is `''` then it's falsey, which is what we want here.

PR #2864, opened 2020-01-16
Add guidance to main and senior contact to stop them from entering shortened names or nicknames

davidrapson:
Add guidance to main and senior contact to stop them from entering shortened names or nicknames.

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1587,6 +1587,10 @@ applyNext:
   fields:
     fullName:
       firstName: First name
+      firstNameExplanation: >
+        This can't be a nickname—or a short version of their name.
+        We need their full first name to check their identity.
+        If you don't give us this, it could delay your application.
       lastName: Last name
     dates:
       startDate: Start date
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -1587,6 +1587,10 @@ applyNext:
   fields:
     fullName:
       firstName: First name
+      firstNameExplanation: >
```
davidrapson: Have sent this for translation

davidrapson: Have sent this for translation

controllers/apply/awards-for-all/fields.js
```diff
@@ -1372,10 +1372,6 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
                 en: 'Full name of main contact',
                 cy: 'Enw llawn y prif gyswllt'
             }),
-            explanation: localise({
```
davidrapson: Removing this on the basis that:

- We already validate the address against a UK-only postcode, so the help text is redundant
- It doesn't always apply (i.e. if you are a statutory body where we don't ask for address)

Have queried this on the related card. If we _do_ need to keep it this help text it should go alongside the address field instead.

davidrapson: Moved this out to a task in the backlog

PR #2865, opened 2020-01-17
Replace redirectForLocale with consistent localify use

davidrapson:
Every time I came across the `redirectForLocale` method I had to double check the behaviour. Realised it's an alias for `localify`. Given we use `localify` in 99% of cases directly I've opted to remove `redirectForLocale` in favour of using `localify` directly.

### DIFF
--- a/common/authed.js
+++ b/common/authed.js
@@ -1,5 +1,5 @@
 'use strict';
-const { localify, redirectForLocale } = require('./urls');
+const { localify } = require('./urls');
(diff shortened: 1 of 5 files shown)


common/authed.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-const { localify, redirectForLocale } = require('./urls');
```
davidrapson: Good example here where we use both in the same file.

PR #2874, opened 2020-01-20
Split up common and user specific locals for apply section

davidrapson:
Splits up the common middleware in the apply section so that common locals are set early enough to be available for the start page when it comes before log in. Specifically form model related locals like `formTitle` and `formBaseUrl`.


davidrapson:
[Add 'Before you start' page to NLA4A form](https://trello.com/c/7WRA0wkx/1018-add-before-you-start-page-to-nla4a-form)

PR #2889, opened 2020-01-27
Add edit mode links

mattandrews:
This adds support for `?edit` in step URLs, which changes the button wording to "Back to Summary", and means that after submission (validation errors or not), you'll end up back on the Summary screen.

### DIFF
--- a/controllers/apply/form-router/steps.js
+++ b/controllers/apply/form-router/steps.js
@@ -70,11 +70,7 @@ module.exports = function(formId, formBuilder) {
          * redirect to the first step in the section.
          */
         if (!stepNumber) {
-            return res.redirect(
-                `${res.locals.formBaseUrl}/${section.slug}/1${
-                    isEditing ? '?edit' : ''
-                }`
-            );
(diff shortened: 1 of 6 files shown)


davidrapson:
[Review the experience of answering or amending a question from the summary](https://trello.com/c/hlxsn300/1052-review-the-experience-of-answering-or-amending-a-question-from-the-summary)

controllers/apply/form-router/steps.js
```diff
@@ -69,7 +70,11 @@ module.exports = function(formId, formBuilder) {
          * redirect to the first step in the section.
          */
         if (!stepNumber) {
-            return res.redirect(`${res.locals.formBaseUrl}/${section.slug}/1`);
+            return res.redirect(
+                `${res.locals.formBaseUrl}/${section.slug}/1${
```
davidrapson: Not sure about this line, do we need this logic?

mattandrews: Perhaps not anymore, it was only relevant when you could click the section titles from the summary screen (whereas now it's just the "Change" links). Will remove.

PR #2890, opened 2020-01-27
Correct fallback date format for .calendar

davidrapson:
.calendar() uses 'L' as the date format once the date is more than two weeks into the past/future. We were passing 'en' as the locale when we actually wanted `en-GB` as the former equals an American locale. Decided to also pass a custom date format for `sameElse` which determines the fallback format as it would be better if this was our common longer date format.

**Before**

image

**After**

image

### DIFF
--- a/common/locals.js
+++ b/common/locals.js
@@ -187,10 +187,15 @@ module.exports = function(req, res, next) {
      * @param dateString
      */
     res.locals.formatCalendarTime = function(dateString) {
-        return moment(dateString)
-            .tz('Europe/London')
-            .locale(locale)
-            .calendar();
+        return (
+            moment(dateString)
+                .tz('Europe/London')
+                // @TODO: Handle this consistently across date formatters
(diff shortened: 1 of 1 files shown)


davidrapson:
[Bugfix: Submitted and last updated dates showing in US format](https://trello.com/c/98Wx0p83/1056-bugfix-submitted-and-last-updated-dates-showing-in-us-format)

common/locals.js
```diff
@@ -187,10 +187,16 @@ module.exports = function(req, res, next) {
...
      */
     res.locals.formatCalendarTime = function(dateString) {
-        return moment(dateString)
-            .tz('Europe/London')
-            .locale(locale)
-            .calendar();
+        return (
+            moment(dateString)
+                .tz('Europe/London')
+                // .locale(locale)
```
mattandrews: Could probably remove this line?

PR #2892, opened 2020-01-28
Override userNavigationLinks on confirmation screen

davidrapson:
We previously had a 'summary' link on the confirmation screen which no longer applied.

**Before**

Screenshot 2020-01-28 at 11 29 01

**After**

Screenshot 2020-01-28 at 11 28 09


davidrapson:
[Remove summary link from confirmation screen](https://trello.com/c/tIrgU8bp/1055-remove-summary-link-from-confirmation-screen)

PR #2894, opened 2020-01-28
Debug cypress

davidrapson:
Opened to debug https://github.com/biglotteryfund/blf-alpha/pull/2892 failing


davidrapson:
In the process of doing this the build passed again, seems like there may have been something flaky on the `common` integration test, but it's working OK now.

PR #2902, opened 2020-01-30
Update turn around times

davidrapson:
Update turn around times to reflect reduced 12 week turnaround time only applying to Scotland and Northern Ireland for now.

~Sent this copy for translation, will update the welsh versions once I get this back.~ Updated this commit with latest welsh copy.


davidrapson:
[Review country specific lead times](https://trello.com/c/EBhxQygj/1013-review-country-specific-lead-times)

PR #2903, opened 2020-01-31
Add custom email templates based on form type

davidrapson:
In order to better contextualise expiry emails depending on the type of form I've opted to use a dedicated email template per form ID rather than using a combined one. Allows them to use specific language without jumping over ourselves by introducing variables for the parts that change.

New copy needs translation so will add that once I've got it.


controllers/apply/emails/standard-expiry.njk
```diff
@@ -0,0 +1,49 @@
...
+            Bilingual message – please scroll down for Welsh.<br />
+            Neges ddwyieithog – y Gymraeg i’w weld isod.
+        </strong></p>
+    {% endif %}
+
+    <p>Hello,</p>
+
+    <p>
+        <strong>Your funding proposal isn’t finished yet</strong><br/>
+        We noticed you started telling us about {{ projectName | safe }}.
```
mattandrews: This variable will need a guard around it as it doesn't always exist.

controllers/apply/emails/standard-expiry.njk
```diff
@@ -0,0 +1,49 @@
...
+        But you haven’t finished yet.
+    </p>
+
+    <p>
+        <strong>
+            You have until {{ expiryDate.en }} to finish your funding proposal.
+        </strong><br/>
+        If you don’t complete it at that time, we’ll delete it -
```
mattandrews: "at that time"... maybe "before that time" is clearer?

davidrapson:
@mattandrews Mind giving this another look. Ended up reworking this to add some mock tests to give me some more confidence in the changes I was making. Involves a fair bit of restructuring to support that so appreciate a good tyre-kicking.

controllers/apply/expiries/send-expiry-email.js
```diff
@@ -0,0 +1,181 @@
...
+
+    const baseLink = `/apply/emails/unsubscribe?token=${unsubscribeToken}`;
+
+    const templateData = {
+        isBilingual: isBilingual,
+        projectName: get(applicationData, 'projectName'),
+        countryPhoneNumber: getPhoneFor(projectCountry),
+        countryEmail: getEmailFor(projectCountry),
+        unsubscribeLink: {
+            en: `https://tnlcommunityfund.org.uk/${baseLink}`,
```
mattandrews: Can we add `www.` in here? You also don't need the trailing slash before `${baseLink}` (it starts with a slash!)

davidrapson:
@mattandrews Updated this with the changes we discussed. Notably added some more unit and integration tests to increase coverage around the expiry logic and made the change to retain the relative time when adding items to the expiry queue. Once this is out we can update the lambda function to run more frequently rather than processing a single batch each day.

common/secrets.js
```diff
@@ -106,7 +106,8 @@ const EMAIL_EXPIRY_SECRET =
     process.env.EMAIL_EXPIRY_SECRET || getParameter('emailExpiry.secret');
 
 const EMAIL_EXPIRY_TEST_ADDRESS =
-    process.env.APPLICATION_EXPIRY_EMAIL || getParameter('emailExpiry.testEmail');
+    process.env.EMAIL_EXPIRY_TEST_ADDRESS ||
```
mattandrews: 💯

controllers/apply/expiries/send-expiry-email.js
```diff
@@ -0,0 +1,177 @@
...
+            cy: expiresOn.locale('cy').format(dateFormat)
+        };
+    }
+
+    const baseUrl = 'https://www.tnlcommunityfund.org.uk';
+
+    const unsubscribeLinkUrlPath = `/apply/emails/unsubscribe?token=${unsubscribeToken}`;
+    const unsubscribeLink = {
+        en: `${baseUrl}${unsubscribeLinkUrlPath}`,
+        cy: `${baseUrl}welsh${unsubscribeLinkUrlPath}`
```
mattandrews: Is this missing a slash still? eg. won't it output `https://www.tnlcommunityfund.org.ukwelsh/apply/emails/unsubscribe?`?

davidrapson:
This is ready to go now, but have forwarded Lily a copy of the new standard emails to double check before we put this out.

davidrapson:
Everyone is happy with the amended emails so this is good to go now.

PR #2905, opened 2020-01-31
Tidy up use of blog tags

davidrapson:
Makes a change to blog posts to remove tags and, as a result, simplify the post metadata section.

Ran this by @LynseyJReynolds and aggree this makes sense as the practical use of this is a) low usage based on analytics and b) noisy. with many tags per post leading to a messy taxonomy.

We already have (more stable) categories for posts so having two taxonomies here doesn't make sense given our current use case.

Will give the appropriate people a heads up before this goes out and we might want to refine the meta section a bit, but hopefully the benefit of the before and after is clear.

**Before**

Screenshot 2020-01-31 at 16 17 23

**After**

Screenshot 2020-01-31 at 16 15 12

### DIFF
--- a/assets/sass/components/_badges.scss
+++ b/assets/sass/components/_badges.scss
@@ -28,6 +28,21 @@
     }
 }
 
+.o-tag {
+    display: inline-block;
+    border: 1px solid get-color('brand', 'primary');
+    color: get-color('brand', 'primary');
+    display: inline-block;
+    padding: 2px 7px;
+    text-decoration: none;
+    font-size: 14px;
+    margin: 3px 3px 3px 0;
+    @include on-interact() {
+        border-color: darken(get-color('brand', 'primary'), 10%);
+        color: darken(get-color('brand', 'primary'), 10%);
+    }
+}
+
 /* =========================================================================
    Pills
    ========================================================================= */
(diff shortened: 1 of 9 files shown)


mattandrews:
Not quite sold on this TBH. We only have three(!) categories for blogs currently which doesn't feel particularly wide-ranging, and I don't think we should drop the concept of tagging just because this particular blogpost has a large amount of (lengthy) tags. The only benefit of the before/after is... that the visual appearance of that side panel is a bit cleaner?! If the issue is lots of one-item tags being created, maybe the fix here is to chat to those teams about better, sitewide taxonomies, rather than just remove them altogether?

davidrapson:
Completely get where you are coming from. Used this post as example, but it applies across the board. My take on this would be that we have 16 posts across a two year period, even if we increase the rate we publish at it'd take a _long_ time to reach the point where you have many posts where free-form tags are useful.

Completely up for having a conversation about better taxonomy, but we could happily use categories for that I think. The friction in creating them becomes a strength as it makes it easier to stick to. If there's a strong feeling we need both then I'd be keen to _limit_ the maximum we show here, just to provide some better guard rails to discourage over-use.

davidrapson:
image

After discussion agreed it was worth keeping blog tags and working to improve how they are used rather than removing them all together. Keeping them just in the footer of the post.

PR #2915, opened 2020-02-04
Fix preview URLs for new Craft version

mattandrews:
Looks like some preview URLs Craft generates in the latest version don't include `?token` URL parameters, which means our `X-Frame-Options` doesn't get unset and the preview iframe fails. This change allows them through again.


mattandrews:
Not sure but this doesn't seem to be happening anymore so may be fixed. See https://github.com/craftcms/cms/issues/5359#issuecomment-574875037 for more

PR #2917, opened 2020-02-05
Cleanup lead time flags

davidrapson:
This took a bit of work which is a reminder that there was too much stuff behind this feature flag. Reminder to future self: don't tie two changes to a single feature flag 🙈 

Includes new project date fields in schema docs and add deprecated note to `projectDateRange`. Whilst this is a backwards compatible change also bumping the version number to reflect the addition of new fields.


controllers/apply/awards-for-all/transforms.js
```diff
@@ -39,11 +38,10 @@ function transformOrgHasDifferentTradingName(applicationData) {
...
 function transform(applicationData) {
     // Add any default transforms here
-    let transformsToRun = [transformOrgHasDifferentTradingName];
-
-    if (config.get('awardsForAll.enableNewDateRange') === true) {
-        transformsToRun.push(transformProjectDateRange);
-    }
+    const transformsToRun = [
+        transformOrgHasDifferentTradingName,
+        transformProjectDateRange
```
mattandrews: I wonder if we should add notes to these (or even calendar reminders) set 3 months after they went live, so we can delete them? There'll come a point when all the pending applications with the old format have been deleted, and any submitted ones will have either used the new format, or were submitted before this change (and therefore shouldn't be affected)?

davidrapson: Sent a calendar invite for three months time. We log usage so we can confirm they are no longer used and then delete.

davidrapson:
[Remove supporting feature flag code from rollout of new date fields and lead times](https://trello.com/c/vDUe8FDA/1067-remove-supporting-feature-flag-code-from-rollout-of-new-date-fields-and-lead-times)

PR #2920, opened 2020-02-05
Remove need for enableHotjar flag

davidrapson:
Remove need for `enableHotjar` flag in favour of only checking existence of `hotjarId` config item.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -23,7 +23,6 @@
     }
   },
   "features": {
-    "enableHotjar": false,
     "enableRelatedGrants": false,
     "enableSalesforceConnector": true,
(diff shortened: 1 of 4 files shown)


config/default.json
```diff
@@ -41,6 +40,6 @@
     "cookieLogin": "blf-auth-state",
     "expiryInSeconds": 7200
   },
-  "hotjarId": 1036292,
+  "hotjarId": null,
```
davidrapson: This is necessary otherwise `config.get` errors on an undefined key. Set to null by default and then defined for test and production environments only. We don't want to run hotjar in development.

PR #2931, opened 2020-02-07
Proof of concept query to get the status of a funding request

davidrapson:
Not proposing this be merged but raising this pull request to show the process of querying salesforce via SOQL to get the current funding request status. The shape of the response is a bit verbose but the query itself is surprisingly lean, you are able to query fields on linked tables without explicit JOIN statements.

### DIFF
--- a/controllers/apply/form-router/lib/salesforce.js
+++ b/controllers/apply/form-router/lib/salesforce.js
@@ -23,6 +23,19 @@ class Salesforce {
             body: body
         });
     }
+    getFundingRequestStatus(formDataSalesforceId) {
+        return this.get(
+            `/services/data/${this.apiVersion}/query?q=${encodeURIComponent(
+                `SELECT Funding_Request__r.Status__c FROM Form_Data__c WHERE ID='${formDataSalesforceId}'`
+            )}`
+        ).then(function(result) {
+            if (result.records.length > 0) {
+                return result.records[0]['Funding_Request__r']['Status__c'];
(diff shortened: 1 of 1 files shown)


davidrapson:
[What status information can we get from Salesforce](https://trello.com/c/UzYGSvzL/1028-what-status-information-can-we-get-from-salesforce)

controllers/apply/form-router/lib/salesforce.js
```diff
@@ -30,7 +35,40 @@ class Salesforce {
             )}`
         ).then(function(result) {
             if (result.records.length > 0) {
-                return result.records[0]['Funding_Request__r']['Status__c'];
+                const rawStatus =
+                    result.records[0]['Funding_Request__r']['Status__c'];
+
+                let friendlyStatus = 'Unknown';
+                switch (rawStatus) {
```
davidrapson: Added this based on the full list of pick-list values. I've been pretty arbitrary here so consider this a rough indication of what we might need to do, not real attempt.

mattandrews:
Oh wow, this is pretty simple, yet powerful. Nice work! 👍

Wonder if we can calculate expected increase in API calls based on adding this to every dashboard pageview for each submitted app..?

davidrapson:
Added some notes around constraints to the trello card. Something to think about when we come to do this for real. Going to close this now as I think it's enough to establish what the query looks like should we need to do this.

PR #2939, opened 2020-02-11
Add form brands [WIP]

mattandrews:
This is a bit of rough work really to see how feasible this kind of thing could be. In short, this is what the changes do:

- Application forms can be called with a new `brand` parameter, eg. `/apply/your-funding-proposal?brand=kingsFund`

- If the form in question allows this branding, then a `formBrand` object is stored in the session (the object is of the form: `{ slug: "kingsFund", title: "The Kings Fund"}`)

- When the form is first saved to the database, we add the `formBrand` object to a new `metadata` column against `PendingApplication`

- When the form is submitted, the `forSalesforce()` prepends `formBrand.title` to the Project Name field (eg. "The Kings Fund: My Project Name")

Depending on GMS options we may be able to avoid modifying the project name and just sending the form brand as metadata and filtering from there, but waiting to hear back on that.

Comments/thoughts on this approach are very welcome!


controllers/apply/awards-for-all/form.js
```diff
@@ -1348,6 +1349,10 @@ module.exports = function({
             );
         }
 
+        if (metadata && metadata.formBrand) {
+            enriched.projectName = `${metadata.formBrand}: ${enriched.projectName}`;
```
davidrapson: Sounds like we might try to handle this purely through a metadata field rather so we might not need this?

controllers/apply/form-router/index.js
```diff
@@ -64,10 +70,22 @@ function initFormRouter({
         next();
     }
 
+    function setFormBrand(req, res, next) {
+        const validBrand = allowedBrands.find(
```
davidrapson: If using an object for these this could become `const validBrand = allowedBrands[req.query.brand]` or `const validBrand = get(allowedBrands, req.query.brand)`

controllers/apply/form-router/index.js
```diff
@@ -64,10 +70,22 @@ function initFormRouter({
...
+            brand => brand.slug === req.query.brand
+        );
+        if (validBrand) {
```
davidrapson: I'd perhaps write this as:

```js

if (validBrand) {
  set(req.session, formBrandSessionKey(), validBrand);
  req.session.save(() => {
    next();
  });
} else {
  next();
}
```

So that we only need to save the session if touching it.

controllers/apply/form-router/index.js
```diff
@@ -220,10 +238,22 @@ function initFormRouter({
...
                 userId: req.user.userData.id
-            });
+            };
+
+            const formBrand = get(req.session, formBrandSessionKey());
+            if (formBrand) {
+                newApplication.metadata = {
+                    formBrand: formBrand
+                };
+                unset(req.session, formBrandSessionKey());
```
davidrapson: Do we need to call `session.save` here as well?

davidrapson:
Also, wondering if we should say programme rather than brand? That's what it ultimately is, right?

controllers/apply/awards-for-all/form.js
mattandrews: Yeah potentially, will leave it for now until it's clearer what the solution will be.

mattandrews:
Updated this after comments, but haven't renamed to `programme` just yet... slightly wary of a slippery slope if we start linking it explicitly to funding programmes rather than it just being a "branded" version of an existing fund. But easy to change if we need to 👍

mattandrews:
Have just renamed `formBrand` to `programme` – can you give this another look @davidrapson?

davidrapson:
👍

PR #2953, opened 2020-02-14
Migrate content API to got

davidrapson:
Migrate content API to got. This is obviously fairly wide reaching so happy if we want to break this down into smaller PRs and do a content-type at a time or something.

### DIFF
--- a/common/content-api.js
+++ b/common/content-api.js
@@ -1,66 +1,45 @@
 'use strict';
-const filter = require('lodash/fp/filter');
 const find = require('lodash/fp/find');
 const get = require('lodash/fp/get');
 const getOr = require('lodash/fp/getOr');
 const head = require('lodash/fp/head');
 const map = require('lodash/fp/map');
 const pick = require('lodash/pick');
-const sortBy = require('lodash/fp/sortBy');
 
-const request = require('request-promise-native');
+const got = require('got');
 const querystring = require('querystring');
 
+const { CONTENT_API_URL } = require('./secrets');
+const { sanitiseUrlPath, stripTrailingSlashes } = require('./urls');
 const logger = require('./logger');
+
+const queryContentApi = got.extend({
+    prefixUrl: CONTENT_API_URL,
+    headers: { 'user-agent': 'tnlcf-www' },
+    hooks: {
+        beforeRequest: [
+            function(options) {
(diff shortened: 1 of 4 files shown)


common/content-api.js
```diff
@@ -8,59 +10,37 @@ const map = require('lodash/fp/map');
...
- *   return `/v1/${reqLocale}/funding-programmes`
- * }).then(responses => ...)
- * ```
- */
-function fetchAllLocales(toUrlPathFn, options = {}) {
-    const urlPaths = ['en', 'cy'].map(toUrlPathFn);
-    const promises = urlPaths.map(urlPath => fetch(urlPath, options));
-    return Promise.all(promises);
-}
+const queryContentApi = got.extend({
```
davidrapson: The `extend` syntax and hooks for this library makes for a really nice API. No more confusing `fetch` wrapper just so we can log what content API requests are made locally.

common/content-api.js
```diff
@@ -360,7 +398,6 @@ module.exports = {
     // API methods
     getAlias,
     getProjectStory,
-    getProjectStories,
```
davidrapson: `getProjectStories` was never used

common/content-api.js
```diff
@@ -8,59 +10,37 @@ const map = require('lodash/fp/map');
...
-    const promises = urlPaths.map(urlPath => fetch(urlPath, options));
-    return Promise.all(promises);
-}
+const queryContentApi = got.extend({
+    prefixUrl: CONTENT_API_URL,
+    headers: { 'user-agent': 'tnlcf-www' },
+    hooks: {
+        beforeRequest: [
+            function(options) {
+                logger.debug(`Fetching ${options.url.href}`);
```
mattandrews: Does this end up in DataDog? It would massively inflate our usage (eg. we pay more... not a lot more, but still!) if it logs for essentially every webpage request (multiple times in some cases). Or do the `debug` requests only log locally? Either way, would be nice to get the queryParams in here too as they sometimes form the bulk of the useful info (eg. calls to the listing endpoint).

davidrapson: debug only logs locally and `options.url.href` includes the query params. This is essentially a 1-to-1 replacement for what was already there.

PR #2957, opened 2020-02-21
Add new confirmation message for AFA

mattandrews:
Tweaks wording for confirmation screen.

### DIFF
--- a/controllers/apply/awards-for-all/lib/contacts.js
+++ b/controllers/apply/awards-for-all/lib/contacts.js
@@ -1,20 +1,10 @@
 'use strict';
 const get = require('lodash/fp/get');
-const { safeHtml } = require('common-tags');
 
-function getContactFullName(contactData, html = true) {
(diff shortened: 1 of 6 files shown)


controllers/apply/awards-for-all/lib/contacts.js
```diff
@@ -2,14 +2,19 @@
 const get = require('lodash/fp/get');
 const { safeHtml } = require('common-tags');
 
-function getContactFullName(contactData) {
+function getContactFullName(contactData, html = true) {
```
davidrapson: Rather than a boolean could we instead move the HTML to where it's called?

mattandrews: Good shout, can you give this another look?

davidrapson: Yeah, that's way simpler. Cherrs for that!

PR #2968, opened 2020-02-27
Fix path reference for unsubscribe template

mattandrews:
Fixes [this issue](https://sentry.io/organizations/big-lottery-fund/issues/1519880798/?project=226416&referrer=slack) where a template path wasn't right. Means unsubscribes haven't been working for a while :(


davidrapson:
Might give a think about how we can get a Cypress test in for this. Might need a bit of seeding of unsubscribe tokens, but should be possible.

mattandrews:
Yeah would be good. Will get this out now though to restore the functionality!

PR #2979, opened 2020-03-06
Add Publications search for A Better Start

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/218

This adds a new clone of the Insights Document search for the new "Publications" section, which is a new method for publishing arbitrary content pages/document associated with programmes that need to be tagged a certain way and searchable. The initial use-case is A Better Start, but BBO will be able to use this to (to replace the cumbersome resources page).

It looks like this:

image

(copy is TBC)

There's also a change to re-use the blog grid for Strategic programme pages which have the new "related content" field, eg:

image


mattandrews:
Could you give this another look @davidrapson? Made a few tweaks with new copy and redirecting the old pages to the new ones I created this morning.

server.js
```diff
@@ -204,6 +204,18 @@ app.get('/:region?/global-content/programmes/:country/:slug', (req, res) => {
     res.redirect(301, `${locale}/funding/programmes/${req.params.slug}`);
 });
 
+/**
+ * Handle migrated A Better Start child pages as wildcards
+ * (eg. redirect them to /funding/publications/a-better-start/<slug>)
+ */
+app.get('/funding/strategic-investments/a-better-start/:slug', (req, res) => {
+    const locale = req.params.region === 'welsh' ? '/welsh' : '';
```
davidrapson: I'm missing something I don't see a `region` property in the params at all? Should this be using localise instead?

mattandrews: Oh lol, that's from me copy-pasting the bit above. Will fix!

PR #2988, opened 2020-03-09
Phased got content api migration

davidrapson:
With fresh eyes https://github.com/biglotteryfund/blf-alpha/pull/2953 was a bit too wide-reaching for a single PR. Raising this one to do a gradual migration of the content API over to `got`.

### DIFF
--- a/common/content-api.js
+++ b/common/content-api.js
@@ -1,12 +1,10 @@
 'use strict';
-const filter = require('lodash/fp/filter');
 const find = require('lodash/fp/find');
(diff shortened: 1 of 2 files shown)


common/content-api.js
```diff
@@ -228,15 +233,18 @@ function getFundingProgrammes({
     });
 }
 
-function getRecentFundingProgrammes({ locale, limit = 3 }) {
```
davidrapson: This method is only used once so we don't need to pass the same `limit` around

PR #2993, opened 2020-03-09
Add new section to Funding landing page

mattandrews:
Adds this new item to the Funding page:

image

Pages are already live in the CMS but this page itself isn't dynamic... yet.


davidrapson:
Although, should this come first in the list?

PR #3011, opened 2020-03-11
Programmes docs and cleanup

davidrapson:
Same process as https://github.com/biglotteryfund/blf-alpha/pull/3003

- Adds README.md
- Replaces injectCopy with direct i18n use
- Updates related content API methods to use got


.prettierignore
```diff
@@ -1,4 +1,5 @@
-*.json
+package.json
```
davidrapson: Not 100% sure I want to keep this. Wanted to prettier format a json mock file, but it's excluded. This is pretty safe as prettier is an opt-in thing but thoughts welcome.

controllers/funding/README.md
```diff
@@ -0,0 +1,3 @@
+# Funding landing router
+
+Router for the top-level `/funding` page. This is largely a static page with a dynamic list of the latest programmes pulled in from the CMS.
```
mattandrews: My PR isn't merged yet but this will shortly include the "Publications" section. I can add to these docs when this gets merged, though.

PR #3012, opened 2020-03-11
Apply docs

davidrapson:
WIP

Using this as an opportunity to note down any places where we explicitly reference awards for all that we might want to change to make the simple funding form usable by other under £10,000 programmes.


davidrapson:
Will revisit this once this is merged https://github.com/biglotteryfund/blf-alpha/pull/3036

PR #3019, opened 2020-03-13
Fix contact email duplicate bug

mattandrews:
Kind of a funny one, we thought we'd fixed it here: https://github.com/biglotteryfund/blf-alpha/pull/2738

Looking in the database after that change, the only time we encounter duplicate emails are where the **senior** contact uses non-lowercase text and the **main** contact is lowercase.

That's because the change above forces the main contact address to be transformed to lowercase – so if you filled out the form with `Example@foo.com` as Senior Contact, then entered `example@foo.com` as Main Contact, you wouldn't get an error, because Joi doesn't think they match – only the Main Contact gets lowercased.

If you did this the other way around, though (eg. entered `Example@foo.com` as Main Contact and `example@foo.com` as Senior), the Main Contact version gets converted to lowercase and it matches.

The test didn't help catch this because it only tests one of those two paths. This change forces the Senior Contact to lowercase too, and adds a test for the other condition.

### DIFF
--- a/controllers/apply/awards-for-all/__snapshots__/form.test.js.snap
+++ b/controllers/apply/awards-for-all/__snapshots__/form.test.js.snap
@@ -30,7 +30,6 @@ Array [
   "seniorContactDateOfBirth: Enter a date of birth",
   "seniorContactAddress: Enter a full UK address",
   "seniorContactAddressHistory: Enter a full UK address",
-  "seniorContactEmail: Enter an email address",
   "seniorContactPhone: Enter a UK telephone number",
   "bankAccountName: Enter the name of your organisation, as it appears on your bank statement",
   "bankSortCode: Enter a sort code",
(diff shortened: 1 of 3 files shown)


controllers/apply/awards-for-all/__snapshots__/form.test.js.snap
```diff
@@ -30,7 +30,6 @@ Array [
   "seniorContactDateOfBirth: Enter a date of birth",
   "seniorContactAddress: Enter a full UK address",
   "seniorContactAddressHistory: Enter a full UK address",
-  "seniorContactEmail: Enter an email address",
```
mattandrews: Not entirely sure why this changed, but probably fine...

davidrapson: Dug into this and made a small change, added a `.required()` to each. This is not _strictly_ required as the .email() validation acts as a requirement as far as form submissions are concerned but it does make the snapshot tests clearer and is a belt and braces check.

controllers/apply/awards-for-all/fields.js
```diff
@@ -1541,7 +1542,11 @@ module.exports = function fieldsFor({ locale, data = {} }) {
             explanation: localise({
                 en: `We’ll use this whenever we get in touch about the project`,
                 cy: `Byddwn yn defnyddio hwn pan fyddwn yn cysylltu ynglŷn â’r prosiect`
-            })
+            }),
+            schema: Joi.string()
```
davidrapson: If we wanted to we could probably push the `.lowercase()` into the base `EmailField` class to avoid duplicating but this is OK for now.

PR #3025, opened 2020-03-18
Update phone numbers

davidrapson:
Updates static references to country phone numbers with new versions. This is expected to be a temporary change so we can revert this when we need to change back.


controllers/apply/lib/joi-extensions/phone-number.test.js
```diff
@@ -20,10 +20,10 @@ test('return error for invalid phone numbers', () => {
 
 test('format valid phone numbers', () => {
     const validUkPhoneNumbers = [
```
davidrapson: Updated references in test files even though they are not public facing just to avoid any confusion.

PR #3027, opened 2020-03-18
Fixup divergent use of relatedContent macro

davidrapson:
A `usePromo` flag could be passed into the `relatedContent` macro making them behave very differently.

There were only two types of use:
1. homepage related news
2. flexible content related entries.

Noticed that the related-content classes were not even needed for the flexible content use-case. It can just be a flex-grid. So I've been able to simplify this to:

1. Inline the related-content markup into the homepage for that use-case
2. Simplify the flexible content use-case to a set of flex-grid classes
3. Delete the relatedContent macro

In the process I've added a post date to the news items on the homepage:

Screenshot 2020-03-18 at 16 57 01


views/components/flexible-content-next/macro.njk
```diff
@@ -45,11 +45,22 @@
...
-                            articles = part.content,
-                            title = part.heading,
-                            usePromo = true
-                        ) }}
+                        <h2 class="t--underline">{{ part.heading }}</h2>
+                        <ul class="flex-grid flex-grid--3up">
+                            {% for article in part.content %}
+                                <li class="flex-grid__item">
+                                    {{ promoCard({
+                                        "title": article.title + "floop",
```
mattandrews: er... should `floop` be in here?

views/components/flexible-content/macro.njk
```diff
@@ -14,11 +14,22 @@
...
-            articles = part.content,
-            title = part.heading,
-            usePromo = true
-        ) }}
+        <h2 class="t--underline">{{ part.heading }}</h2>
+        <ul class="flex-grid flex-grid--3up">
+            {% for article in part.content %}
+                <li class="flex-grid__item">
+                    {{ promoCard({
+                        "title": article.title + "floop",
```
mattandrews: image

mattandrews: birthdayfloop

mattandrews: Lotteryfloop

mattandrews: Also those bullet points look like a regression, there's some gnarly CSS to fix that (`.related-content .related-content__item`)

davidrapson: Hmm which page type was that on?

mattandrews: The 25th fund page: `/about/the-national-lotterys-25th-birthday`

davidrapson: Worked it out, this page was using the _old_ flexible content template, where the flex-grid was still inside a `s-prose` area. Made a root cause fix by resetting the list-style and margin explicitly, should be pretty safe but worth doing a scan of the site to confirm this doesn't cause any regressions.

assets/sass/scaffolding/_grid.scss
```diff
@@ -29,6 +29,8 @@ $flexGridSpacing: $spacingUnit;
     > .flex-grid__item {
         max-width: 100%;
         margin-bottom: $flexGridSpacing;
+        list-style: none !important;
```
davidrapson: Made a root cause fix to account for flex-grid being inside `s-prose` areas. Resetting the list-style and margin explicitly. Should be pretty safe but worth doing a scan of the site to confirm this doesn't cause any regressions.

davidrapson:
image

PR #3037, opened 2020-03-19
minimist audit fix and set audit-level

davidrapson:
We have a security advisory for `minimist` (https://www.npmjs.com/advisories/1179) which is depended on by many of our node_modules. 

This is a low severity advisory and has a couple of modules that are still pending a fix. Attempting to balance running npm audit in report only mode (i.e. don't fail the build) which we do now vs. blocking too aggressively. 

This PR tests setting `--audit-level=moderate` to only fail the build if there is a moderate level advisory or above in our production dependencies.

We can confirm this works if the build passes but we see three low level security advisories in the build output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ cache:
 
 script:
   - shellcheck deploy/*.sh
-  - npm audit --production || true
+  - npm audit --production --audit-level=moderate
   - npm run lint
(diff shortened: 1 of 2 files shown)


davidrapson:
This worked as expected. Build passed, audit output still listed low severity items as reference.

image

PR #3039, opened 2020-03-19
Consistent use of console logger during test runs

davidrapson:
Sometimes seeing local tests runs try and fail to log to cloudwatch.

After doing some digging this is because the node environment is set to `test` which also determines which config file to use. This means that cloudwatch logs are enabled as the app uses the same config as the test server.

The longer term fix for this would be to re-configure the staging site to used something other than test as the environment. Basically: https://glebbahmutov.com/blog/do-not-use-node-env-for-staging/

In the short term if rejigged the logger config to include a check for an explicit `IS_TEST_RUN` variable to make sure the console logger is always used in these cases.

### DIFF
--- a/common/logger.js
+++ b/common/logger.js
@@ -5,16 +5,30 @@ const WinstonCloudWatch = require('winston-cloudwatch');
 
 const { environment, buildNumber, isTestServer, isDev } = require('./appData');
 
-function enableCloudWatchLogs() {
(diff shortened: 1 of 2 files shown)


davidrapson:
Realised this doesn't apply if running jest directly, so going to take another look at approaches for this. Might be better to just try and fix the root cause here.

PR #3061, opened 2020-03-24
Upgrade to Cypress 4

davidrapson:
Branched off https://github.com/biglotteryfund/blf-alpha/pull/3060

Cypress 4 introduces some initial cross browser support beyond Chrome. Chromium based Edge is supported and Firefox is in beta. It also includes upgrades to the bundled versions of chai/mocha which in theory would have breaking changes but we're not using anything more than the basic syntax so it all works fine.

This PR makes no attempt to run the tests in other browsers within CI, but I have been able so successfully run all the tests locally in Firefox.

Once we've merged this and have been running the new version for a few builds we can decide how/if we want to test in other browsers in CI.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5183,6 +5183,15 @@
         "fastq": "^1.6.0"
       }
     },
+    "@samverschueren/stream-to-observable": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.0.tgz",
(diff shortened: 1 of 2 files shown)


davidrapson:
@dependabot-preview why did you close this??

davidrapson:
Worked out that linking this to `https://github.com/biglotteryfund/blf-alpha/tree/dependabot/npm_and_yarn/cypress-file-upload-4.0.4` triggered some of the dependabot behaviour. Good to know.

PR #3062, opened 2020-03-25
Fixup spacing on covid-19 banner

davidrapson:
Uses `.has-static-header` class to adjust the spacing.

image

image


assets/sass/components/_messages.scss
```diff
@@ -143,3 +139,11 @@
         display: none;
     }
 }
+
+.announcement-banner.announcement-banner--major {
+    font-size: 16px;
```
davidrapson: Bumped this up so the text is at least the same size as body copy.

PR #3076, opened 2020-03-26
Cross env + tests added

mo-durvesh:
Changes for compatibility on Windows for the npm run start-test command using cross-env.

Added two tests to the integration.spec: an invalid password test for non matching passwords and an invalid email address test for an incorrect format.


package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "start": "node ./bin/www",
     "start-dev": "nodemon --inspect ./bin/www",
-    "start-test": "DB_CONNECTION_URI=sqlite://:memory PORT=8090 TEST_SERVER=true npm run start",
+    "start-test": "cross-env DB_CONNECTION_URI=sqlite://:memory PORT=8090 TEST_SERVER=true npm run start",
```
davidrapson: 💪 Nice! Glad this sorted things for you!

package.json
```diff
@@ -97,6 +97,7 @@
     "config": "3.3.0",
     "connect-session-sequelize": "6.0.0",
     "connect-slashes": "^1.4.0",
+    "cross-env": "7.0.2",
```
davidrapson: Mind installing this under `devDependencies`? You can do this by running `npm install --save-dev cross-env` (you may need to uninstall the current version first). This makes sure we don't ship the dependency to production if we don't need it there.

cypress/integration/integration_spec.js
```diff
@@ -260,6 +260,10 @@ it('should prevent registration with invalid passwords', () => {
     createAccount(username, '5555555555');
     cy.findByTestId('form-errors').should('contain', 'Password is too weak');
 
+    cy.visit('/user/register');
```
davidrapson: 😍

cypress/integration/integration_spec.js
```diff
@@ -278,6 +282,15 @@ it('should prevent registration with invalid passwords', () => {
     });
 });
 
+it('should prevent registration with an invalid email address', () => {
```
davidrapson: 😍

package.json
davidrapson: (just ran `npm install --save-dev cross-env` locally and confirms that it cleans things up on its own, no no other step needed)

PR #3081, opened 2020-03-27
Upgrading jest and babel-jest together to 25.2.3

davidrapson:
Upgrading jest and babel-jest together to 25.2.3

In place of https://github.com/biglotteryfund/blf-alpha/pull/3080. Looks like jest and babel-jest version are kept in sync so made sense to bump both at once.


davidrapson:
Closes https://github.com/biglotteryfund/blf-alpha/pull/3082

PR #3095, opened 2020-03-30
Remove remaining request uses in content-api

davidrapson:
Now uses `got` via `queryContentApi` across the board.


davidrapson:
@mattandrews Only place I couldn't work out how to test properly was the publications. How do I check those?

mattandrews:
With a current version of the production CMS data, visit `/funding/publications/a-better-start`

PR #3106, opened 2020-03-30
[WIP] Migrate salesforce queries to use got

davidrapson:
This allows us to full migrate away from request in favour of got. The main difference here is needing to construct the form-data object directly rather than relying on requests wrapping of it. Aside from this, the overall result is a little simpler.

Definitely considering this work-in-progress as it needs a fair bit of testing against Salesforce, but overall the changes feel straightforward. The file-attachment step is the main bit that needs more experimentation.


controllers/apply/form-router/lib/salesforce.js
```diff
@@ -1,78 +1,69 @@
...
+        form.append(
+            'entity_content',
+            JSON.stringify({
+                FirstPublishLocationId: recordId,
+                ReasonForChange: `Application attachment ${attachmentName}`,
+                PathOnClient: attachmentName,
+            }),
+            { contentType: 'application/json' }
+        );
+        form.append('VersionData', versionData.value, versionData.options);
```
davidrapson: This makes me think that much of the code in `buildMultipartData` should live in here rather than in the file-uploads module as the output is specific to what is needed for form-data. But for now making the minimum changes needed.

package-lock.json
```diff
@@ -15251,6 +15264,16 @@
             "uuid": "^3.3.2"
           },
           "dependencies": {
+            "form-data": {
```
davidrapson: This is a nested dependency of `cypress` which uses `request` somewhere under the hood. Really just npm doing dedupe work.

package-lock.json
```diff
@@ -20049,6 +20072,16 @@
         "uuid": "^3.3.2"
       },
       "dependencies": {
+        "form-data": {
```
davidrapson: Likewise, this is a nested dependency of `jsdom` which uses `request` somewhere under the hood.

controllers/apply/form-router/lib/salesforce.js
```diff
@@ -1,78 +1,69 @@
...
+const FormData = require('form-data');
 const { SALESFORCE_AUTH } = require('../../../../common/secrets');
 
 class Salesforce {
     constructor(apiUrl, accessToken) {
-        this.apiUrl = apiUrl;
         this.apiVersion = 'v45.0';
-        this.headers = { Authorization: `Bearer ${accessToken}` };
-    }
-    get(urlPath) {
```
davidrapson: Got rid of this get helper as a) we don't use it and b) even if we did it's no longer any more code than using the got instance directly.

controllers/apply/form-router/lib/salesforce.js
```diff
@@ -1,78 +1,69 @@
...
-        return request.get({
-            uri: `${this.apiUrl}${urlPath}`,
-            headers: this.headers,
-            json: true,
-        });
-    }
-    postJson(urlPath, body) {
```
davidrapson: Inlined this as using the got instance directly is now a similar amount of code to using this abstraction, and basically doing the same thing but at a lower level.

controllers/apply/form-router/lib/salesforce.js
```diff
@@ -1,78 +1,69 @@
...
-                    }),
-                    options: {
-                        contentType: 'application/json',
-                    },
-                },
-                VersionData: versionData,
-            },
-        });
+
+    contentVersion({ recordId, attachmentName, versionData }) {
```
davidrapson: Removed the `async` from this as we don't use `await` inside the body of this method, so it isn't needed.

davidrapson:
Needs a bit more testing in sandbox environment, will reopen this soon.

PR #3114, opened 2020-04-01
Ensuring every main element has id="content" (for skip links).

mo-durvesh:
id="content" added to the main tags on several pages where not already existing.

### DIFF
--- a/controllers/home/views/home.njk
+++ b/controllers/home/views/home.njk
@@ -24,7 +24,7 @@
                 {% endif %}
             </figure>
             <div class="super-hero__content">
-                <div class="super-hero__header" id="content">
+                <div class="super-hero__header">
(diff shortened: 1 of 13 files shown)


controllers/home/views/home.njk
```diff
@@ -7,7 +7,7 @@
 {% set title = copy.title %}
 
 {% block content %}
-    <main role="main">
+    <main role="main" id="content">
```
davidrapson, line 10: Looks like the Cypress tests are failing on the home page because we have two `content` IDs. If you search further down this page you should find there's another ID on a later element, we can probably remove that one if favour of this new one for consistency.

mo-durvesh: Fixed, thanks!

PR #3119, opened 2020-04-02
TODO review

davidrapson:
Reviewing any TODOs in the code. Resolving anything small, deleting things that are non-actionable, and will move anything else to an issue.

### DIFF
--- a/assets/sass/components/_logos.scss
+++ b/assets/sass/components/_logos.scss
@@ -2,8 +2,6 @@
    Logo
    ========================================================================= */
 
-// Duplicates .card a little
-// @TODO see if it makes sense to merge this as a .card variant in future
 .logo-card {
     padding: 0;
     background-color: get-color('background', 'light-neutral');
--- a/common/locals.js
+++ b/common/locals.js
@@ -183,15 +183,12 @@ module.exports = function (req, res, next) {
      * @param dateString
      */
     res.locals.formatCalendarTime = function (dateString) {
-        return (
-            moment(dateString)
-                .tz('Europe/London')
-                // @TODO: Handle this consistently across date formatters
-                .locale(locale === 'en' ? 'en-gb' : 'cy')
-                .calendar(null, {
-                    sameElse: 'D MMMM, YYYY',
-                })
-        );
+        return moment(dateString)
+            .tz('Europe/London')
+            .locale(locale === 'en' ? 'en-gb' : 'cy')
+            .calendar(null, {
+                sameElse: 'D MMMM, YYYY',
+            });
     };
 
     /**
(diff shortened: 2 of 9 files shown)


assets/sass/components/_logos.scss
```diff
@@ -2,8 +2,6 @@
    Logo
    ========================================================================= */
 
-// Duplicates .card a little
-// @TODO see if it makes sense to merge this as a .card variant in future
```
davidrapson, line 6: This is OK. Better to duplicate than create an awkward abstraction

common/locals.js
```diff
@@ -183,15 +183,12 @@ module.exports = function (req, res, next) {
      * @param dateString
      */
     res.locals.formatCalendarTime = function (dateString) {
-        return (
-            moment(dateString)
-                .tz('Europe/London')
-                // @TODO: Handle this consistently across date formatters
```
davidrapson, line 189: Double checked the original issue and this check is only needed on the `.calendar` method as it's the one place where date formats are generated (rather than explicitly defined).

controllers/apply/standard-proposal/fields.js
```diff
@@ -149,7 +149,7 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
              * pending proposals that used checkbox selection.
              *
              * Multiple selection disabled until UK-portfolio is enabled.
-             * @TODO: Remove and switch back to checkbox when launching UK-portfolio
```
davidrapson, line 152: Removed the TODO as this is more a longer-term product decision when we launch the UK-portfolio. Nothing _actionable_ in this code.

controllers/apply/under10k/fields/project-postcode.js
```diff
@@ -8,9 +8,6 @@ const Field = require('../../lib/field-types/field');
 module.exports = function (locale) {
     const localise = get(locale);
 
-    /**
-     * @TODO: Is it worth this being a dedicated PostcodeField type?
```
davidrapson, line 12: Not worth adding this abstraction at this point, maybe if we build another form with a single post-code field.

db/models/applications.js
```diff
@@ -155,17 +156,17 @@ class PendingApplication extends Model {
...
         metadata = null,
     }) {
-        // @TODO: Should this be defined in config?
-        const expiresAt = customExpiry
-            ? customExpiry
-            : moment().add('3', 'months').toDate();
+        function expiresAt() {
+            const [amount, unit] = config.get('applications.expiryPeriod');
+            return moment().add(amount, unit).toDate();
+        }
```
mattandrews, line 162: 💯

PR #3127, opened 2020-04-02
Add TOC titles

mattandrews:
Allows custom Table of Contents titles to be used instead of a Flexible Content's block `title` element, if provided. Means we can shorten lengthy lists like these:

image

Pairs with [a CMS change which adds these fields](https://github.com/biglotteryfund/craft-dev/pull/241). Also fixes a previewing issue which would throw 500 errors when a string was `null`.

### DIFF
--- a/views/components/segment-links/macro.njk
+++ b/views/components/segment-links/macro.njk
@@ -12,20 +12,3 @@
         </div>
     {% endif %}
 {% endmacro %}
-
-
-{% macro tableOfContentLinks(flexibleContent, prefix = null, anchor = 'segment') %}
-    {% if flexibleContent.length > 1 %}
-        <div class="segment-links">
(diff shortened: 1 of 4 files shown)


views/components/segment-links/macro.njk
```diff
@@ -12,3 +12,20 @@
         </div>
     {% endif %}
 {% endmacro %}
+
+
+{% macro tableOfContentLinks(flexibleContent, prefix = null, anchor = 'segment') %}
```
davidrapson, line 17: Can we make this a new file / directory please? Should only really be one macro inside each of these.

mattandrews: Done!

davidrapson:
:shipit:

PR #3132, opened 2020-04-03
Static image cleanup

davidrapson:
🏕 Being a good camper. As I've just added some extra images I should remove some old ones left behind.


config/locales/cy.yml
```diff
@@ -225,87 +225,6 @@ toplevel:
           rhaglen benodol ar ein gwefan ddydd Llun.
         </p>
       linkText: Darllenwch ein diweddariadau COVID-19 diwethaf
-  wales:
```
mattandrews, line 228: I can't believe I forgot to delete this when I moved these pages to the CMS. You just stole these 200 lines of deletion from me.

davidrapson: sorry not sorry

PR #3133, opened 2020-04-06
Upgrade @testing-library/cypress to 6.0.0

davidrapson:
With @testing-library/cypress upgrade the approach for checking the terms checkboxes was breaking. Likely because of a small change in how find* queries are chained. Turns out .click takes an option of `{ multiple: true }` which is a much simpler API and fixes the tests.

### DIFF
--- a/cypress/integration/integration_spec.js
+++ b/cypress/integration/integration_spec.js
@@ -924,17 +924,13 @@ it('should submit full application for under £10,000', () => {
     function sectionTermsAndConditions() {
         cy.checkA11y();
 
-        cy.findAllByLabelText('I agree').each(($el) => {
-            cy.wrap($el).click();
-        });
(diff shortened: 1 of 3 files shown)


cypress/integration/integration_spec.js
```diff
@@ -924,17 +924,13 @@ it('should submit full application for under £10,000', () => {
     function sectionTermsAndConditions() {
         cy.checkA11y();
 
-        cy.findAllByLabelText('I agree').each(($el) => {
-            cy.wrap($el).click();
-        });
+        cy.findAllByLabelText('I agree').click({ multiple: true });
 
-        cy.findByLabelText('Full name of person completing this form', {
```
davidrapson, line 931: Removed the `exact: false` here because it's not needed.

PR #3139, opened 2020-04-06
Applications dashboard cleanup

davidrapson:
Adds mapping object, allows us to avoid hard-coding the form id multiple times and instead returns a single object for the config that varies based on form ID.


davidrapson:
[Update dashboard to reference "Apply for funding under £10,000"](https://trello.com/c/JuNzwk6K/1104-update-dashboard-to-reference-apply-for-funding-under-10000)

PR #3141, opened 2020-04-06
Filters cleanup

davidrapson:
Noticed that quite a lot of our nunjucks filters were unused. Used this as an excuse to clean things up and streamline the test synxtax.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
   ],
   "eslintIgnore": [
     "public/",
-    "views/",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -72,7 +72,6 @@
   ],
   "eslintIgnore": [
     "public/",
-    "views/",
```
davidrapson, line 75: This was preventing `views/filter.js` from being linted

PR #3149, opened 2020-04-07
Update table of contents logic and macros

davidrapson:
Noticed that the page that prompted the new `tocTitle` change (https://www.tnlcommunityfund.org.uk/about/covid-19-how-we-are-supporting-our-grant-holders-at-this-difficult-time) wasn't actually using the  new table of contents macro. Looks like it's using an older page type (rendering `information-page.njk`).

Went through the templates and:

- Standardised `tableOfContents` and cleaned up the macro
- Inlined the one exceptional use case into the insights document template
- Renamed the CSS class-name to be `table-of-contents` instead of the confusing `segment-links`


views/components/table-of-contents/macro.njk
```diff
@@ -1,10 +1,9 @@
...
-    {% if flexibleContent.length > 1 %}
-        <div class="segment-links">
-            {% if prefix %}<p class="u-margin-bottom-s">{{ prefix }}</p>{% endif %}
+{% macro tableOfContents(items, anchor) %}
+    {% if items.length > 1 %}
+        <div class="table-of-contents">
             <ul>
-                {% for block in flexibleContent %}
-                    {% set titleToUse = block.tocTitle or block.title %}
+                {% for item in items %}
```
mattandrews, line 5: I think I preferred the original version here because it's clearer that it expects to operate on a block of Flexible Content?

davidrapson: ah you are right, think we can switch that back. I did that originally because it didn't _always_ operate on flexible content, but I think it does now!

PR #3152, opened 2020-04-08
Add log out link to user dashboard

davidrapson:
Occasionally getting reports of users missing the log out link in the header (either due to not seeing it, or potentially it not switching from the default state). A bit belt and braces but adding a log out link to the account screen.

Screenshot 2020-04-08 at 10 44 22


davidrapson:
[Add log out option to the user account screen](https://trello.com/c/P9Gh79uc/1114-add-log-out-option-to-the-user-account-screen)

PR #3156, opened 2020-04-08
Live preview check fix

mo-durvesh:
Updated the check for if it is a live preview, it will only require the x-craft-live-preview or the token rather than both. Also implemented the same for the share link.


mo-durvesh:
> The change itself looks good but you've got a failing test ("it should protect access to staff-only tools") so can you have a look at what's breaking that?

Found the reason to be; if both token and xcraft-live-preview were present, it would only return one because of the OR statement. An if statement was added to check if both are present which would return both and if not; check which one specifically is present to return the correct value.

PR #3162, opened 2020-04-09
Experimental: Router structure

davidrapson:
This is an experimental branch to explore simplifying how we mount section routers. Prompted by two things:

1. Noticed that the publications was in the "wrong" place, inlined in `funding/index.js` where other sections—like programmes, and past grants—have their own directories. But on reflection the publications is in a perfectly reasonable place and if anything the top level grants directory is more confusing.
2. When Mark, Josh, and Mo each had a go at adding new routers the most natural thing to do was to add it to an existing directory. This is the right call and points to our code being confusing if we have multiple ways to do this. 

Between these two points it's fairly clear that the way of mounting new routers in `routes.js` in unintuitive. More to the point the `pages` array doesn't even map to _pages_ it's just an abstraction on top of normal express routing.

This PR is an experiment to try restructuring things so we don't need the pages array and routers are structured around the sections of the site. It's also been a good exercise to see which pages really need to be migrated to other parts of the CMS.

**Structure before**
image

**Structure after:**
image

This branch is not intended to be merged but @mattandrews if you wouldn't mind checking this branch out and seeing if it makes more or less sense than before.


server.js
```diff
@@ -187,56 +160,68 @@ app.use([
...
 /**
- * Initialise section routes
- * - Creates a new router for each section
- * - Apply shared middleware
- * - Apply section specific controller logic
- * - Add common routing (for static/fully-CMS powered pages)
+ * Initialise section routers
  */
-forEach(routes, function (section, sectionId) {
-    const router = express.Router();
```
davidrapson, line 233: This is the big change, the "routes" definition is now a flatter list of sections and routers. Jobs and data are still top level routers, but that's just a sign that we need to move them later.

controllers/about/our-people/views/profiles.njk
```diff
@@ -50,7 +50,7 @@
                     {# @TODO: Should we share this component? #}
                     <nav class="nested-menu">
                         <h2 class="nested-menu__section t4">
-                            <a class="nested-menu__section-link" href="./">
+                            <a class="nested-menu__section-link" href="../../../..">
```
davidrapson, line 53: This is a bit of auto-refactoring weirdness. Not going to merge this branch so can ignore for now.

controllers/funding/materials/views/materials.njk
```diff
@@ -2,7 +2,7 @@
 {% from "components/icons.njk" import iconClose %}
 {% from "./materials-fields.njk" import textInput, radios with context %}
 
-{% extends "controllers/common/views/cms-page.njk" %}
+{% extends "../../../common/views/cms-page.njk" %}
```
davidrapson, line 5: Again, auto refactoring weirdness. The previous path works fine due to how nunkucks resolves paths.

mattandrews: >nunkucks
image

controllers/funding/index.js
```diff
@@ -2,101 +2,72 @@
...
-        } catch (error) {
-            next(error);
-        }
-    }
+router.use('/grants', require('./grants'));
+
+// @TODO: Can this page be migrated to Funding structure in the CMS?
+router.use('/the-big-lunch', basicContent());
+
+// @TODO: Can this section be migrated to Funding structure in the CMS?
```
mattandrews, line 36: I think most/all of the pages in this section are in Funding and/or out of date, so a big fat YES from me.

server.js
mattandrews: Probably Contact too? Ideally we'd have no "top level" pages except the homepage. Could probably live in About?

davidrapson:
Closing, but will raise some smaller PRs to re-do this.

PR #3166, opened 2020-04-09
Remove injectCopy from user routes

davidrapson:
Removes injectCopy from user routes in favour of using translation methods directly.

Like with previous removals this reduces the overhead of working out what "copy" refers to in this context. Particularly notable here as there were nested injectCopy calls so copy could either be the top-level user locale or a more specific one (always seemed to be overridden but still possible.)

### DIFF
--- a/controllers/user/index.js
+++ b/controllers/user/index.js
@@ -8,7 +8,6 @@ const { isDev } = require('../../common/appData');
 const { localify } = require('../../common/urls');
 const { noStore } = require('../../common/cached');
 const { requireNotStaffAuth } = require('../../common/authed');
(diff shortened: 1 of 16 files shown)


controllers/user/index.js
```diff
@@ -74,7 +73,7 @@ router.get('/session', function (req, res) {
  * Public user routes
  * Disallow staff from this point on
  */
-router.use(requireNotStaffAuth, injectCopy('apply'), function (req, res, next) {
```
davidrapson, line 77: This was the most egregious one, this was set here. Making `res.locals.copy` equal the `apply` locale key…which was then never used and consistently overridden with a `user.*` key in individual routers.

PR #3172, opened 2020-04-09
Make sure checkPreviewMode returns booleans, not values

davidrapson:
Thanks to the test @mo-durvesh added in https://github.com/biglotteryfund/blf-alpha/pull/3156 I spotted that the current behaviour of `checkPreviewMode` was returning the _values_ for the checks rather than a boolean. 

Technically this works because our code does truthy checks on these values but is unexpected behaviour. This PR fixes this and updates the tests.

### DIFF
--- a/common/check-preview-mode.js
+++ b/common/check-preview-mode.js
@@ -1,12 +1,13 @@
 'use strict';
+const has = require('lodash/has');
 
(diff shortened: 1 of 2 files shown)


davidrapson:
@mo-durvesh Just to clarify, your change was good, your test just exposed an existing quirk in the original behaviour.

PR #3176, opened 2020-04-14
Handle 404 errors correctly for Insights pages

mattandrews:
Seeing a bit of log noise over the weekend about dodgy Insights page URLs which return 500 errors. This uses the same approach as other controllers to return a 404 when the content is missing.

### DIFF
--- a/controllers/insights/index.js
+++ b/controllers/insights/index.js
@@ -110,7 +110,6 @@ router.get('/:slug/:child_slug?', async function (req, res, next) {
             next();
         }
     } catch (error) {
-        console.log(error.response.statusCode);
         if (error.response.statusCode >= 500) {
             next(error);
         } else {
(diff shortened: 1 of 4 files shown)


controllers/insights/index.js
```diff
@@ -71,7 +71,11 @@ router.get(
                 pagination: research.pagination,
             });
         } catch (error) {
-            next(error);
+            if (error.statusCode >= 500) {
```
davidrapson, line 74: I _think_ this might need to change, started taking a look at this on Thursday and noticed that this response format changed slightly with `got`. This might need to be something like `error.response.statusCode`

mattandrews: Hmm, it worked locally when I tried it...?

controllers/insights/index.js
```diff
@@ -110,7 +110,8 @@ router.get('/:slug/:child_slug?', async function (req, res, next) {
             next();
         }
     } catch (error) {
-        if (error.statusCode >= 500) {
+        console.log(error.response.statusCode);
```
davidrapson, line 113: This meant to be here?

PR #3179, opened 2020-04-14
Top level routers

davidrapson:
Builds on from https://github.com/biglotteryfund/blf-alpha/compare/top-level-routers to do the remaining work to move over to structure described in #3162

Migrates remaining top-level routers to new structure. Main change is to introduce a new `about/` directory to contain all related routing code.


controllers/about/index.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+const express = require('express');
+
+const { basicContent, flexibleContent } = require('../common');
+
+const router = express.Router();
+
+router.get('/', flexibleContent());
+// @TODO: Can we move the our people section to About structure? Add "profile" type to flexible content
```
davidrapson, line 9: We already have a thorny page at https://www.tnlcommunityfund.org.uk/about/our-people/advisory-group-membership which is a good argument to migrate these pages to flexible content with a new "profile" content type.

mattandrews: Could be a task for the IT team?

PR #3186, opened 2020-04-15
Refactor common CMS routing

davidrapson:
Does some work to clean up how CMS pages are rendered and reduce tight coupling between new and legacy types of CMS pages. Tried to keep each commit granular so worth looking at each step.

- ~Inlines misleadingly named `injectFlexibleContent` code. This method _looked_ like it was generic but is actually only used once, for the about landing page. To make this behaviour clear, and to allow us to simplify the rest of the shared template code, moving this behaviour into a dedicated router~ Edit: Realised we can update the about landing page to use the newer flexible content template and remove this code completely.
- Merged cms-page and flexible-content templates, after extracting the about page they ended up being idendical
- Add a logger for legacy content pages so we can get some stats on usage to help migration (might be easier to do this by auditing the CMS?)
- Removed `customTemplate` from `basicContent`. Only used for logos page. Removing this and inlining the template into funding/ allows us to a) remove the flag b) handle logos page the same as materials c) remove the concept of "views/static-pages" so we don't have to explain why it exists for only one page.
- Split flexibleContentPage from legacy basicContent rather than using a parameter. We'll know we've migrated everything when pages are only using flexibleContentPage.


controllers/about/index.js
```diff
@@ -1,11 +1,35 @@
...
+const contentApi = require('../../common/content-api');
+const { setCommonLocals } = require('../../common/inject-content');
+const { basicContent } = require('../common');
 
 const router = express.Router();
 
-router.get('/', flexibleContent());
+router.get('/', async function (req, res, next) {
+    try {
+        const entry = await contentApi.getFlexibleContent({
```
davidrapson, line 13: This misleadingly named `getFlexibleContent` is only ever used on the about single. Inlined the code here. Likely another task on the CMS to either scope this API method to the about single or migrate to the newer flexible content type.

davidrapson:
Realised we can update the about landing page to use the newer flexible content template and remove the confusing `getFlexibleContent` API method compeltely.

Pages now looks like this:

Screenshot 2020-04-12 at 15 08 50

controllers/common/index.js
```diff
@@ -37,37 +34,28 @@ function basicContent({ customTemplate = null, cmsPage = false } = {}) {
...
+         * @TODO: Deprecate these templates
          */
-        if (customTemplate) {
-            res.render(customTemplate);
-        } else if (cmsPage) {
-            res.render(path.resolve(__dirname, './views/cms-page'));
-        } else if (content.children) {
-            // @TODO: Deprecate these templates in favour of CMS pages (above)
+        if (content && content.children) {
+            logLegacyContentType('listing-page');
```
mattandrews, line 51: 💯

PR #3191, opened 2020-04-15
Cleanup tests and single-route endpoints

davidrapson:
Makes any single-router handlers under `controllers/` follow the same pattern of just exporting a handler (not a router). Avoids nesting routers for a single method and allows to more easily write unit tests for each endpoint.


cypress/integration/integration_spec.js
```diff
@@ -70,17 +70,6 @@ it('should 404 unknown routes', () => {
     check404('/not/a/page');
 });
 
-it('should redirect search queries to a google site search', () => {
```
davidrapson, line 73: This test has been moved to a unit test with jest. Made more sense as we don't invoke a browser UI here.

controllers/search.js
```diff
@@ -1,12 +1,7 @@
...
-const express = require('express');
 
-const { noStore } = require('../common/cached');
-
-const router = express.Router();
-
-router.get('/', noStore, (req, res) => {
+module.exports = function (req, res) {
     if (req.query.q) {
         const term = querystring.escape(req.query.q);
```
mattandrews, line 6: While we're here, I wonder if we should log this somewhere (Datadog)? We don't have very good insight into this here as it's not tracked via GA and Google Search Console is a bit opaque sometimes.

PR #3194, opened 2020-04-15
Improve application form documentation

davidrapson:
- Moves docs to top level `docs/` folder. This high level documentation is meant to be for a wider audience, and whilst I _like_ colocated documentation the previous location was _too_ colocated and hard to find unless you are a developer with some familiarity with our existing code structure.
- Adds a new `questions.md` document for the over 10k form which is basically a condensed summary of all the questions that are asked and their validation rules. Written for a general audience.
- Adds a really rough overview document with a few high-level bullet points. Needs improving, but this is a start
- Merges the schema tests, duplicate test code so made one test file to cover both documents. And avoids people browsing docs folders needing to worry about what a `.js` file does.

💭 Thoughts very welcome


davidrapson:
[Produce high-level overview document for online forms](https://trello.com/c/ZANbe43V/1120-produce-high-level-overview-document-for-online-forms)

bin/generate-form-docs
```diff
@@ -0,0 +1,49 @@
...
+function generateFormDocs(formBuilder, directory) {
+    const form = formBuilder();
+
+    const fieldSummary = Object.values(form.allFields)
+        .map(function (field) {
+            return stripIndents`
+                ## ${field.name}
+
+                | Rule           | Message              |
+                | -------------- | -------------------- |
```
mattandrews, line 16: If you wanted to be a mega nerd here you could find the longest `item.type` in the list and make the Rule dashes here long enough to accomodate it, or does Markdown do that for you regardless?

mattandrews: Update, never mind, looks like it does!

davidrapson: Yep, plus prettier can tidy it all up afterwards too

PR #3196, opened 2020-04-16
Remove array.includes polyfill

davidrapson:
Rolled back vue-i18n to fix root cause rather than symptom in https://github.com/biglotteryfund/blf-alpha/pull/3193. To avoid any future confusion removing this unneeded polyfill.


mattandrews:
Nice 👍

PR #3203, opened 2020-04-17
Add homepage welsh language regression test

davidrapson:
Follow on from https://github.com/biglotteryfund/blf-alpha/pull/3202

Add a smoke test for the welsh language version of the homepage to avoid regressions.

### DIFF
--- a/cypress/integration/integration_spec.js
+++ b/cypress/integration/integration_spec.js
@@ -15,8 +15,25 @@ function acceptCookieConsent() {
 
 it('should test common interactions', () => {
     cy.visit('/');
-
     cy.checkA11y();
+
+    // Switch to welsh language version of homepage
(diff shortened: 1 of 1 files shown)


cypress/integration/integration_spec.js
```diff
@@ -15,9 +15,26 @@ function acceptCookieConsent() {
...
+    // When people are in the lead, communities thrive
+    cy.findByText('Rydym yn cefnogi pobl a chymunedau i ffynnu').should(
+        'be.visible'
+    );
+
+    // Switch back to English
+    cy.get('.qa-lang-switcher').click();
+
+    // Check welsh language text
+    // When people are in the lead, communities thrive
```
mattandrews, line 33: Quite like this just being a comment in the source but probably don't need it twice?

PR #3204, opened 2020-04-17
Update homepage to be a single top-level GET route

davidrapson:
Avoids homepage being mounted using app.use with covers paths _starting_ with / and allows the homepage router to inadvertently be used for other pages. Using a  single GET handler resolves this.

### DIFF
--- a/server.js
+++ b/server.js
@@ -17,7 +17,7 @@ module.exports = app;
 
 const appData = require('./common/appData');
 const { SENTRY_DSN } = require('./common/secrets');
-const { localify, makeWelsh } = require('./common/urls');
+const { localify } = require('./common/urls');
 const { defaultMaxAge, noStore, sMaxAge } = require('./common/cached');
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -164,7 +164,14 @@ app.use('/tools', require('./controllers/tools'));
 app.use('/patterns', require('./controllers/pattern-library'));
 
 /**
- * Handle archived paths first:
+ * Homepage handler
+ * Mount as a one-off GET route in both English and Welsh
+ */
+app.get('/', require('./controllers/home'));
```
mattandrews, line 170: Would this be better as:

`app.get(['/', '/welsh'], require('./controllers/home'))`?

(eg. avoiding the duplication)

Or is it better explicit like here?

davidrapson: Possibly, but that array syntax via `app.use` caused the original but so opted to be explicit.

PR #3210, opened 2020-04-20
Remove transformOrgHasDifferentTradingName

davidrapson:
It has been three months since we launched this change so this should no longer be needed. Metrics show this was last called on 15/04/2020. We can probably leave this a couple of days to confirm, but should be fairly safe to go.


davidrapson:
Thinking about this the expiry date for the other transform is under two weeks away. Think I'll hold off on this until we can see all transforms drop off and then remove this whole process.

PR #3212, opened 2020-04-20
Update .compare() validation to compare normalised values

davidrapson:
Spotted our `.compare()` validation is a little lax. Previous behaviour was to do a shallow equality check on the objects. This is fine if the object has exactly the same values e.g.

```js
isEqual({ a: 'one', b: 'two' }, { a: 'one', b: 'two' }) // true
```

But fails for cases where the keys contained extra spaces or a mix of cases:

```js
isEqual({ a: ' oNe ', b: 'twO' }, { a: 'one', b: 'two' }) // false
```

Updating this check to compare the list of normalised values instead. The trade-off here is that assumes all object keys can be stringified. For our use case this is a reasonable trade-off as we only use this check for comparing nested text fields like addresses and names.

### DIFF
--- a/controllers/apply/lib/joi-extensions/compare-object.js
+++ b/controllers/apply/lib/joi-extensions/compare-object.js
@@ -3,6 +3,18 @@ const isEqual = require('lodash/isEqual');
 const isObject = require('lodash/isObject');
 const sortBy = require('lodash/sortBy');
 
+/**
+ * Convert object values to an array of normalised strings
+ * Assumes a constraint that object only contains strings.
+ */
+function normaliseValues(sourceObject) {
+    return sortBy(
+        Object.values(sourceObject).map(function (value) {
+            return value.toString().toLowerCase().trim();
+        })
+    );
+}
+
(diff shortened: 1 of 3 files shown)


davidrapson:
[Add validation to prevent shared DOB and name between Main and Senior contacts](https://trello.com/c/1CljVYfO/1118-add-validation-to-prevent-shared-dob-and-name-between-main-and-senior-contacts)

controllers/apply/lib/joi-extensions/compare-object.js
```diff
@@ -15,18 +17,39 @@ module.exports = function (joi) {
...
+                     * @returns {string}
+                     */
+                    function normalise(value) {
+                        return value.toString().toLowerCase().trim();
+                    }
+
+                    if (isObject(value) && isObject(referenceValue)) {
+                        if (
+                            isEqual(
+                                sortBy(Object.values(value).map(normalise)),
```
mattandrews, line 38: Could make this (`Object.values(value).map(normalise)`) a function?

PR #3213, opened 2020-04-20
Remove old fieldAddress in favour of AddressField

davidrapson:
Removes redundant `fieldAddress` code in favour of `AddressField` field type.


davidrapson:
Hopefully aiming to do a few related pull requests in this area to bring this form in line with the standard one.

PR #3214, opened 2020-04-20
Allow field type schema property to be a function

davidrapson:
This is a bit experimental but this PR allows extending rather than completely overriding a field types schema. The thinking here is that it allows fields to _append_ to the field schema without having to duplicate the original schema, hopefully reducing a vector for bugs. Supporting this in addition to overriding the schema for now.

In this example case it allows us to remove the need for the full-name extension and defer to the default schema of the field type.


davidrapson:
Just cleans up the common use case. You can still override it completely, either by:

```js
{ schema: totallyNewSchema }
```

or if using the new function version you could return a completely different schema:

```js
schema(_notUsingTheOriginalSchema) {
    return joi.somethingCustom().required().
}
```

PR #3215, opened 2020-04-20
Organisation finances cleanup

davidrapson:
- Extract organisation finances fields from the mega `fields.js` file
- Convert totalIncomeYear to CurrencyField type
- Remove minTimeAgo rule. The `minTimeAgo` validation rule never actually _validated_ anything, but instead mutated data. We are better making this check a shared function instead.


controllers/apply/under10k/fields/organisation-finances.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+const get = require('lodash/fp/get');
+const { oneLine } = require('common-tags');
+
+const Joi = require('../../lib/joi-extensions');
+const { CurrencyField } = require('../../lib/field-types');
+const isNewOrganisation = require('../lib/new-organisation');
+
+module.exports = {
```
davidrapson, line 9: Preferring this pattern of grouping fields by section rather than one file per field which is a little overkill.

davidrapson:
Just kicked the build off again, got stuck as cancelled.

davidrapson:
Going to revisit this, getting a weird error.

PR #3222, opened 2020-04-21
Add "next" versions of all joi extensions

davidrapson:
This is preparatory work for migrating everything over to joi v17.x. Adds a new `joi-extensions-next` directory with migrated versions of every extension. Had to add a few missing unit tests to give confidence during the migration. But otherwise this was far easier than when we tried it last year.

This should largely be safe to merge. Upgrading from v16 to v17 contains some small changes and is used within our user routes so needs a bit of testing but all the new extensions are unused outside of the tests so can me merged in ahead of time.


package.json
```diff
@@ -84,7 +84,7 @@
   },
   "dependencies": {
     "@hapi/joi": "15.1.1",
-    "@hapi/joi16": "npm:@hapi/joi@16.1.1",
+    "@hapi/joiNext": "npm:@hapi/joi@17.1.1",
```
davidrapson, line 87: Renamed this `joiNext` as having `joi16` as an alias for joi 17 gave me a headache.

controllers/apply/lib/validate-model.js
```diff
@@ -1,128 +0,0 @@
...
-        'date',
-        'date-range',
-        'day-month',
-    ];
-
-    // Check that we haven't used a multipart field on a step with array values
-    // because we use a custom body parser (Formidable) which doesn't handle these
-    // in the same way as body-parser (eg. using qs library). Throws an error if
-    // we attempt to add an array field to one of these steps.
-    formModel.sections = formModel.sections.map((section) => {
```
davidrapson, line 109: Might come back and re-add this to `step-model.js` as this is the one check that is genuinely useful here.

davidrapson:
[Tech debt: use the same version of validation library across the codebase.](https://trello.com/c/nDaDEcgr/1124-tech-debt-use-the-same-version-of-validation-library-across-the-codebase)

PR #3223, opened 2020-04-21
Experimental: Migrate form schemas to latest version of joi

davidrapson:
On the basis that we can merge https://github.com/biglotteryfund/blf-alpha/pull/3222 this PR does a quick experiment to confirm what's involved in migrating the form schemas to the latest version of `joi`

Definitely ⚠️experimental ⚠️. Mostly just to get a sense of the changes involved, not expecting to merge this.


controllers/apply/lib/field-types/checkbox.js
```diff
@@ -29,7 +29,7 @@ class CheckboxField extends Field {
     defaultSchema() {
         const options = this.options || [];
         const baseSchema = Joi.array()
-            .items(Joi.string().valid(options.map((option) => option.value)))
+            .items(Joi.string().valid(...options.map((option) => option.value)))
```
davidrapson, line 32: This is the most common change we need to make. `.valid` now accepts a list of values, rather than an array.

controllers/apply/lib/field-types/checkbox.test.js
```diff
@@ -20,7 +20,7 @@ test('valid field', function () {
 
     field.withValue(['option-1', 'option-2']);
     expect(field.displayValue).toBe('Option 1,\nOption 2');
-    expect(field.validate().error).toBeNull();
+    expect(field.validate().error).toBeUndefined();
```
davidrapson, line 23: Noisiest non-schema change, the `error` property is now `undefined` instead of `null` if the schema is valid.

controllers/apply/standard-proposal/fields.js
```diff
@@ -187,21 +189,21 @@ module.exports = function fieldsFor({ locale, data = {}, flags = {} }) {
 
         function schema() {
             const isEnglandSelected = Joi.array().items(
-                Joi.string().only('england').required(),
```
davidrapson, line 190: `.only` was previously an alias of `.valid` but that's changed in the latest version. Using `.valid` directly here.

controllers/apply/under10k/fields.js
```diff
@@ -1466,9 +1466,7 @@ module.exports = function fieldsFor({ locale, data = {} }) {
                 en: 'This person has to live in the UK.',
                 cy: 'Rhaid i’r person hwn fyw ym Mhrydain',
             }),
-            schema: Joi.fullName()
-                .compare(Joi.ref('mainContactName'))
```
davidrapson, line 1470: This constraint already applied for plain references but not seems to apply to references in extensions too. Looks like we can only have the error on one of the two fields. Annoying but not the end of the world.

controllers/apply/under10k/fields/charity-number.js
```diff
@@ -18,15 +18,15 @@ module.exports = function (locale, data) {
     const excludeRegex = /^[^Oo]+$/;
 
     const baseSchema = Joi.string()
-        .regex(excludeRegex)
+        .pattern(excludeRegex)
```
davidrapson, line 21: `.regex` as renamed to `.pattern`

controllers/apply/under10k/form.test.js
```diff
@@ -687,20 +687,18 @@ test('finance details required if organisation is over 15 months old', function
 
     const validData = mockResponse({
         organisationStartDate: { month: now.month() + 1, year: now.year() },
-        accountingYearDate: null,
```
davidrapson, line 690: This is the one change I'm not sure why we needed to make it. Actual form looked OK, so think this is more a difference in null vs nonexistent properties when we are testing the form like this. Need to double check this one.

docs/__tests__/schema.test.js
```diff
@@ -10,7 +10,7 @@ function validateSchemaDocumentation(formBuilder, schemaDocPath) {
     const docContents = fs.readFileSync(schemaDocPath, 'utf8');
 
     describe(`${form.title} schema documentation`, function () {
-        test.each(Object.keys(form.schema.describe().children))(
+        test.each(Object.keys(form.schema.describe().keys))(
```
davidrapson, line 13: `schema.describe` has changed quite a bit. For our purpose this is the only change for now, but in theory means we could do more with it.

davidrapson:
So…turns out this isn't that many changes once you account for the extensions. Good to know.

controllers/apply/lib/joi-extensions/word-count.js
```diff
@@ -3,49 +3,61 @@ const countWords = require('../count-words');
...
     return {
+        type: 'string',
         base: joi.string(),
-        name: 'string',
-        language: {
-            maxWords: 'must have less than {{max}} words',
-            minWords: 'must have at least {{min}} words',
+        messages: {
+            'string.minWords': 'must have at least {{#min}} words',
+            'string.maxWords': 'must have no more than {{#max}} words',
```
mattandrews, line 10: Why the wording change?

davidrapson: The previous wording was incorrect, it said "less than" when it's really "up to". This is just an internal message but wanted to make sure it was accurate.

davidrapson:
[Tech debt: use the same version of validation library across the codebase.](https://trello.com/c/nDaDEcgr/1124-tech-debt-use-the-same-version-of-validation-library-across-the-codebase)

PR #3224, opened 2020-04-21
Remove minTimeAgo validation rule

davidrapson:
Re-do of https://github.com/biglotteryfund/blf-alpha/pull/3215 with _just_ the validation rule changes.

The `minTimeAgo` validation rule never actually _validated_ anything, but instead mutated data. We are better making this check a shared function instead.

### DIFF
--- a/controllers/apply/under10k/__snapshots__/form.test.js.snap
+++ b/controllers/apply/under10k/__snapshots__/form.test.js.snap
@@ -20,6 +20,8 @@ Array [
   "organisationStartDate: Enter a month and year",
   "organisationAddress: Enter a full UK address",
   "organisationType: Select a type of organisation",
+  "accountingYearDate: Enter a day and month",
+  "totalIncomeYear: Enter a total income for the year (eg. a whole number with no commas or decimal points)",
   "mainContactName: Enter first and last name",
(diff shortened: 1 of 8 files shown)


controllers/apply/under10k/__snapshots__/form.test.js.snap
```diff
@@ -20,6 +20,8 @@ Array [
   "organisationStartDate: Enter a month and year",
   "organisationAddress: Enter a full UK address",
   "organisationType: Select a type of organisation",
+  "accountingYearDate: Enter a day and month",
```
davidrapson, line 23: This is the result of moving the logic up a level. It essentially means that until until you answer the `organisationStartDate` question we assume you are going to need to answer these questions. This is arguably more intentional behaviour. Affects the summary screen:

**Before**

image

**After**

image

This doesn't change the end result. But a judgement call here, new behaviour is to allow you to fill in these fields and strip them only once you've told us the date, vs. assuming they are not needed to begin with. Feels more honest as the new organisation rule is the exception rather than the default.

If we want to retain the old behaviour we should update the `isNewOrganisation` check to return `true` if a start date hasn't been provided yet. It defaults to `false` right now.

PR #3225, opened 2020-04-22
Launch new standard location questions

davidrapson:
- Makes `enableNewLocationQuestions` default to `true`. We can come back and clean up the feature flag code afterwards
- Removes `-beta` suffix from the schema version


mattandrews:
This looks good to me 👍

PR #3236, opened 2020-04-23
Add last updated support to flexible content

davidrapson:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/262

Adds support for a last updated message to flexible content pages.  Main motivation is the new insights pages https://www.tnlcommunityfund.org.uk/insights/covid-19-resources/responding-to-covid-19/covid-19-the-immediate-issues which have manual last updated dates which is error prone and already incorrect on those pages.

Opted to just show the day for now, not sure we need to show exact times.

image

Required a little bit of reworking of the flexible content macro so that we pass the full entry in rather than just the flexible content part. Gives us access to the last updated date.

### DIFF
--- a/views/components/flexible-content-next/macro.njk
+++ b/views/components/flexible-content-next/macro.njk
@@ -63,10 +63,18 @@
                         ) }}
                     {% elseif part.type === 'factRiver' %}
                         {{ factRiver(part.content) }}
-                    {% elseif part.type === 'tableOfContents' and flexibleContent %}
+                    {% elseif part.type === 'tableOfContents' %}
                         <div class="s-prose u-constrained-content-wide">
                             {{ part.content | safe }}
                         </div>
+
+                        {% if part.lastUpdated %}
+                            <div class="message message--minor">
+                                <strong>{{ __('global.misc.lastUpdated') }}</strong>:
+                                {{ formatDate(part.lastUpdated.date, "dddd D MMMM, YYYY") }}
+                            </div>
+                        {% endif %}
+
                         {{ tableOfContents(flexibleContent, anchor = flexAnchor) }}
                     {% else %}
                         <div class="s-prose u-constrained-content-wide">
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -158,6 +158,7 @@ global:
     day: Diwrnod
     month: Mis
     year: Blwyddyn
+    lastUpdated: Wedi’i ddiweddaru diwethaf
   programmes:
     programmeLabels:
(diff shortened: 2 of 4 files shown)


controllers/common/index.js
```diff
@@ -92,7 +95,9 @@ function renderFlexibleContentChild(req, res, entry) {
     res.render(path.resolve(__dirname, './views/flexible-content-page'), {
         breadcrumbs: res.locals.breadcrumbs.concat({ label: res.locals.title }),
         // @TODO: Can we rename API responses to consistently name this flexibleContent?
-        content: { flexibleContent: entry.content },
+        content: Object.assign(entry, {
+            flexibleContent: entry.content,
```
davidrapson, line 99: This is a little awkward. For some reason the API uses `flexibleContent` for the main flexible content pages and `content` for the occasional child pages that support flexible content. We should come back and update the API to use the same name but otherwise this works.

controllers/common/index.js
```diff
@@ -92,7 +95,9 @@ function renderFlexibleContentChild(req, res, entry) {
     res.render(path.resolve(__dirname, './views/flexible-content-page'), {
         breadcrumbs: res.locals.breadcrumbs.concat({ label: res.locals.title }),
         // @TODO: Can we rename API responses to consistently name this flexibleContent?
-        content: { flexibleContent: entry.content },
+        entry: Object.assign(entry, {
```
davidrapson, line 98: This is a little awkward. For some reason the API uses `flexibleContent` for the main flexible content pages and `content` for the occasional child pages that support flexible content. We should come back and update the API to use the same name but otherwise this works.

mattandrews:
Would it be simpler to achieve this if the API returned `showLastUpdatedDate` as a date rather than a boolean? Then if it's not null, we can just display its value without having to pass around entries etc?

views/components/flexible-content-next/macro.njk
```diff
@@ -63,10 +63,18 @@
                         ) }}
                     {% elseif part.type === 'factRiver' %}
                         {{ factRiver(part.content) }}
-                    {% elseif part.type === 'tableOfContents' and flexibleContent %}
+                    {% elseif part.type === 'tableOfContents' %}
```
davidrapson, line 66: Remove this because we are inside a loop for `flexibleContent` so this is always true.

davidrapson:
@mattandrews Good shout! Updated both pull requests to reflect that change

mattandrews:
Can we get this (and the CMS change) out today?

PR #3253, opened 2020-04-28
Update Your Funding Proposal form to reflect that England will no longer be accepting proposals for non COVID-19 projects

davidrapson:
## Before you start page

start-after

Main change here is to **remove the process map**. This is currently inaccurate and too variable so the best thing we can do for now is to remove it.

## Confirmation page

Currently we have the following for "What happens next" on the confirmation page:

confirmation-other-countries

This text is inaccurate for England. We'll be keeping the existing text for all other countries but varying this if in England. New confirmation text is as follows:

confirmation-england

Added snapshot tests for the confirmation text, like we do for simple, now this is variable. Also updated our integration tests to submit a separate test proposal for both England and Northern Ireland.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1111,7 +1111,7 @@ yourFundingProposal:
       title: Before you start
       items:
         - Have a look at our website for <a href="/funding/over10k">information about funding over £10,000</a>
-        - If you'd rather chat to us on the phone about your idea—<a href="/contact">talk to us</a>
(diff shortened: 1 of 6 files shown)


davidrapson:
[Update Your Funding Proposal form to reflect that England will no longer be accepting proposals non COVID-19 projects](https://trello.com/c/Gxfg6Mu7/708-update-your-funding-proposal-form-to-reflect-that-england-will-no-longer-be-accepting-proposals-non-covid-19-projects)

PR #3260, opened 2020-04-29
Standard regions validation fix

davidrapson:
Spotted an issue with the validation rule for standard regions where it was possible to bypass the validation rules if you submitted an empty answer.

- We had a unit test for this but it was testing with `null` rather than true _absence_ of the field so was false positive. Updated this to completely omit the field
- Also added a missing test to validate that the region answers were stripped outside of England. This was working fine but need a test at our backs to help verify this fix
- Added an additional Cypress test to guard against this validation rule

### DIFF
--- a/controllers/apply/standard-proposal/fields.js
+++ b/controllers/apply/standard-proposal/fields.js
@@ -207,7 +207,10 @@ module.exports = function fieldsFor({ locale, data = {} }) {
 
             return Joi.when(Joi.ref('projectCountries'), {
                 is: isEnglandSelected,
-                then: [validAllEngland, validRegionOptions],
+                then: Joi.alternatives([
+                    validAllEngland,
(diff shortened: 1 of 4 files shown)


controllers/apply/standard-proposal/fields.js
```diff
@@ -207,7 +207,10 @@ module.exports = function fieldsFor({ locale, data = {} }) {
 
             return Joi.when(Joi.ref('projectCountries'), {
                 is: isEnglandSelected,
-                then: [validAllEngland, validRegionOptions],
+                then: Joi.alternatives([
```
davidrapson, line 210: This is the resulting fix. The `[]` syntax maps to `Joi.alternatives`. Needed an explicit `.required()` otherwise the absence of the field is considered OK.

PR #3270, opened 2020-04-29
Add flex sidebar to child pages

mattandrews:
This enables child page grids/lists to show a sidebar of arbitrary HTML content, defined in the controller. 

image

At the moment the intention is for this sidebar to show alongside a grid of child pages, so that's where we render it, but it also works for text lists:

image

I couldn't think of a more CMS-y way of achieving this ("sidebar" block in FC?) but given this is hardcoded to a specific page I think we're okay for now.

In future it's probable that this will need some filtering, eg. only showing specifically-tagged blogposts etc, but for now Engagement are happy with it just showing all of them.

### DIFF
--- a/controllers/insights/index.js
+++ b/controllers/insights/index.js
@@ -1,5 +1,7 @@
 'use strict';
 const express = require('express');
+const nunjucks = require('nunjucks');
+const Sentry = require('@sentry/node');
 const path = require('path');
 const clone = require('lodash/clone');
 const compact = require('lodash/compact');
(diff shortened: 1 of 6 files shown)


controllers/insights/index.js
```diff
@@ -78,7 +80,34 @@ router.get('/documents/:slug?', async function (req, res, next) {
...
+
+        const context = {
+            ...res.locals,
+            ...req.app.locals,
+            ...{
+               blogposts: blogposts.result
+            }
+        };
+
+        nunjucks.render(
```
davidrapson, line 98: This is pretty gnarly. Not worth blocking on this but would be good to come back to this if we can and see if there's a way of extending off `flexible-content-page` and capturing a block with the sidebar content, rather than having to call out to nunjucks directly.

controllers/insights/views/covid-sidebar.njk
```diff
@@ -0,0 +1,25 @@
+{% from "components/blog/macro.njk" import blogTrail with context %}
+
+<div class="u-tone-background-tint u-padded u-margin-bottom">
+    <h4>Latest blogposts</h4>
```
davidrapson, line 4: Can we put this in the locale file, assuming this needs to be translated?

PR #3279, opened 2020-04-30
Reinstate ebulletin signup

mattandrews:
This isn't ready to launch yet but we could likely merge this for test environments.

Say hello to the new provider... same as the old provider.

I was able to reinstate most of the deleted code for the old DotMailer API integration. The main changes were migrating the code to use Joi (v16 because I needed `error` to be `null` rather than `undefined`) and `got`. 

Template is the same as before, eg:

image

There's still discussion within the brand team about which fields to capture so for now we only store the email, firstname and lastname fields. If/when this is expanded at the DotDigital end I'll update this, but this is a working example which puts data into the API.

Not sure what the test strategy should be here – probably worth holding off until we know what the final schema looks like?


config/locales/cy.yml
```diff
@@ -251,6 +251,40 @@ toplevel:
...
+      firstName: Enw cyntaf
+      lastName: Cyfenw
+      organisation: Mudiad
+      emailAddress: Cyfeiriad e-bost
+    errors:
+      firstName: Darparwch y maes enw cyntaf
+      lastName: Darparwch y maes cyfenw
+      emailMissing: Darparwch y maes cyfeiriad e-bost
+      emailInvalid: Rhowch gyfeiriad e-bost dilys
+      location: Please choose a country @TODO
```
mattandrews, line 275: Have asked for translation for this, it was missing in the original version!

common/ebulletin.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
```
davidrapson, line 1: This should probably live under `controllers/ebulletin/lib/newsletter-service` as we only use it in that router right now

controllers/ebulletin/index.js
```diff
@@ -0,0 +1,72 @@
...
+const { noStore, csrfProtection } = require('../../common/cached');
+const { sanitiseRequestBody } = require('../../common/sanitise');
+const validateSchema = require('../../common/validate-schema');
+const newsletterService = require('../../common/ebulletin');
+const logger = require('../../common/logger').child({ service: 'ebulletin' });
+
+const router = express.Router();
+
+function subscribe(formData) {
+    const subscriptionData = {
```
davidrapson, line 18: Rather than constructing this here, should this live in the newsletter service? Perhaps we just pass the `email` and `dataFields` at this point?

controllers/ebulletin/lib/contact-schema.js
```diff
@@ -0,0 +1,84 @@
...
+        const locations = buildValidLocations(i18n);
+
+        const email = Joi.string().email().required();
+        const firstName = Joi.string().required();
+        const lastName = Joi.string().required();
+        const location = Joi.string().valid(...locations.map(_ => _.value)).required();
+        const organisation = Joi.string().allow('').optional();
+
+        return {
+            schema: Joi.object({
```
davidrapson, line 45: Mind adding a couple of small unit tests for the schema? Like we do for account schemas. This isn't complex validation logic so just a test or two.

mattandrews:
Can you give this another look @davidrapson? Have updated it!

PR #3291, opened 2020-05-01
Add automaticContentList Flexible Content component

mattandrews:
Pairs with https://github.com/biglotteryfund/craft-dev/pull/265 and allows adding automatically-fetched content to pages:

image

Uses existing patterns/macros to display the three supported types of content (blogs, press releases and funding programmes). 

Also fixed a related bug on pages [like this](https://www.tnlcommunityfund.org.uk/about/the-national-lotterys-25th-birthday) which use the old FC macros and didn't have the title logic.
image

### DIFF
--- a/controllers/updates/views/listing/press-releases.njk
+++ b/controllers/updates/views/listing/press-releases.njk
@@ -29,7 +29,7 @@
                             {% for entry in entries %}
                                 <li class="article-listing__item">
                                     {% set subtitle %}
-                                        <time class="article-teaser__pubdate" datetime="{{ props.postDate.date }}">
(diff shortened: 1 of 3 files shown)


controllers/updates/views/listing/press-releases.njk
```diff
@@ -29,7 +29,7 @@
                             {% for entry in entries %}
                                 <li class="article-listing__item">
                                     {% set subtitle %}
-                                        <time class="article-teaser__pubdate" datetime="{{ props.postDate.date }}">
+                                        <time class="article-teaser__pubdate" datetime="{{ entry.postDate.date }}">
```
davidrapson, line 32: nice catch!

PR #3292, opened 2020-05-01
Run prettier against ebulletin code

davidrapson:
Run prettier against ebulletin code. My editor did this automatically so just committing.


mattandrews:
Ahh thanks, I think my newly-formatted Mac doesn't have this set up properly. Will fix!

davidrapson:
Mine is _overkeen_ if anything, even if I revert the changes it will just reformat again later. I think I prefer your version!

PR #3310, opened 2020-05-06
Add Covid-19 funding info

mattandrews:
This adds two things (and pairs with [these CMS changes](https://github.com/biglotteryfund/craft-dev/pull/268)):

- A new paragraph showing a specific country's funding status for Covid-19:

image
image
image

- A new "trail text" field for funding programmes which replaces their (longer) intro/description text as it used to be:

image

(this will need to be populated in the CMS)

### DIFF
--- a/common/content-api.js
+++ b/common/content-api.js
@@ -214,7 +214,7 @@ function getFundingProgrammes({
     ]).then((responses) => {
         const [enResults, cyResults] = responses.map(mapAttrs);
         return {
-            meta: head(responses).meta,
+            meta: locale === 'en' ? head(responses).meta : responses[1].meta,
             result: mergeWelshBy('slug')(locale, enResults, cyResults),
         };
     });
--- a/common/content-api.js
+++ b/common/content-api.js
@@ -4,7 +4,6 @@ const find = require('lodash/fp/find');
 const flatten = require('lodash/flatten');
 const get = require('lodash/fp/get');
 const getOr = require('lodash/fp/getOr');
-const head = require('lodash/fp/head');
 const map = require('lodash/fp/map');
 const pick = require('lodash/pick');
 const sortBy = require('lodash/fp/sortBy');
--- a/common/content-api.js
+++ b/common/content-api.js
@@ -212,10 +212,14 @@ function getFundingProgrammes({
         queryContentApi.get('v2/en/funding-programmes', requestOptions).json(),
         queryContentApi.get('v2/cy/funding-programmes', requestOptions).json(),
     ]).then((responses) => {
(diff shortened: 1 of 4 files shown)


common/content-api.js
```diff
@@ -214,7 +214,7 @@ function getFundingProgrammes({
     ]).then((responses) => {
         const [enResults, cyResults] = responses.map(mapAttrs);
         return {
-            meta: head(responses).meta,
+            meta: locale === 'en' ? head(responses).meta : responses[1].meta,
             result: mergeWelshBy('slug')(locale, enResults, cyResults),
         };
```
davidrapson, line 219: Wondering if we could rewrite this block to make what `locale === 'en' ? head(responses).meta : responses[1].meta` is doing clearer.  Maybe something like:

```js
const [enResponse, cyResponse] = responses;
return {
    meta: locale === 'en' ? enResponse.meta : cyResponse.meta,
    result: mergeWelshBy('slug')(
        locale,
        enResponse.map(mapAttrs),
        cyResponse.map(mapAttrs)
    ),
};
```

controllers/funding/programmes/index.js
```diff
@@ -48,6 +48,21 @@ router.get('/', injectHeroImage('rosemount-1-letterbox-new'), async function (
...
+            countryCovidStatus = covidStatuses.find(
+                (status) => status.country === locationParam
+            );
+            // Fall back to the UK-wide statement
+            if (!countryCovidStatus) {
+                countryCovidStatus = covidStatuses.find(
+                    (status) => status.country === 'ukWide'
+                );
+            }
+            res.locals.countryCovidStatus = countryCovidStatus;
```
davidrapson, line 63: You might be able to combine this logic like this:

```js
res.locals.countryCovidStatus = covidStatuses.find(function (status) {
    return (
        status.country === locationParam ||
        status.country === 'ukWide'
    );
});
```
Or, possibly:

```js
res.locals.countryCovidStatus = covidStatuses.find(function (
    status
) {
    if (status.country === locationParam) {
        return true;
    } else {
        return status.country === 'ukWide';
    }
});
```

Might not be following this logic though so ignore this if I've oversimplified things

common/content-api.js
mattandrews: Good call – needed to be `mapAttrs(enResponse)` but much clearer 👍

(FYI the reason for this change was to allow the `meta` field to return the current locale's version, which is now translateable for the global)

controllers/funding/programmes/index.js
mattandrews: No this is a good shout, used the first approach 👍

PR #3312, opened 2020-05-06
Schema update for the project date field

mo-durvesh:
Validation on if a disabled radio field has been selected.

### DIFF
--- a/controllers/apply/under10k/fields/project-dates-next.js
+++ b/controllers/apply/under10k/fields/project-dates-next.js
@@ -1,5 +1,6 @@
 'use strict';
 const get = require('lodash/fp/get');
+const has = require('lodash/fp/has');
 const moment = require('moment');
 const { oneLine } = require('common-tags');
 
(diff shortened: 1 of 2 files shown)


mo-durvesh:
We may want to make this apart of the default schema for the radio field.

controllers/apply/under10k/fields/project-dates-next.js
```diff
@@ -58,6 +59,9 @@ module.exports = {
                 cy: `Pryd hoffech chi gael yr arian os ydych chi'n cael eich dyfarnu?`,
             }),
             options: options(),
+            schema:  Joi.string().valid(
+                options().filter((option) => !has('disabled')(option.attributes)).map((option) => option.value)
```
mattandrews, line 63: There's quite a lot going on in one line here, maybe we could make these named functions to make it clearer what's happening? Something like:

```
    get schema() {
        const enabledOptions = (option) => !has('disabled')(option.attributes);
        const optionValues = (option) => option.value;
        return options().filter(enabledOptions).map(optionValues);
    }
```

mattandrews:
There's a [failing test](https://travis-ci.com/github/biglotteryfund/blf-alpha/builds/164235395#L477) here too!

davidrapson:
Just pushed a fix for the tests, couple of our test cases were asserting the wrong thing.

PR #3320, opened 2020-05-07
Show notice for pending England applications

davidrapson:
Tells people with pending applications for under £10,000 in England that our priorities are different and their application might not be suitable any more.

image

image

### DIFF
--- a/controllers/apply/dashboard/index.js
+++ b/controllers/apply/dashboard/index.js
@@ -11,9 +11,11 @@ const {
     PendingApplication,
     SubmittedApplication,
 } = require('../../../db/models');
+const logger = require('../../../common/logger').child({ service: 'apply' });
 
 const enrichAwardsForAll = require('../under10k/enrich');
 const enrichStandard = require('../standard-proposal/enrich');
+const getNotice = require('./lib/get-notice');
 
 const router = express.Router();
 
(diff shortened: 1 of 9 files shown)


davidrapson:
Needs Welsh translation

controllers/apply/dashboard/index.js
```diff
@@ -106,9 +108,15 @@ router.get(
                 }),
             ]);
 
+            const notice = getNotice(req.i18n.getLocale(), pendingSimple);
+            if (notice) {
+                logger.info('Notice shown on dashboard');
```
davidrapson, line 113: Added logs here as it's useful to see how many people see this notice.

controllers/apply/dashboard/lib/get-notice.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+const config = require('config');
+const get = require('lodash/fp/get');
+const { oneLine } = require('common-tags');
+
+module.exports = function getNotice(locale, pendingSimple = []) {
```
mattandrews, line 6: The `pendingSimple` name here is a bit confusing, could we call it `pendingApplications` or something to make it clearer what kind of array it's operating over? (I know it's Simple apps as opposed to Standard but it's not clear it's _applications_ from the name).

davidrapson: Good shout, forgot to rename this.

davidrapson:
Added welsh translation. Need to double check if this is OK to go out earlier or if it's for the 13th, in which case I'll use a different feature flag.

PR #3331, opened 2020-05-11
Refactor ebulletin logic to allow stakeholder signups [WIP]

mattandrews:
Adds a second ebulletin signup form with additional fields:

image

This isn't finished yet – needs translation and confirmation that the fields are right (not sure if "Organisation" will be there too) but the approach is ready for review.

### DIFF
--- a/controllers/ebulletin/index.js
+++ b/controllers/ebulletin/index.js
@@ -32,7 +32,7 @@ function renderForm(req, res, data = null, errors = []) {
 }
 
 router
-    .route('/:contactType(stakeholder)?')
+    .route('/:contactType(policy)?')
     .all(noStore, csrfProtection)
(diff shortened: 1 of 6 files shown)


controllers/ebulletin/index.js
```diff
@@ -17,30 +22,38 @@ const router = express.Router();
...
         validLocations: buildValidLocations(req.i18n),
+        validSectors: buildValidSectors(req.i18n),
         formValues: data,
         errors: errors,
     });
 }
 
 router
-    .route('/')
+    .route('/:contactType(stakeholder)?')
```
davidrapson, line 35: Might suggest just splitting this into two separate route handlers, might be simpler compared to conditional params. But this works fine for now, so don't let this block you.

PR #3353, opened 2020-05-14
Additional notice about eligibility in England

davidrapson:
Change is to add the second line to mention changed organisation type eligibility. We already have a stronger message in place on the summary screen if you have an actual application in flight for a statutory groups. This is to cover things more generally for anyone with a pending application, who may not have selected organisation type yet.

image

### DIFF
--- a/controllers/apply/lib/notices.js
+++ b/controllers/apply/lib/notices.js
@@ -38,7 +38,7 @@ module.exports = {
                     en: oneLine`If you've started an application already,
                     and it's not related to supporting your community
                     or organisation through the pandemic, we won't be
-                    able to accept it. But you could decide to start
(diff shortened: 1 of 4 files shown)


controllers/apply/lib/notices.js
```diff
@@ -47,6 +47,15 @@ module.exports = {
                     un newydd sy'n canolbwyntio ar COVID-19 yn lle.`,
                 }),
             });
+
+            if (enableGovCOVIDUpdates) {
+                notices.push({
+                    title: oneLine`We've also changed our eligibility
```
davidrapson, line 53: No translation for this, it's only shown in England. Took the same approach as the England only terms for now.

PR #3354, opened 2020-05-14
Revised priorities

davidrapson:
Before:

Screenshot 2020-05-14 at 10 59 08

After:

Screenshot 2020-05-14 at 10 58 49

### DIFF
--- a/controllers/apply/under10k/fields/your-idea.js
+++ b/controllers/apply/under10k/fields/your-idea.js
@@ -90,7 +90,7 @@ module.exports = {
             ],
         });
     },
-    fieldYourIdeaPriorities(locale, data) {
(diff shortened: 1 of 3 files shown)


controllers/apply/under10k/fields/your-idea.js
```diff
@@ -99,9 +99,39 @@ module.exports = {
         const minWords = 50;
         const maxWords = 150;
 
-        function guidanceText() {
-            const prioritiesCOVID19 = localise({
-                en: `<ol>
+        function prioritiesCOVID19() {
+            if (flags.enableGovCOVIDUpdates) {
```
davidrapson, line 103: Behind a flag as this priorities don't come into effect until we launch new funding

PR #3356, opened 2020-05-14
Launch new form changes

davidrapson:
This launches the following changes:

- Revised priorities text
- Removes statutory groups as options in England along with notices for people with pending applications from these groups
- Enables additional Ts&Cs in England

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -30,7 +30,7 @@
     "expiryPeriod": [3, "months"]
   },
   "fundingUnder10k": {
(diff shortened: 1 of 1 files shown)


mattandrews:
This is fine but I think we're still awaiting final links for the new T&Cs before this can go out?

davidrapson:
Yep, that's correct, have updated the description to reflect that  👍

davidrapson:
Merging to test site ahead of tomorrow morning.

PR #3360, opened 2020-05-14
Under £10,000 terms updates

davidrapson:
Cleaned up the field definitions but critically updates links in new terms:

- [x] State aid guidance
- [ ] New privacy policy


davidrapson:
Yeah, still waiting on that. Going to merge this now so the revised terms wording is out on the test site (still behind a feature flag) and the add the privacy link when we have it.

Issue #3351, opened 2020-05-13
Convert remaining fields to use field types


davidrapson:
The following fields in the under10k form need converting to field types (they currently use plain objects):

- [x] projectBudget
- [x] beneficiariesGroupsCheck
- [x] beneficiariesGroups
- [x] beneficiariesGroupsOther
- [x] beneficiariesGroupsEthnicBackground
- [x] beneficiariesGroupsGender
- [x] beneficiariesGroupsAge
- [x] beneficiariesGroupsDisabledPeople
- [x] beneficiariesGroupsReligion
- [x] beneficiariesGroupsReligionOther
- [x] beneficiariesWelshLanguage
- [x] beneficiariesNorthernIrelandCommunity
- [x] organisationLegalName
- [x] organisationStartDate
- [x] companyNumber
- [x] charityNumber
- [x] educationNumber
- [x] accountingYearDate
- [x] mainContactAddressHistory
- [x] mainContactLanguagePreference
- [x] mainContactCommunicationNeeds
- [x] seniorContactAddressHistory
- [x] seniorContactLanguagePreference
- [x] seniorContactCommunicationNeeds
- [x] bankAccountName
- [x] buildingSocietyNumber

In this list there are four types of fields that need new field type classes creating:

- [x] budget
- [x] month-year
- [x] day-month
- [x] address-history

Once these fields are using field types you can delete some supporting code, notably:

- [x] `this._isClass` in the base `field.js` definition
- [x] Can rename `enrichFieldClass` to `enrichField` and delete the old non-class method from `form-mode`
- [x] Remove most, if not all, the code in `formatters.js`. This has been largely replaced by a `displayValue` method on the field types themselves

davidrapson:
Raised https://github.com/biglotteryfund/blf-alpha/pull/3375 as a reference pull request as it covers the two most common tasks here: a) creating new field types where they are missing and b) updating existing plain-object fields to use a field type.

PR #3375, opened 2020-05-18
Field types finances

davidrapson:
Part of https://github.com/biglotteryfund/blf-alpha/issues/3351

Raising a pull request as a useful reference example for @marksmithtnlcf, @mo-durvesh and @joshgriff465 

- Groups organisation finance fields together
- Add a new `DayMonthField` type with tests
- Update `accountingYearDate` to use `DayMonthField`


controllers/apply/lib/field-types/day-month.js
```diff
@@ -0,0 +1,50 @@
...
+
+    defaultSchema() {
+        return Joi.dayMonth().required();
+    }
+
+    defaultMessages() {
+        return [
+            {
+                type: 'base',
+                message: this.localise({
```
davidrapson, line 20: Because these message don't reference anything to do with the specific field (e.g. accounting year end) we can included them as default messages

controllers/apply/lib/field-types/day-month.js
```diff
@@ -0,0 +1,50 @@
...
+                    en: 'Enter a real day and month',
+                    cy: 'Rhowch ddiwrnod a mis go iawn',
+                }),
+            },
+        ];
+    }
+
+    get displayValue() {
+        if (this.value) {
+            const dt = moment({
```
davidrapson, line 37: This copies the previous `formatDayMonth` method in `formatters.js`. Eventually that file should be removed.

PR #3376, opened 2020-05-18
Add a separate success screen for ebulletin signups

mattandrews:
image

This is to allow us to use a generic signup success URL within DotDigital, as well as on the site.

Also disables the plain signup form (in favour of the pending "policy" one) for now, until that one is ready to go.

### DIFF
--- a/controllers/ebulletin/index.js
+++ b/controllers/ebulletin/index.js
@@ -33,7 +33,13 @@ function renderForm(req, res, data = null, errors = []) {
 
 router
     .route('/:contactType(policy)?')
-    .all(noStore, csrfProtection)
+    .all(noStore, csrfProtection, (req, res, next) => {
+        // Temporarily disable non-policy signup form which will launch later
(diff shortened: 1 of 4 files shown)


mattandrews:
(pending Welsh text)

controllers/ebulletin/index.js
```diff
@@ -33,7 +33,13 @@ function renderForm(req, res, data = null, errors = []) {
 
 router
     .route('/:contactType(policy)?')
-    .all(noStore, csrfProtection)
+    .all(noStore, csrfProtection, (req, res, next) => {
+        // Temporarily disable non-policy signup form which will launch later
+        if (!req.params.contactType || req.params.contactType !== 'policy') {
```
davidrapson, line 38: Might write this as:

```js
if (!req.params.contactType || req.params.contactType !== 'policy') {
  res.redirect('/');
} else {
  next();
}

without the early return. Doesn't really matter though.

PR #3390, opened 2020-05-19
Pre-fill end date in England when ASAP

davidrapson:
Paired with @joshgriff465 

Pre-fill end date in England when ASAP. This means that we don't ask for _either_ the start or end date in England. End date is pre-filled to be 6 months from submission.

- [x] Added a standalone feature flag
- [x] Update `forSalesforce` to pre-fill end date like we do with the start date
- [x] Update form schema and form model to exclude end date field in England when ASAP.
- [x] Update integration tests
- [x] Add additional guidance text to ASAP check to explain about 6 month spend rule (copy pending)


controllers/apply/under10k/fields/project-dates.js
```diff
@@ -193,29 +194,38 @@ module.exports = {
             const maxDate = moment().add(getMaxDurationMonths(), 'months');
 
             /**
-             * When projectStartDateCheck is asap
-             * we don't show the project start date question
-             * so we allow the end date to be any future date
+             * For projects in England when asap we skip the project dates
```
mattandrews, line 197: Can we change `when asap` to be `when projectStartDateCheck is asap` (like before) so it's clearer what this means? Same for line 200 below.

davidrapson:
@joshgriff465 You mind handling the comment changes Matt mentioned?

PR #3395, opened 2020-05-20
Under £10,000 England specific priorities

davidrapson:
We had made the assumption that for the new priorities text we would be using the same across the board. Had confirmation this new text should only be for England (this existing text has different wording for the first two bullet points outside of England).

### DIFF
--- a/controllers/apply/under10k/fields/your-idea.js
+++ b/controllers/apply/under10k/fields/your-idea.js
@@ -26,6 +26,43 @@ module.exports = {
         const minWords = 50;
         const maxWords = 300;
 
+        function guidanceText() {
+            return localise({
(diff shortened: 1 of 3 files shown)


controllers/apply/under10k/fields/your-idea.js
```diff
@@ -100,7 +99,7 @@ module.exports = {
         const maxWords = 150;
 
         function prioritiesCOVID19() {
-            if (flags.enableGovCOVIDUpdates) {
+            if (flags.enableGovCOVIDUpdates && projectCountry === 'england') {
```
davidrapson, line 102: This the key change adding a `projectCountry === 'england'` check for this version of the priorities.

PR #3405, opened 2020-05-21
Terms and conditions update, 

joshgriff465:
1.8 bullet point added, semi-colon added to second section title, removed repeating 2.2


davidrapson:
Nope, these terms are England only so we are happy with just English for now (there's a separate bilingual set of terms outside England that isn't being updated). Cheers for checking though!

PR #3406, opened 2020-05-22
Add fallback page title to flexible content page

davidrapson:
If a flexible content page didn't have a hero image then no title was shown. Adds the same pageTitle fallback from other templates.


mattandrews:
Looks good, merge this now and we can get it live this morning! 👍

PR #3426, opened 2020-05-29
Add dummy GMS seed script

mattandrews:
This is a utility script for submitting a bunch of GMS applications in one batch. Previously to do this we'd run Cypress tests but this is quite slow as it steps through the form page-by-page. 

This won't run on its own without configuring environment variables properly to point at the relevant Salesforce environment – I used this yesterday to send 200 applications (by request of the GMS team) into the Covid-19 sandbox. Committing it here as it may be useful for the future!

### DIFF
--- a/bin/seed-salesforce
+++ b/bin/seed-salesforce
@@ -1,0 +1,51 @@
+#!/usr/bin/env node
+'use strict';
+require('dotenv').config();
+
+const moment = require('moment');
+const times = require('lodash/times');
+
+const mockAwardsForAll = require('../controllers/apply/under10k/mocks');
+const formBuilder = require('../controllers/apply/under10k/form');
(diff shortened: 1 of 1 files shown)


bin/seed-salesforce
```diff
@@ -0,0 +1,52 @@
...
+const times = require('lodash/times');
+
+const mockAwardsForAll = require('../controllers/apply/under10k/mocks');
+const formBuilder = require('../controllers/apply/under10k/form');
+const salesforceService = require('../controllers/apply/form-router/lib/salesforce');
+
+const submitMockApplication = async (applicationId) => {
+
+    const dummyData = mockAwardsForAll.mockResponse({
+        projectLocation: 'bournemouth'
```
mattandrews, line 15: The Covid-19 GMS sandbox is configured to only process applications in this location, FYI!

PR #3458, opened 2020-06-09
Add new "people" block to Flexible Content

mattandrews:
This adds a new "person" block type to Flexible Content to allow us to migrate the Our People section into flexible content / About page types. Looks like this:

image

There are some changes in the flexible content template to show a sidebar of sibling pages, which for now is manually enabled for a page type in the controller, though we might wish to roll this out to other sections at some stage.

Pairs with https://github.com/biglotteryfund/craft-dev/pull/283


assets/sass/components/_navigation.scss
```diff
@@ -152,11 +152,14 @@
         }
     }
     .split-nav__prev {
+        margin-right: $spacingUnit;
```
mattandrews, line 155: This fixes a display issue where these elements could butt up against one another.

PR #3578, opened 2020-07-27
Send bulk emails

mattandrews:
This change adds a utility script which can be used to send bulk HTML emails, branded like our other messages, to a large number of users via our AWS Simple Email Service.

It requires a bit of configuration to run locally – specifically

- a CSV file of email addresses to send to (which will be converted into a SQLite database)
- a Nunjucks template containing the email message to send (which will be uploaded to SES as a reusable template)

When these things are provided, it will send emails within our sending limits (14/second currently) and report any errors. It will also mark users as having been emailed, so repeat runs of the script won't email the same user again.

Not something we should use very often, but handy to have around in the scenario where we need to contact more people than our Exchange/Outlook systems will allow.


mattandrews:
Here's how the email looks:

image

(still waiting for the real text!)

PR #3607, opened 2020-08-03
Add datepicker to remaining staff tools

mattandrews:
A few of the staff tools didn't have a datepicker (which is helpful for the ones that get used in monthly reporting, like the website satisfaction survey) so I've added it to those ones and made the others a bit more consistent.

Adds this to Surveys and Feedback pages:

image

### DIFF
--- a/controllers/tools/applications/views/applications.njk
+++ b/controllers/tools/applications/views/applications.njk
@@ -46,7 +46,7 @@
             </h1>
 
             <p>
-                <strong>Date range</strong>: {{ formatDate(dateRange.start) }}–{{ formatDate(dateRange.end) }}
+                {{ formatDate(dateRange.start) }}–{{ formatDate(dateRange.end) }}
                 (<a href="?">Reset dates</a>)
             </p>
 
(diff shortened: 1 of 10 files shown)


controllers/tools/applications/views/applications.njk
```diff
@@ -46,7 +46,7 @@
             </h1>
 
             <p>
-                <strong>Date range</strong>: {{ formatDate(dateRange.start) }}–{{ formatDate(dateRange.end) }}
+                {{ formatDate(dateRange.start) }}–{{ formatDate(dateRange.end) }}
```
mattandrews, line 49: This was just for consistency with the other tools (which don't have this "Date range:" prefix here)

controllers/tools/surveys/lib/percentages.js
```diff
@@ -3,10 +3,9 @@ const partition = require('lodash/partition');
 
 module.exports = function percentagesFor(responses) {
     const [yes, no] = partition(responses, ['choice', 'yes']);
-
     return {
         yesCount: yes.length,
         noCount: no.length,
-        percentageYes: Math.round((yes.length / responses.length) * 100),
+        percentageYes: ((yes.length / responses.length) * 100).toFixed(1),
```
mattandrews, line 9: This is because the numbers are typically the same month to month, so adding in a decimal place gives a bit more to differentiate with.

PR #3634, opened 2020-08-12
New ccsf form changes

joshgriff465:
These are the changes based on https://tnlcommunityfund.sharepoint.com/:w:/r/teams/England-ServiceDesign-TM/_layouts/15/guestaccess.aspx?e=L8PYeG&share=EQ59Or6bxfRMtFq6y93xXb4B1e8oX7hWAmYP1f_Sk9ncDw. I also changed the enrich.js files in the standard-proposal folder and the enrich.js file in the under10k folder. Not sure why the change is not being documented on github but I looked at the code set and I can see the change. The change was essentially undoing work from a previous commit (https://github.com/biglotteryfund/blf-alpha/pull/3620/files)


controllers/apply/under10k/fields/your-idea.js
```diff
@@ -89,174 +89,61 @@ module.exports = {
             ],
         });
     },
-    fieldYourIdeaPriorities(locale, data = {}, flags = {}) {
+    fieldYourIdeaPriorities(locale) {
         const localise = get(locale);
 
-        const projectCountry = get('projectCountry')(data);
-        const supportingCOVID19 = get('supportingCOVID19')(data);
+        // not needed currently const projectCountry = get('projectCountry')(data);
```
mattandrews, line 95: We can probably just remove this line, no need to comment things out like this when we can just revisit the history on Git etc!

PR #3655, opened 2020-08-19
Misc minor fixes

mattandrews:
A few things in here, just doing some final cleanups 😭 

- Update documentation (added a few missing bits)
- Remove an unused secret (for showing Google Data Studio embeds on application dashboard – this is not used anymore)

And one slightly more interesting change:

David and I pointed out a while ago that the application expiry emails are sent relative to when the application is created. eg:

- I create a new application at 20:23 on the 1st January
- An email is scheduled to be sent at 20:23 on 31st March (eg. one day before expiry)
- On 31st March, the 9am expiry job runs **and doesn't send an email** (because the `dateToSend` value is `2020-31-03 20:23:00`)
- Later that day, the 4pm expiry job runs **and still doesn't send an email** (because 20:23 is still in the future).

The next day at 9am, the expiry job runs again and this time will send the email, but this means the customer only has half a day to finish their application, rather than (at least) 24 hours.

The reason is this query:

    dateToSend: {
        [Op.lte]: moment().toDate(),
    },

This looks up any scheduled email where the `dateToSend` is before/equal to now. My change here makes it:

    dateToSend: {
        [Op.lte]: moment().endOf('day').toDate()
    },

This will look up any scheduled email where the `dateToSend` is before **the end of the current day**, eg. `2020-31-03 23:59:00`. 

This means that in my example above, the 9am job will send the expiry email. It actually means in basically all cases, the 9am job will send all of that day's emails, and the 4pm one will be an "insurance" or backup one, just in case the 9am one fails. Right now, both jobs send a similar number of emails depending on what time applications were created.

Anyway, that explanation was longer than the code change, but hopefully it makes sense!

### DIFF
--- a/common/secrets.js
+++ b/common/secrets.js
@@ -105,8 +105,6 @@ const POSTCODES_API_KEY =
 const S3_KMS_KEY_ID =
     process.env.S3_KMS_KEY_ID || getParameter('s3.kms.key.id');
 
-const DATA_STUDIO_UNDER10K_URL = getParameter('dataStudio.url.awardsForAll');
-
 /**
  * Email expiry secret
  * A shared token sent in POST requests to the email API endpoints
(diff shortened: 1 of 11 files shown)


mo-durvesh (approved):
Great! 👍

joshgriff465 (approved):
Looks good

PR #3657, opened 2020-08-20
tnlcf-lightning sandbox configuration

mo-durvesh:
Config to handle requests from the test.sandbox domain and point to the salesforce sandbox (tnlcf-lightning).

Parameters added to /Web/Test to avoid creating a new environment and additional repeated params (e.g. db connection)

Salesforce class updated to have an alternative function when gms sandbox enabled.

### DIFF
--- a/common/secrets.js
+++ b/common/secrets.js
@@ -83,6 +83,12 @@ const SALESFORCE_AUTH = {
     instanceId:
         process.env.SALESFORCE_INSTANCE_ID ||
         getParameter('salesforce.instanceId'),
+    sandboxApiUrl:
+        process.env.SANDBOX_SALESFORCE_API_URL ||
+        getParameter('sandbox.salesforce.apiUrl'),
+    sandboxUsername:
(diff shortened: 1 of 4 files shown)


common/secrets.js
```diff
@@ -129,6 +141,14 @@ const DOTDIGITAL_API = {
         process.env.DOTDIGITAL_PASS || getParameter('dotdigital.api.password'),
 };
 
+/**
+ * Sandbox domain
+ * Used to specify the requested salesforce endpoint
+ * */
+
+const SANDBOX_DOMAIN =
```
mattandrews, line 149: Might be worth calling this something like `SALESFORCE_SANDBOX_DOMAIN` for clarity as we already have `CONTENT_API_SANDBOX_URL` so there's a bit of ambiguity about which sandbox is which etc.

mattandrews (approved):
Nice work! Well done for figuring out the Salesforce authentication hell, hopefully we don't have to do it again for a while...

👍

PR #3680, opened 2020-09-02
Using length to check for optgroup

mo-durvesh:
optgroup check for if null doesnt work because of the assignment of {}. It needs its length checked to be more than 0


Shayan03 (approved):
Looks good

PR #3688, opened 2020-09-16
Updated sandbox authorise function to use got

mo-durvesh:
sandbox function mirrors original salesforce function using got


marksmithtnlcf (approved):
Good spot! 👍

PR #3713, opened 2020-10-07
Carbon copy of standard proposal form

mo-durvesh:
To support new changes other than the form changes, old and new versions of the standard proposal form have been made and is hidden behind a feature flag


marksmithtnlcf (approved):
Looking good! 💯

PR #3723, opened 2020-10-12
V2 further updates

mo-durvesh:
- content fixes to existing form


joshgriff465 (approved):
Good work, all requested updates look like they are included now

PR #3725, opened 2020-10-14
Content review of error messages

scottoakley:
I've used the [GOV.UK design system component on error messages](https://design-system.service.gov.uk/components/error-message/) to review the error messages - and make them consistent with existing messages designed the same way.

Here's extra rationale for some specific changes: 
- For `fieldOrganisationSupport` I've changed to error message to `Enter a number`  - this should be if they leave the field blank. If they add a number but it's not a whole number, can we have a separate error message to resolve that? It should be `Number of people supported by your organisation must be a whole number (for example, no decimal point)`. This means the customer gets more specific guidance based on the mistake they've made and are more likely to be able to resolve it themselves without calling us
- same as `fieldOrganisationSupport` for `fieldOrganisationVolunteers`, except second error message should be `Number of volunteers at your organisation must be a whole number (for example, no decimal point)`

### DIFF
--- a/controllers/apply/standard-proposal-v2/fields.js
+++ b/controllers/apply/standard-proposal-v2/fields.js
@@ -914,14 +914,14 @@ module.exports = function fieldsFor({ locale, data = {} }) {
                 {
                     type: 'base',
                     message: localise({
-                        en: 'Please tell us your organisation trading name.',
+                        en: 'Enter your organisation's day-to-day name',
                         cy: '',
                     }),
                 },
                 {
                     type: 'any.invalid',
                     message: localise({
-                        en: 'Trading name must not be the same as legal name',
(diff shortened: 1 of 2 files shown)


scottoakley:
Also in [senior-contact-role.js](https://github.com/biglotteryfund/blf-alpha/blob/master/controllers/apply/standard-proposal-v2/fields/senior-contact-role.js) I noticed the following error message, but couldn't work out when it appears to customers:

`type: 'any.allowOnly',
                    message: localise({
                        en: 'Senior contact role is not valid',
                        cy: 'Nid yw’r rôl uwch gyswllt yn ddilys',`

Would be good to know so we can make the error message give more specific instructions so customers can resolve the problem without contacting the Fund.

PR #3734, opened 2020-10-22
Trail text added to all programme list. 

joshgriff465:
If the trail text does not exist for the programme then it will use the description instead


marksmithtnlcf (approved):
Thumbs up

PR #3736, opened 2020-10-23
Warnings do not display for new applications (YFP)

mo-durvesh:
warnings only show if created before launch date (sample set to 16th Nov 2020)


marksmithtnlcf (approved):
Good catch!

Seal of approval

PR #3753, opened 2020-11-02
Flags and fixes

mo-durvesh:
Flags added behind added functionality
supporting covid question removed
length check question removed
their sections removed 
england now same as other countries for start and end date (e.g. leadtime and maxmonths)


marksmithtnlcf (approved):
👍

PR #3770, opened 2020-11-11
use standardv2 for buttons

mo-durvesh:
- buttons configured to use standard v2 links


marksmithtnlcf (approved):
This one made my head hurt.

PR #3787, opened 2020-11-26
Remove v2 flags

mo-durvesh:
- enablestandardv2 removed
- enablesimplev2 removed
- bannersLaunch removed
- conditional logic for flags removed
- tests updated
- snaps updated


marksmithtnlcf (approved):
👋

PR #3845, opened 2021-01-29
Cookie consent updates

joshgriff465:
List of changes:

- Cookie policy page to use yml files
- Welsh translations added
- Text updates to english version of site
- New cookie policy select component added


marksmithtnlcf:
Just a note: when this goes live _all users will need to select their cookie preferences_, regardless of if they have visited the site prior. This is intentional.

marksmithtnlcf (approved):

PR #3876, opened 2021-02-17
WEB-80 Updated media card to use H3 instead of H4.

marksmithtnlcf:
Updated media-object to use an H3 for title instead of H4 as this was causing skipped header levels.


joshgriff465 (approved):
Looks fine

zchughtai (approved):

zchughtai (approved):

PR #3900, opened 2021-03-15
WEB-94 fix joi trim issue

mo-durvesh:
The positioning of the fields seem to be effecting the behavior of Joi.trim()


marksmithtnlcf (approved):
I don't understand why it works, but it does and that's the important part.

